PR #24, opened 2015-10-21
Add account unit tests that target CouchDB and Cloudant

alfinkel:
_What:_

This is the first in a series of PRs that will be adding unit tests that target CouchDB and Cloudant.

- Add account unit tests that run against a CouchDB instance
- Add account unit tests that run against Cloudant
- Add the CouchDB tests to run as part of the Travis CI

_Why:_

We need a set of unit tests that confirm that the library functions as expected while targeting an actual account on CouchDB or Cloudant.

_How:_

- Add account unit tests that target a CouchDB instance
- Add account unit tests that extend the CouchDB tests and add additional tests to target Cloudant
- Add the CouchDB tests to travis.yml so that the CouchDB tests are part of the automated testing for builds.  Leaving Cloudant tests out of the automated run for now.

reviewer: @tomblench
reviewer: @ricellis

BugId: 54193


tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,600 @@
...
+import json
+import base64
+from datetime import datetime
+
+from cloudant.account import CouchDB, Cloudant
+from cloudant.errors import CloudantException
+
+# Set RUN_CLOUDANT_TESTS to True if you wish to execute the Cloudant 
+# class tests.  Defaults to False in order to skip Cloudant tests in Travis-CI.
+RUN_CLOUDANT_TESTS = False
```
tomblench: These need a way of being set from command line or environment variables. Ideally they should be similar to the ones used for other libraries like java-cloudant and sync-android.

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,600 @@
...
+        finally:
+            self.client.disconnect()
+
+    def test_get_db_via_getitem(self):
+        """
+        Test __getitem__ when retrieving a database
+        """
+        try:
+            self.client.connect()
+            self.client.create_database('unittestdb')
```
tomblench: It would be a good idea to uniquify these database names using a helper method - that way multiple parallel test runs can be done on the same instance without any conflicts occurring.

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,600 @@
...
+        So we override the CouchDBAccountTests tearDownClass and do nothing.
+        """
+        pass
+
+    def setUp(self):
+        """
+        Set up test attributes for Cloudant Account tests
+        """
+        self.http = 'https'
+        self.host = '{0}.cloudant.com'.format(self.account)
```
tomblench: You might want a method for running 'cloudant' tests against other hosts, like testy clusters, cloudant local, etc

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,600 @@
...
+        For the Cloudant tests we expect an account to already be created.
+        So we override the CouchDBAccountTests tearDownClass and do nothing.
+        """
+        pass
+
+    def setUp(self):
+        """
+        Set up test attributes for Cloudant Account tests
+        """
+        self.http = 'https'
```
tomblench: This should probably be switchable to http from env vars/command line - it's useful to troubleshoot network traffic with wireshark or similar tools and this can't be done with https.

tests/unit/db/account_tests.py
evansde77: RUN_CLOUDANT_TESTS = os.environ.get('RUN_CLOUDANT_TESTS') != None or False

would let you control that from the environment, then doing export RUN_CLOUDANT_TESTS=1 would let you enable the tests

tests/unit/db/account_tests.py
evansde77: (Came here cos I saw the mail for this comment and got distracted by the env thing...) 
We already have a database unique name helper etc in emissions test: 
https://github.com/cloudant/emissions_test/blob/develop/emissions_test/harnesses/database.py

Might be useful. We can always pop it out into its own testing package if you find it generally useful

tests/unit/db/account_tests.py
alfinkel: These variable are now read from the OS environment.  Done in 806e530.

tests/unit/db/account_tests.py
alfinkel: I took a look at the emissions_test db harness and I think that it may wind up of use to us at some point in the future if/when we start writing tests that hit multiple databases from single sets of tests.  For now, I've added a helper method to the test class that appends a uuid.uuid4() string to the end a database name.  This will provide uniqueness for the purposes of the tests in account_tests.py.  Done in 806e530.

tests/unit/db/account_tests.py
alfinkel: Added the ability to specify a `CLOUDANT_URL` via environment variable.  This should serve as a way to run the Cloudant tests against other hosts.  Done in 806e530.

tests/unit/db/account_tests.py
alfinkel: Added the ability to specify a `CLOUDANT_URL` via environment variable. This should serve as a way to specify http instead of https, if desired. Done in 806e530.

ricellis:
It looks like the Cloudant tests extend the CouchDB tests, so couldn't they share `*_USER` and `*_PASSWORD` env variables (and some common set up) or are you expecting to run different creds in the same test run?

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,640 @@
...
+        """
+        Set up CouchDB instance to be used by all tests
+        """
+        admin_ep = '{0}/_config/admins'.format(
+            os.environ.get('COUCHDB_HOST', '127.0.0.1:5984')
+            )
+        user = os.environ.get('COUCHDB_USER', 'unittest')
+        pwd = os.environ.get('COUCHDB_PASSWORD', 'unittestpassword')
+        resp = requests.put(
+            'http://{0}/{1}'.format(admin_ep, user),
```
ricellis: Do you anticipate any need to use `https` with CouchDB?

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,640 @@
...
+        finally:
+            self.client.disconnect()
+
+    def test_all_dbs(self):
+        """
+        Test getting a list of all of the databases in the account
+        """
+        try:
+            self.client.connect()
+            self.assertIsInstance(self.client.all_dbs(), list)
```
ricellis: Is this expected to be an empty list?
Is there any value in asserting that it contains something (or nothing) to validate that the deserialization of the list of dbs from the server is functioning as expected?

tests/unit/db/account_tests.py
```diff
@@ -0,0 +1,640 @@
...
+        finally:
+            self.client.disconnect()
+
+    def test_create_existing_database(self):
+        """
+        Test creation of already existing database
+        """
+        try:
+            self.client.connect()
+            self.client.create_database('_replicator', throw_on_exists=True)
```
ricellis: I think this would be better creating a database then testing it. I vaguely recall that `_replicator` databases are not created in Cloudant accounts until the first replication job, so this test would fail on a new Cloudant account.

tests/unit/db/account_tests.py
tomblench: I suggested making this switchable when it was previously `https` (to make wireshark etc easier) - maybe I wasn't clear about that - the user should be able to specify either.

alfinkel: Sorry @tomblench, I read your previous comment as a suggestion for just the Cloudant tests and not the CouchDB tests as well.  That has already been fixed and one can specify a URL for the Cloudant tests which will allow them to use either http or https.  The CouchDB tests have not changed based on that comment.  However, after seeing the comment from @ricellis I'll be adding something similar to the CouchDB tests.

tests/unit/db/account_tests.py
alfinkel: Good point.  Changed in d2f7c4a.

tests/unit/db/account_tests.py
alfinkel: Changed to create a set of databases and to check for those databases in the returned list from the call to `all_db()`.  Done in d2f7c4a.

tests/unit/db/account_tests.py
alfinkel: Added the ability to specify the CouchDB URL as an env variable.  This will allow us to specify http or https URLs.  Changed in d2f7c4a.

alfinkel:
Re: https://github.com/cloudant/python-cloudant/pull/24#issuecomment-150240706

As of now the plan is to either run the Cloudant tests along with the CouchDB tests or to only run the CouchDB tests.  So in the case where we run the Cloudant tests along with the CouchDB tests we need separate env variables for each.

ricellis:
+1

tomblench:
+1

PR #25, opened 2015-11-02
Fix database iterator and bulk docs method call

alfinkel:
_What:_

We need to fix the database object iterator so that it does not drop into an infinite loop when returning a set of documents that is greater than the fetch limit.  The database iterator also needs to return objects wrapped as Documents.  Finally the bulk_docs functionality needs to be clarified.

_Why:_

The iterator issue is a bug and needs to be fixed.  The bulk_docs functionality currently is redundant and confusing.

_How:_

- Fix iterator to not fall into infinite loop when documents being retrieved is greater than the fetch limit on the database object.
- Fix iterator to return a Document or a DesignDocument object.
- Fix iterator to attach a Document or a DesignDocument to the database object.
- Remove the current bulk_docs method from the database.py module.
- Rename the bulk_insert method as bulk_docs
- Make "keys" a valid argument in results.py so that it can be passed to the all_docs method which would replicate the functionality of the soon to be removed bulk_docs method.
- Update comments appropriately.
- Remove now unnecessary mocked test for bulk_docs.
- Fix changes integration test bug.

reviewer: @gadamc 

BugId: 55526

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,9 +1,15 @@
-2.0.0b1 (Unreleased)
+2.0.0a2 (Unreleased)
 ====================
 
 - [NEW] Added unit tests targeting CouchDB and Cloudant databases.
 - [FIX] Fixed bug in database create validation check to work if response code
 is either 201 (created) or 202 (accepted).
+- [FIX] Fixed database iterator infinite loop problem and to now yield a 
+Document object.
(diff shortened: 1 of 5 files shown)


gadamc:
Looks good, except could you write a tests for the database object iterator?

alfinkel:
Yes the tests are coming in a follow up PR.

gadamc:
+1

PR #26, opened 2015-11-02
Add common database unit tests

alfinkel:
_What:_

We are adding new common CouchDB and Cloudant database tests.

_Why:_

In order to more accurately test the CouchDatabase and CloudantDatabase class functionality we need to ensure that we have a solid set of unit tests that tests functionality against both CouchDB and Cloudant instances.

_How:_

- Add unit tests that are common to both Couch and Cloudant database implementations.
- Additional Cloudant database specific tests will be added in a future PR.
- Refactor the account module and introduce a unit test base class in order to consolidate common testing functionality.

reviewer: @gadamc

BugId: 54523

### DIFF
--- a/tests/unit/db/database_tests.py
+++ b/tests/unit/db/database_tests.py
@@ -1,0 +1,473 @@
+#!/usr/bin/env python
+# Copyright (c) 2015 IBM. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+_database_tests_
+
+database module - Unit tests for CouchDatabase and CloudantDatabase classes
+
+See configuration options for environment variables in unit_t_db_base
+module docstring.
+
+"""
+
+import unittest
+import requests
+import posixpath
+import os
+
+from cloudant.database import CouchDatabase, CloudantDatabase
+from cloudant.result import Result
+from cloudant.errors import CloudantException
+from cloudant.document import Document
+from cloudant.design_document import DesignDocument
+
+from unit_t_db_base import UnitTestDbBase
+
+class DatabaseTests(UnitTestDbBase):
+    """
+    CouchDatabase/CloudantDatabase unit tests
+    """
+
+    def setUp(self):
+        """
+        Set up test attributes for CouchDB Database tests
+        """
+        super(DatabaseTests, self).setUp()
+        self.client.connect()
+        self.test_dbname = self.dbname()
+        self.db = self.client._DATABASE_CLASS(self.client, self.test_dbname)
+        self.db.create()
+
+    def tearDown(self):
+        """
+        Ensure the client is new for each test
+        """
+        self.db.delete()
+        self.client.disconnect()
+        del self.test_dbname
+        del self.db
+        super(DatabaseTests, self).tearDown()
+
+    def test_constructor(self):
+        """
+        Test instantiating a database
+        """
+        self.assertEqual(self.db.cloudant_account, self.client)
+        self.assertEqual(self.db.database_name, self.test_dbname)
+        self.assertEqual(self.db.r_session, self.client.r_session)
+        self.assertIsInstance(self.db.result, Result)
+
+    def test_retrieve_db_url(self):
+        """
+        Test retrieving the database URL
+        """
+        self.assertEqual(
+            self.db.database_url,
+            posixpath.join(self.client.cloudant_url, self.test_dbname)
+            )
+
+    def test_retrieve_creds(self):
+        """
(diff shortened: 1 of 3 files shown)


tests/unit/db/unit_t_db_base.py
```diff
@@ -0,0 +1,150 @@
...
+
+    @classmethod
+    def setUpClass(self):
+        """
+        If targeting CouchDB, Set up a CouchDB instance otherwise do nothing.
+          
+        Note: Admin Party is currently unsupported so we must create a 
+          CouchDB user for tests to function with a CouchDB instance if one is
+          not provided.
+        """
```
gadamc: ConfigParser is pretty standard in python: https://docs.python.org/2/library/configparser.html. Can fix it later.

tests/unit/db/database_tests.py
```diff
@@ -0,0 +1,480 @@
...
+            self.assertEqual(doc['name'], 'julia')
+            self.assertEqual(doc['age'], age)
+            docs.append(doc)
+            age += 1
+        self.assertEqual(len(docs), 103)
+        # Check that the local database object has been populated
+        # with the appropriate documents
+        expected_keys = []
+        for i in range(103):
+            expected_keys.append('julia{0:03d}'.format(i))
```
gadamc: expected_keys = ['julia{0:03d}'.format(i) for i in xrange(103)] is more pythonic, though not really a big deal here, so just a suggestion

tests/unit/db/database_tests.py
```diff
@@ -0,0 +1,480 @@
...
+            self.assertTrue(doc['_rev'].startswith('1-'))
+            self.assertEqual(doc['name'], 'julia')
+            self.assertEqual(doc['age'], age)
+            docs.append(doc)
+            age += 1
+        self.assertEqual(len(docs), 3)
+        # Check that the local database object has been populated
+        # with the appropriate documents
+        expected_keys = []
+        for i in range(3):
```
gadamc: expected_keys = ['julia{0:03d}'.format(i) for i in xrange(3)] is more pythonic, though not really a big deal here... so just a suggestion

tests/unit/db/database_tests.py
```diff
@@ -0,0 +1,480 @@
...
+        expected_ddoc = self.db.get_design_document('_design/ddoc01')
+        expected_ddoc.add_view('view01', map_func)
+        expected_ddoc.save()
+
+        # Add three standard documents
+        self.populate_db_with_documents(3)
+
+        # Test __get_item__ for standard document
+        doc = self.db['julia001']
+        self.assertIsInstance(doc, Document)
```
gadamc: This set of assertions is repeated in a lot of places. Does it make sense to build a function in order to be more DRY? Or maybe being explicit is better.

tests/unit/db/database_tests.py
```diff
@@ -0,0 +1,480 @@
...
+        self.assertEqual(len(data.get('rows')), 3)
+        self.assertEqual(data['rows'][0]['key'], 'julia006')
+        self.assertEqual(data['rows'][1]['key'], 'julia024')
+        self.assertEqual(data['rows'][2]['key'], 'julia045')
+
+    def test_custom_result_context_manager(self):
+        """
+        Test using the database custom result context manager
+        """
+        self.populate_db_with_documents()
```
gadamc: should we test all of the different kinds of keyword option? https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/result.py#L28-L43

tests/unit/db/database_tests.py
```diff
@@ -0,0 +1,480 @@
...
+        Test retrieving raw output from a design document view
+        """
+        map_func = 'function(doc) {\n emit(doc._id, 1); \n}'
+        data = {'_id': '_design/ddoc01','views': {'view01': {"map": map_func}}}
+        self.db.create_document(data)
+        self.populate_db_with_documents()
+
+        raw_rslt = self.db.get_view_raw_result('_design/ddoc01', 'view01')
+        self.assertIsInstance(raw_rslt, dict)
+        self.assertEqual(len(raw_rslt.get('rows')), 100)
```
gadamc: Is there ever a time when this fails because the server hasn't built the view? probably not...

gadamc:
my comments are only questions or suggestions... so +1 otherwise

tests/unit/db/unit_t_db_base.py
alfinkel: Added FB Case 55684 (https://cloudant.fogbugz.com/f/cases/55684) to track this.

tests/unit/db/database_tests.py
alfinkel: I've gone back and forth with this in the past but have settled on being explicit with testing and assertions.  DRY code everywhere else for sure.

tests/unit/db/database_tests.py
alfinkel: I think you are absolutely right in that we should test all of the keyword options.  But those tests will be part of the Results unit tests that will be upcoming in a future PR.  The purpose of this test is to test the functionality of the database context manager associated with custom results, which I think this test accomplishes.

tests/unit/db/database_tests.py
alfinkel: Thanks.  I made this change throughout.  Done in  985c1f2.

tests/unit/db/database_tests.py
alfinkel: Thanks.  I made this change throughout.  Done in  985c1f2.

PR #28, opened 2015-11-06
Add document module unit tests and make Document class fixes

alfinkel:
_What:_

- Add unit tests that check Document class functionality against a database.
- Enhance the Document class based on unit test results.

_Why:_

Unit tests that target a database provide a more accurate testing platform.

_How:_

- Add Document unit tests that test functionality against either a CouchDB or a Cloudant database.
- Change the Document class in the following ways:
  - Only construct the document_url if the _document_id has a value.
  - Only execute the exists() request if a _document_id has a value.
  - Ensure that a create() request will only create and not update a document remotely.
  - Ensure that a fetch() request replaces the entire dictionary contents of the current Document object. 
  - Rename field_append to list_field_append, rename field_remove to list_field_remove, and rename field_replace to field_set for added clarity.
  - Add validation to list_field_append, list_field_remove, and field_set methods.
  - Clear the dictionary contents, aside from the _id of a Document object that was deleted remotely.
  - Ensure that the dictionary key '_id' and the private attribute _document_id are kept in sync.
  - Retrieve the latest _rev value for put_attachment, get_attachment, and delete_attachment by performing a fetch() call.  The fetch() call provides validation checks as well as refreshes the local Document object.
  - After a successful add/update/delete of an attachment, refresh the local Document object to have the latest content.

reviewer: @ricellis 

BugId: 55710

### DIFF
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -22,6 +22,8 @@ import json
 import posixpath
 import urllib
 import requests
+import copy
+
 from requests.exceptions import HTTPError
 
 from .errors import CloudantException
@@ -51,20 +53,27 @@ class Document(dict):
         self._database_name = database.database_name
         self.r_session = database.r_session
         self._document_id = document_id
+        if self._document_id is not None:
+            self['_id'] = self._document_id
         self._encoder = self._cloudant_account.encoder
 
-    document_url = property(
-        lambda x: posixpath.join(
-            x._database_host,
-            urllib.quote_plus(x._database_name),
-            x._document_id
+    @property
+    def document_url(self):
+        """constructs and returns the document URL"""
+        if self._document_id is None:
+            return None
+        return posixpath.join(
+            self._database_host,
+            urllib.quote_plus(self._database_name),
+            self._document_id
         )
-    )
 
     def exists(self):
         """
         :returns: True if the document exists in the database, otherwise False
         """
+        if self._document_id is None:
+            return False
         resp = self.r_session.get(self.document_url)
         return resp.status_code == 200
 
@@ -86,12 +95,17 @@ class Document(dict):
         """
         if self._document_id is not None:
             self['_id'] = self._document_id
-        headers = {'Content-Type': 'application/json'}
 
+        # Ensure that an existing document will not be "updated"
+        new_doc = copy.deepcopy(self)
+        if new_doc.get('_rev') is not None:
+            new_doc.__delitem__('_rev')
+
+        headers = {'Content-Type': 'application/json'}
         resp = self.r_session.post(
             self._cloudant_database.database_url,
             headers=headers,
-            data=self.json()
+            data=new_doc.json()
         )
         resp.raise_for_status()
         data = resp.json()
@@ -104,11 +118,19 @@ class Document(dict):
         """
         _fetch_
 
-        Fetch the content of this document from the database and update
-        self with whatever it finds
+        Fetch the content of this document from the database and populate
+        self with whatever it finds.  A call to fetch will overwrite
+        any dictionary content currently in self and populate with content
+        found in the database for the document id.
         """
+        if self.document_url is None:
+            raise CloudantException(
+                'A document id is required to fetch document contents.  '
+                'Add an _id key and value to the document and re-try.'
+            )
         resp = self.r_session.get(self.document_url)
         resp.raise_for_status()
+        self.clear()
         self.update(resp.json())
 
     def save(self):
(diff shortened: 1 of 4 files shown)


src/cloudant/document.py
```diff
@@ -293,36 +348,34 @@ def delete_attachment(self, attachment, headers=None):
...
-            headers['If-Match'] = doc_json['_rev']
+            headers['If-Match'] = self['_rev']
 
         resp = self.r_session.delete(
             attachment_url,
             headers=headers
         )
         resp.raise_for_status()
-
+        self.fetch()
```
ricellis: is the `self.fetch()` on 359 insufficient? It seems a shame to do two GETs in such quick succession.

src/cloudant/document.py
```diff
@@ -351,5 +401,5 @@ def put_attachment(self, attachment, content_type, data, headers=None):
...
```
ricellis: Having seen this second example am I correct in thinking that the second `fetch()` call is necessary to update the doc with the new attachment data?

alfinkel: That is correct.  Since we've taken the path of representing each Document as an extension of dict, it made sense that once the attachment is either added/updated/deleted that the local version of the Document dictionary should match what the remote document looks like.  Seems like, the only way to do that is with that second `.fetch()`.  You are right though that it does not help with performance.  We can look later at possibly adding a convenience method to the database object that allows us to perform a sort of "raw" add/update/delete of an attachment to/from a document just based on a document id.  This would give the developer an alternative from using the Document object to handle attachments.  As always they also have the ability to use the endpoint directly as well.

alfinkel: One other possible option is to have a refresh flag either on the method or the entire object to allow a developer to specifically make the choice about whether they would want to have the local dictionary version auto-refreshed.  We've been kicking that idea around for a little while now at the object level (for both databases and documents) but that be a bit of an effort to implement.

ricellis:
I think some of the tests could be broken down a little bit further. I'm quite a big fan of unit tests testing a single thing so the failure is easier to diagnose. For example currently if `test_constructor` has an assertion failure for `self.assertEqual(doc.r_session, self.db.r_session)` it takes just that bit more effort to work out if it was for the `with_id` or `no_id` case. If it was two separate tests `test_constructor_no_id` and `test_constructor_with_id` it would be much easier. The same applies to `test_create_document`, `test_fetch_document`, `test_save_document` each of which is testing 2 or 3 cases. I can see the benefit of testing multiple things in one method , for example in the attachment test case where splitting would add a fair bit of set-up overhead.

src/cloudant/document.py
ricellis: Is it not possible to `add`|`remove` the attachment from the local dictionary if the `put`|`delete` request was successful in place of re-fetching the whole document?

alfinkel: I think it is the attachment metadata that we would need to manipulate in the local dictionary.  As in:
```json
"_attachments": {
    "blah.txt": {
      "content_type": "text/plain",
      "revpos": 11,
      "digest": "md5-bQ7PngG1dTfWpoCLVlLpPw==",
      "length": 1573,
      "stub": true
    }
  }
```

In the case of add/update attachments, I suppose we would be able to do that for all but the digest?  I'm not sure how the digest is populated.

In the case of delete we could remove the second `.fetch()` and programmatically remove the attachment entry from the local dictionary.  I will do that with `delete_attachment`.

alfinkel: Actually calculating attachment length might be sketchy as well.

ricellis: Looks like CouchDB returns a [`Content-MD5` response header](http://docs.couchdb.org/en/1.6.1/api/document/attachments.html#put--db-docid-attname)
I don't want to sidetrack you from the primary objective of this PR, perhaps it is an optimization that can be added later.

alfinkel: I definitely don't see this as a sidetrack.  This is actually the feedback I was looking for w.r.t. the double fetch in the attachment management.  

I just checked and unfortunately it does not appear that the response headers contain what the docs promise and the content length looks to differ by 1 when I ran my test.  In fact, the example provided with the link you referenced from the CouchDB docs shows no Content-MD5 entry in the response header and the content-length differing by 1 as well.  I suppose a little more digging is needed here...  To that end, I've opened https://cloudant.fogbugz.com/f/cases/55930 to track optimization of the library and I've referenced this conversation there.  I'll make the necessary changes to the `document.delete_attachment` method and we can circle back to the `document.put_attachment` optimzation at a later time.

alfinkel: I made the necessary change to the `delete_attachment` method.  Done in fcc5059.

alfinkel:
I split out the unit tests to be more atomic where appropriate.  This is done in  c7c72e3.

tests/unit/db/document_tests.py
```diff
@@ -0,0 +1,501 @@
...
+        doc['age'] = 6
+        self.assertFalse(doc.exists())
+        self.assertIsNone(doc.get('_id'))
+        self.assertIsNone(doc.get('_rev'))
+        doc.create()
+        self.assertTrue(doc.exists())
+        self.assertIsNotNone(doc.get('_id'))
+        self.assertTrue(doc.get('_rev').startswith('1-'))
+
+    def test_create_document_with_docid(self):
```
jcounts: Duplicate method name. I'm surprised this doesn't throw some error. Maybe it's valid for unittest? Either way it's a bit confusing.

ricellis:
After fixing the test method names:
`test_create_document_with_docid` (duplicate that Jordan pointed out)
`test_fetch_non_existing_document_without_docid` (name I find confusing)
then +1

tests/unit/db/document_tests.py
alfinkel: Clearly an oversight on my part.  Thanks for catching this.  I'm also surprised that it works with duplicate method names.  Not too cool...  Anyway, fixed in 5bc0d93.

src/cloudant/document.py
alfinkel: _UPDATE:_

Issuing a HEAD http request to the "attachment URL" will provide the header information that we desire.  We can get the `content-type`, `content-length`, and `content-md5` from this call.  Obviously, it is still an additional http request but it returns a minimal set of data as compared to the `.fetch()` GET request which returns the entire document.

Still left to do is to confirm that we can gleen the `revpos` correctly in every case but I think that should be possible.  We'll handle this piece in a future PR.

PR #29, opened 2015-11-12
Refactor the replicator and add tests

alfinkel:
_What:_

- Refactor ReplicatorDatabase to Replicator.
- Add Replicator unit tests that target a Cloudant or CouchDB instance.

_Why:_

The new Replicator class has the database as an attribute so the Replicator provides an API that works with both a CloudantDatabase and a CouchDatabase _replicator database.

_How:_

- Based on the account client provided when constructing a Replicator object, the Replicator provides a user friendly API for replication tasks for either a Cloudant or a CouchDB account.
- The list_replications method now returns a list of Document wrapped objects.
- Update the integration/replicator_test.py module based on refactoring.
- Update the mocked/replicator_test.py module based on refactoring.
- Add Replicator tests that target a database instance.

reviewer: @emlaver

BugId: 55839

### DIFF
--- a/tests/unit/mocked/replicator_test.py
+++ b/tests/unit/mocked/replicator_test.py
@@ -23,13 +23,14 @@ import unittest
 import mock
 import requests
 
+from cloudant.database import CouchDatabase
 from cloudant.errors import CloudantException
-from cloudant.replicator import ReplicatorDatabase
+from cloudant.replicator import Replicator
 from cloudant.document import Document
 
-class ReplicatorDatabaseTests(unittest.TestCase):
+class ReplicatorTests(unittest.TestCase):
     """
-    Tests for ReplicatorDatabase class
+    Tests for Replicator class
 
     """
     def setUp(self):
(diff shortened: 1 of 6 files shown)


tests/unit/mocked/replicator_test.py
```diff
@@ -127,34 +142,52 @@ def test_create_replication_errors(self):
...
 
             self.assertEqual(
                 repl.list_replications(),
-                ['replication_1', 'replication_2']
+                [{"_id": "replication_1"}, {"_id": "replication_2"}]
             )
+
     def test_replication_state(self):
         """test replication state method"""
+        def getitem(name):
```
emlaver: `getitem` is defined in several functions.  Would it be possible to create one instance of it that can be shared?

tests/unit/mocked/replicator_test.py
```diff
@@ -127,34 +142,52 @@ def test_create_replication_errors(self):
...
         mock_doc.fetch = mock.Mock()
         mock_doc.get = mock.Mock()
         mock_doc.get.return_value = "STATE"
 
-        repl['replication_1'] = mock_doc
+        repl.database['replication_1'] = mock_doc
         self.assertEqual(repl.replication_state('replication_1'), 'STATE')
 
-        with mock.patch('cloudant.replicator.ReplicatorDatabase.__getitem__') as mock_gi:
+        with mock.patch('cloudant.database.CouchDatabase.__getitem__') as mock_gi:
```
emlaver: Is it worth adding 'cloudant.database.CouchDatabase',  'replication_1', and other values into a constants class? Maybe something like:
http://codeyarns.com/2014/12/23/how-to-define-constant-in-python/

tests/unit/mocked/replicator_test.py
alfinkel: I moved `mock_account` to the setUp since it is used in each test.  I also removed the need for the `getitem` function by replacing the `mock_account.side_effect` with `mock_account.return_value`.  Done in 9507fed.

tests/unit/mocked/replicator_test.py
alfinkel: I think in theory it is good practice to follow what you suggest.  However, the ultimate goal here for the mocked tests is to retire and remove them from the repo once we have sufficient test coverage with the new database centric tests that we are adding.  So for now we'll keep touching up the mocked tests so that they still work and eventually they will all just go away.

emlaver: Yea, after you explained in detail the mock tests on Slack, it makes sense to change this at a later date.

emlaver:
LGTM +1

PR #30, opened 2015-11-16
Add DesignDocument unit tests

alfinkel:
_What:_

Add DesignDocument unit tests that target database instances.

_Why:_

In order to adequately test DesignDocument functionality we need to have tests that run against CouchDB and Cloudant instances.

_How:_

- Ensure that each DesignDocument _id is prefaced with `_design/`.
- Create tests in design_document_tests that exercise all of the methods of the DesignDocument class.

reviewer: @emlaver
reviewer: @tomblench 

bugId: 55838

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@
 - [FIX] The Document local dictionary is now refreshed after an add/update/delete of an attachment.
 - [FIX] The Document ``fetch()`` method now refreshes the Document local dictionary content correctly.
 - [BREAKING] Replace the ReplicatorDatabase class with the Replicator class.  A Replicator object has a database attribute that represents the _replicator database.  This allows the Replicator to work for both a CloudantDatabase and a CouchDatabase.
+- [REMOVED] Removed "not implemented" methods from the DesignDocument.
+- [FIX] Add implicit "_design/" prefix for DesignDocument document ids.
 
 2.0.0a1 (2015-10-13)
 ====================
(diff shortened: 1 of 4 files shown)


tomblench:
@alfinkel this PR seems sound to me, but I'd like to understand how it fits in with the rest of the library. For instance, if a user gets `/doc` or `/_all_docs` on a database, do the design documents get returned as the `DesignDocument` type?

Similarly, if the user creates a `Document` with id `_design/abc123` does it get automatically 'promoted' to a `DesignDocument`? What do the other libraries do in this situation?

alfinkel:
Performing the "gets" from a database would be done by using the database object's `__getitem()__` and `__iter()__` functions respectively.  Both of these will return a `Document` or a `DesignDocument` when appropriate.  For example:

```python
doc = database['my_doc']  #returns a Document
ddoc = database['_design/my_ddoc'] # returns a DesignDocument

# iterate through documents retrieved via _all_docs endpoint
for doc in database:
    if isinstance(doc, DesignDocument):
        print '{0} is a design doc'.format(doc['_id'])
    else:
        print '{0} is a document'.format(doc['_id'])
```

However, when creating a document it does not get "promoted" to a `DesignDocument` if the id provided starts with `_design/`.  IMO this should be changed so that the promotion does takes place.  I have an FB case for this particular issue already.

One thing of note is that obviously with Python you don't have to type your variable as you would in Java or similar language.  So for instance in the case of the iteration above, `doc` has no issue switching from `Document` to `DesignDocument` and the user can pretty much remain oblivious to that and treat each `doc` as a `Document` if they wish.  But the option is there if they want it to treat the `doc` as a `DesignDocument` if they so choose and of course it actually is a `DesignDocument`.

emlaver:
+1

tomblench:
+1

PR #33, opened 2015-11-25
New documentation structure using Sphinx and deploying to ReadTheDocs.org

alfinkel:
_What:_

- Use Sphinx to auto-generate documentation (html) based on `docstring` comments in the code.
- Deploy documentation to readthedocs.org, to be hosted there.

**NOTE:** This PR does NOT include the cleaned up versions of all the modules' documentation.  The purpose of this PR is to only get the structure correct.  Each module's updated/cleaned up content will be added in subsequent PRs.  This was done to ensure that the PR size(s) remains relatively manageable.

_Why:_

This library needs a documentation solution and this one allows us to automatically keep our code comments and documentation in synch with relatively little overhead.

_How:_

- Add Sphinx generated `conf.py` and make files - These allow Sphinx/ReadTheDocs to build our documentation in the format we desire.
- Add navigation by adding .rst files for each module.
- Move Getting started section from readme.rst to separate .rst that will be used to generate html on ReadTheDocs.org.

[python lib generated documentation (readthedocs)](http://python-cloudant.readthedocs.org/en/53377-docs-via-sphinx/) - take a look to see the layout and navigation.  As stated before, the individual modules' content is still a bit of a mess right now. To be fixed soon...

reviewer: @ricellis 
reviewer: @tomblench 

BugId: 53377


tomblench:
The output looks pretty good to me. For some methods (functions?) the output includes the method name pre- and appended by an underscore, for example http://python-cloudant.readthedocs.org/en/53377-docs-via-sphinx/account.html#cloudant.account.Cloudant.bill has `_bill_`. Do we know where that comes from and can we suppress it? I'm not much of a pythoner so I don't really know what it means.

tomblench:
Do we need to update the process here: https://github.com/cloudant/cloudant-sync/blob/master/releasing-python-cloudant.md ?

Obviously this can't be done in this PR but any additional steps required to publish documentation should be listed here.

docs/conf.py
```diff
@@ -0,0 +1,294 @@
...
+project = u'python-cloudant'
+copyright = u'2015, IBM'
+author = u'IBM'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '2.0.0b1'
```
ricellis: Is there anyway to maintain these in just one place, e.g by using the values from the `setup.py` args?

docs/index.rst
```diff
@@ -0,0 +1,12 @@
+python-cloudant documentation
+=============================
+
+This is the official Cloudant client library for Python.
+
+.. toctree::
+   :maxdepth: 4
```
ricellis: I see the toctree maxdepth varies from place to place, and they appear to have been chosen carefully so the generated TOC does look good. However, I didn't see a generated index anywhere, is this toctree level what is used for the search? In which case is there any reason not to go deeper than 4?

ricellis:
I'm +1 on this pending the above.

docs/conf.py
alfinkel: We've got an  [FB case](https://cloudant.fogbugz.com/f/cases/53161/Consider-one-location-for-library-version-number) open for this already to handle the version # in the main __init__.py.  I've added a link to this conversation to the case so we can handle the conf.py as well.

docs/index.rst
alfinkel: There is a [generated index](http://python-cloudant.readthedocs.org/en/53377-docs-via-sphinx/genindex.html).  The link to it is at the bottom of the main doc page.  I believe that the maxdepth level refers to TOC display in the main browser and navigation frames only and not the generated index.  I suppose proof of that is that the generated index has method references.  Incidently, the value of 4 was auto generated when I ran the API reference creation script provided by Sphinx.  It appears that 4 seems to be the "optimal" value for display of TOC in both the main and navigation frames.

alfinkel:
Re:  https://github.com/cloudant/python-cloudant/pull/33#issuecomment-159918412

All of the pre and appended underscores in the class and method descriptions have already been removed and will be PR-ed shortly as part of the content clean up.

alfinkel:
Re: https://github.com/cloudant/python-cloudant/pull/33#issuecomment-159922696

@tomblench:  You are right that the release docs need to be updated.  It's on my radar for sure.  At the very least there needs to be a mention of the additional version number in the conf.py.  I am still working out the document generation process though.  As of right now, readthedocs re-generates the docs nightly and there might be a way to have it re-generate for each commit or PR.  Not sure yet, still working through the specifics.  Once I've settled on the process, I'll update the release docs accordingly.  Definitley before the next release.

tomblench:
+1

PR #37, opened 2015-12-07
Add query execution support - part one

### LINKED ISSUE #39: Add query support to the python-cloudant library
We need to add in query management functionality to library so that users no longer have to use the `_find` endpoint directly to perform queries when using this library.

alfinkel:
_What:_

- Added query management functionality to the library.  Resolves #39.
- Unit tests are complete but will be included in a follow up PR to keep the review process relatively manageable.  All unit tests have run successfully to completion.
- Index management is **NOT** included in this PR either, it will also be included in a future PR.

_Why:_

The library needs to provide query management functionality so that users do not have to access query functionality directly via the `_find` endpoint.

_How:_

- Added a Query class and query.py module to handle query parameter processing and request execution to the `_find` endpoint. Found in 4ff83bc.
- Added a QueryResult class to the result.py module to handle slicing and iteration of query results.  Found in 4ff83bc.
- Added a convenience method to the database.py module to execute queries.  Found in db6e822.
- Updated documentation accordingly.  Found in 587841f, 40a23ac, 552eabd.

reviewer: @evansde77 
reviewer: @ricellis

BugId:  51282

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 2.0.0b1 (Unreleased)
 ====================
 
+- [NEW] Added support for Cloudant Query execution.
 
 2.0.0a4 (2015-12-03)
 ====================
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -22,21 +22,21 @@ from collections import Sequence
 from .errors import CloudantArgumentError
 
 ARG_TYPES = {
-    "descending": bool,
-    "endkey": (basestring, Sequence),
-    "endkey_docid": basestring,
-    "group": bool,
-    "group_level": basestring,
-    "include_docs": bool,
-    "inclusive_end": bool,
-    "key": (int, basestring, Sequence),
-    "keys": list,
-    "limit": (int, types.NoneType),
-    "reduce": bool,
-    "skip": (int, types.NoneType),
-    "stale": basestring,
-    "startkey": (basestring, Sequence),
-    "startkey_docid": basestring,
+    'descending': bool,
+    'endkey': (basestring, Sequence),
+    'endkey_docid': basestring,
+    'group': bool,
+    'group_level': basestring,
+    'include_docs': bool,
+    'inclusive_end': bool,
+    'key': (int, basestring, Sequence),
+    'keys': list,
+    'limit': (int, types.NoneType),
+    'reduce': bool,
+    'skip': (int, types.NoneType),
+    'stale': basestring,
+    'startkey': (basestring, Sequence),
+    'startkey_docid': basestring,
 }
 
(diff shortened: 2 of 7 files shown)


src/cloudant/result.py
```diff
@@ -267,3 +269,138 @@ def __iter__(self):
...
+
+        :returns: Rows data in JSON format
+        """
+        if 'skip' in self.options or 'skip' in self._query.keys():
+            msg = 'Cannot use skip parameter with QueryResult slicing.'
+            raise CloudantArgumentError(msg)
+        if 'limit' in self.options or 'limit' in self._query.keys():
+            msg = 'Cannot use limit parameter with QueryResult slicing.'
+            raise CloudantArgumentError(msg)
+        if (
```
evansde77: There seems to be a lot of common code between the Result and QueryResult classes, I would maybe look at subclassing the Result class in QueryResult to remove some of the dupe code.

evansde77:
For the most part this looks good although the implementation of QueryResult looks almost identical to the Result class itself, so maybe better to subclass rather than reimplement it.

CHANGES.rst
```diff
@@ -10,7 +11,7 @@
 2.0.0a3 (2015-12-03)
 ====================
 
-- [NEW] Added API documentation hosted on readthedocs.org.
+- [NEW] Added API documentation hosted on readthedocs.com.
```
ricellis: I'm confused by this change, AFAICT the documentation is on readthedocs.org not .com

Even if this change is correct it should be done separately to the rest of the work (I think a commit pushed to master would be fine). That way at least if for some reason the query PR was reversed or whatever at least this doc fix would still be in place.

ricellis:
IIUC using without a page size option is going to return 100 documents by default. It might be worth calling this out in the `database.get_query_result` or `Query.*_result` methods as I believe the default on the HTTP API is 25.

CHANGES.rst
alfinkel: I'm equally confused as to why I got it in my head that this needed to be changed.  I'll have to chalk this up to the fog of working with the documentation for so long.  Anyway, I changed this back to .org in d807820.

alfinkel:
re: https://github.com/cloudant/python-cloudant/pull/37#issuecomment-162846834.  

Added default comments to all methods that take the "page_size" parameter.  Done in 1f60cd9.

src/cloudant/result.py
alfinkel: I've refactored the QueryResult to extend the Result to minimize code replication.  Done in 1f60cd9.

evansde77:
+1

ricellis:
+1

PR #47, opened 2015-12-09
Add query execution support - part two (tests)

alfinkel:
_What:_

This PR contains the unit tests for code added in https://github.com/cloudant/python-cloudant/pull/37

_Why:_

Unit tests needed to test code

_How:_

- Added Query unit tests
- Added QueryResult unit tests
- Added database convenience method unit tests for get_query_result

_Issue:_ #39 

reviewer: @ricellis
reviewer: @emlaver

### DIFF
--- a/tests/unit/db/query_result_tests.py
+++ b/tests/unit/db/query_result_tests.py
@@ -1,0 +1,380 @@
+#!/usr/bin/env python
+# Copyright (c) 2015 IBM. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Unit tests for the QueryResult class tested against Cloudant only.
+
+See configuration options for environment variables in unit_t_db_base
+module docstring.
+
+"""
+
+import unittest
+import os
+import posixpath
+import requests
+
+from cloudant.query import Query
+from cloudant.result import QueryResult
+from cloudant.errors import CloudantArgumentError
+
+from unit_t_db_base import UnitTestDbBase
+
+@unittest.skipUnless(
+    os.environ.get('RUN_CLOUDANT_TESTS') is not None,
+    'Skipping Cloudant QueryResult tests'
+    )
+class QueryResultTests(UnitTestDbBase):
+    """
+    QueryResult unit tests
+    """
+
+    def setUp(self):
+        """
+        Set up test attributes
+        """
+        super(QueryResultTests, self).setUp()
+        self.db_set_up()
+
+    def tearDown(self):
+        """
+        Reset test attributes
+        """
+        self.db_tear_down()
+        super(QueryResultTests, self).tearDown()
+
+    def test_constructor_with_options(self):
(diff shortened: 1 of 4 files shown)


ricellis:
I noticed that there are two exceptions being asserted in the tests that are the same for cases where the parameter is not specified or it is empty.

`selector = {}`
>                'No selector in the query.  '
>                'Add a selector to define the query and retry.'

`fields = []`
>                'No fields list in the query.  '
>                'Add a list of fields for the query and retry.'

I'm fine with the same exception being used for both, but I think the message should make it clear that the empty option is not sufficient, particularly in the fields case. "No fields list in the query" when an empty list has been provided could be confusing.
Maybe something like:
"No selector in the query or the selector was empty."
Is the same message used in the case of invalid selector JSON? In which case maybe something like:
"No selector in the query or the selector was invalid."

tests/unit/db/query_result_tests.py
```diff
@@ -0,0 +1,380 @@
...
+        self.populate_db_with_documents(5)
+        result = QueryResult(
+            Query(self.db),
+            selector={'_id': {'$gt': 0}},
+            fields=['_id', '_rev'],
+            page_size=3
+        )
+        self.assertEqual(
+            [doc['_id'] for doc in result],
+            ['julia000', 'julia001', 'julia002', 'julia003', 'julia004']
```
ricellis: There appear to be 5 results here even though the page size is 3. Is this because the iterator has all the results and the QueryResult is what filters it down to 3? I was a bit confused because I thought skip/limit couldn't be used with the page size because they were used under the covers for the pages, in which case I would not have expected documents exceeding the page size to have been returned from the server.

alfinkel: I think what might be the problem is the name of the `page_size` parameter.  Its purpose is to dictate the batch size for Result iterations.  Perhaps a better name would be `batch_size`.  We use skip and limit in the iterator logic along with the page_size to manage the stream of documents.  This is the reason why we don't allow skip and limit parameters on iterations of Results/QueryResults.  So in this example the total number of documents matching the query is 5 but the page_size is set to 3 which tells the iterator to retrieve batches of at most 3 documents at a time until you've retrieved all of the documents matching the query.

ricellis: So are there any tests that assert that the batching works? This test just checks 5 results, but really it is 3 and then 2 under the covers right?

alfinkel: These tests, test that the iteration works when the batch size is <, >, and = to the # of documents satisfying the query which tests the `while` loop construct used in the iteration will return all of the documents as expected in all cases.  There is not a test to see if the iteration is ignoring the `page_size` and just looping through one single result set.

ricellis: Right, based on your explanation of what the page really was doing that's why my second question was driving at - there is no validation that the `page_size` is actually used, only that setting it to values < > = the number of results still provides the expected total number of results.

I think that a test that asserts that the underlying requests are broken down into batches might be valuable because it seems like the kind of thing that could be quite readily broken by someone not understanding the `page_size`. If it is a total pain to test then I'd be willing to let it slide, but I feel better for having at least made the suggestion :)

alfinkel:
Re: https://github.com/cloudant/python-cloudant/pull/47#issuecomment-163906491

I changed the error messages to what you suggested.  Done in 5211c5e.  As for selector validation, we only validate that the selector is in fact a `dict`.  That error message would look something like "Argument selector is not an instance of the expected type: dict"

tests/unit/db/query_result_tests.py
alfinkel: I'm not sure to implement a test to validate the batching mechanism, given the way the iteration routine in Result is currently written.

ricellis:
+1

emlaver:
+1

Issue #40, opened 2015-12-08
Add index management support to library


alfinkel:
We need to add in index management functionality to library so that users no longer have to use the `_index` endpoint directly to perform operations on indexes when using this library.

Issue #38, opened 2015-12-08
Add dbcopy to a view?


jroyal:
Is this possible in python-cloudant? In cloudant-python you would just pass the whole view dictionary which could have dbcopy as a field. 

For example in cloudant-python you could do something like this. 

```python

view = {
    "page_id_tester": {
        "map": "function (doc) {\n if(doc.alarm_id) {\nemit(doc.alarm_id, "
               "doc); }}",
        "reduce": "function (keys, values, rereduce) { var latest = ''; var "
                  "ts = 0; var i = 0; for(i = 0; i < values.length; i++){ "
                  "if(values[i]['timestamp'] > ts){ ts = values[i]["
                  "'timestamp']; latest = values[i];}}return latest;}",
        "dbcopy": "page-view-db"
    }
}

response = design_doc.put(params={
        "_id": "_design/%s" % name,
        "views": view
    })
```

But if i'm reading the documentation correctly, python-cloudant restricts view creation in a design document by allowing you to pass in map and reduce strings. So is it not currently possible, or am I just missing something(totally possible)?

alfinkel:
@jroyal, the easiest way to do what you want to do in this library is:

```python
view_def = {
    "page_id_tester": {
        "map": "function (doc) {\n if(doc.alarm_id) {\nemit(doc.alarm_id, "
               "doc); }}",
        "reduce": "function (keys, values, rereduce) { var latest = ''; var "
                  "ts = 0; var i = 0; for(i = 0; i < values.length; i++){ "
                  "if(values[i]['timestamp'] > ts){ ts = values[i]["
                  "'timestamp']; latest = values[i];}}return latest;}",
        "dbcopy": "page-view-db"
    }
}

# assuming a connection to "database"
with DesignDocument(database, '_design/ddoc001') as ddoc:
    ddoc['views'].update(view_def)
```

This will either create or update the design document `_design/ddoc001` and either create or update the `page_id_tester` view on the server depending on whether they already exist there.

Now, I'll say that you are definitely hitting upon something that needs to be enhanced, which is the way in which we handle views.  The DesignDocument/View API that we are currently providing does appear to be limiting the content of the view.  This should/will be fixed.  But know that you can always circumvent any of these restrictions in the meantime by treating any of these objects (View, DesignDocument, Document, etc...) as a dictionary which they all do extend.

I've labeled this issue as an enhancement so we can work towards making the API a bit more inclusive and friendly.  Let me know if this works for you.

[...]

PR #57, opened 2015-12-22
Add query index management support

alfinkel:
_What:_

Provide API to manage Cloudant Query JSON and TEXT indexes.  This PR contains only the code/logic for this functionality.  A follow-up PR(s) will include the unit tests which are complete and all run to successful completion.  The work has been split up to make the work easier to review.

_Why:_

In order to successfully use Cloudant Query, the ability to create and delete query indexes needs to be provided.

_How:_

- Add an indexes module that handles the creation and deletion of Cloudant Query indexes.
- Modify the views module and add a QueryIndexView class to handle the representation of a JSON index as a view in a design document.
- Modify the design_documents module to handle the new views module updates.
- Add API methods to the database module that provide an interface to creating an index, deleting an index and retrieving the list of all indexes in the database.

_Issues:_

- This PR specifically addresses #40.
- As a side effect, #38 is also addressed as part of this PR.

reviewer: @evansde77 
reviewer: @emlaver

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -24,8 +24,10 @@ from requests.exceptions import HTTPError
 from .document import Document
 from .design_document import DesignDocument
 from .views import View
+from .indexes import Index, SearchIndex, SpecialIndex
+from .index_constants import *
 from .query import Query
-from .errors import CloudantException
+from .errors import CloudantException, CloudantArgumentError
 from .result import python_to_couch, Result
 from .changes import Feed
 
@@ -336,7 +338,7 @@ class CouchDatabase(dict):
 
     def delete(self):
         """
-        Delete the current database from the remote instance.
+        Deletes the current database from the remote instance.
         """
         resp = self.r_session.delete(self.database_url)
         resp.raise_for_status()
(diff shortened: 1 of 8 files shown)


src/cloudant/database.py
```diff
@@ -792,6 +793,128 @@ def shards(self):
...
+
+        url = posixpath.join(self.database_url, '_index')
+        resp = self.r_session.get(url)
+        resp.raise_for_status()
+
+        if raw_result:
+            return resp.json()
+        
+        indexes = []
+        for data in resp.json().get('indexes', None):
```
evansde77: since you are going to iterate over this, you may want [] instead of None, since None isnt iterable

src/cloudant/database.py
```diff
@@ -792,6 +793,128 @@ def shards(self):
...
+        :param dict selector: Optional parameter that can be used to limit the
+            index to a specific set of documents that match a query. It uses
+            the same syntax used for selectors in queries.  Only valid when
+            creating a text index.
+
+        :returns: An Index object representing the index created in the
+            remote database
+        """
+        index = None
+        if index_type == 'json':
```
evansde77: I would suggest defining a set of index type module constants like JSON_INDEX_TYPE='json' and using those constants in the code.

src/cloudant/design_document.py
```diff
@@ -86,14 +90,15 @@ def update_view(self, view_name, map_func, reduce_func=None):
...
-            view.reduce = reduce_func
+        view = View(self, view_name, map_func, reduce_func, **kwargs)
         self.views.__setitem__(view_name, view)
 
     def delete_view(self, view_name):
         """
         Removes a View from the locally cached DesignDocument View dictionary.
+        Do not use this method if you wish to
+        delete a query index view.  Use
+        :func:`~cloudant.database.CloudantDatabase.delete_index` instead.
```
evansde77: what happens if they try it any way? Execption? Maybe call that out in the doc. Otherwise this is just a "do not push this button" type of button

src/cloudant/design_document.py
```diff
@@ -102,19 +107,53 @@ def delete_view(self, view_name):
...
         """
         super(DesignDocument, self).fetch()
         for view_name, view_def in self.get('views', {}).iteritems():
-            self['views'][view_name] = View(
-                self,
-                view_name,
-                view_def.get('map'),
-                view_def.get('reduce')
-            )
+            if self.get('language', None) != 'query':
```
evansde77: would make query a module level constant

src/cloudant/indexes.py
```diff
@@ -0,0 +1,271 @@
...
+    	    'name': self._name,
+    	    'type': self._type,
+    	    'def': self._def
+    	}
+
+    def create(self):
+    	"""
+    	Creates the current index in the remote database.
+    	"""
+        payload = {'type': self._type}
```
evansde77: indentation is off on the docstring

src/cloudant/indexes.py
```diff
@@ -0,0 +1,271 @@
...
+        For example ``fields=[{'_id': 'asc'}]``.
+    """
+    def __init__(self, database, **kwargs):
+        if kwargs.keys() != ['fields']:
+            msg = (
+                '{0} provided as argument(s).  The \"special\" index only '
+                'contains the \"fields\" argument.'
+            ).format(kwargs)
+            raise CloudantArgumentError(msg)
+        if kwargs['fields'] != [{'_id': 'asc'}]:
```
evansde77: This is a pretty specific kwarg, since it is limited, I would maybe make this a boolean in the call like ascending_order=True and then if it is True, just set the fields content to {'_id': 'asc'} 
Make the API feel more like an API than exposing deeper knowledge of the special index, goal is to make it easy for the library user.

src/cloudant/indexes.py
```diff
@@ -0,0 +1,271 @@
...
+                'The \"special\" index \"fields\" argument must be set to {1}.'
+            ).format(kwargs['fields'], [{'_id': 'asc'}])
+            raise CloudantArgumentError(msg)
+        super(SpecialIndex, self).__init__(
+            database,
+            None,
+            '_all_docs',
+            **kwargs
+        )
+        self._type = 'special'
```
evansde77: more constant style stuff. Looks like having an index_constants.py module with these all defined would make it a bit cleaner & easier to determine actual constants from magic strings in the code base.

src/cloudant/views.py
```diff
@@ -325,3 +334,94 @@ def custom_result(self, **options):
...
+        QueryIndexView objects are not callable.  If you wish to execute a query
+        using a query index, use
+        :func:`~cloudant.database.CloudantDatabase.get_query_result` instead.
+        """
+        raise CloudantException(
+            'A QueryIndexView is not callable.  If you wish to execute a query '
+            'use the database \'get_query_result\' convenience method.'
+        )
+
+    def make_result(self, **options):
```
evansde77: Guess this is overriding a base method that doesnt make sense for this object, would maybe spell that out in the docstring.

evansde77:
Mostly stylistic/documentation comments. Hopefully unittests are coming as well?

alfinkel:
Unit tests and doc updates are on the way in follow up PR(s).  The work is complete and the unit tests all ran successfully.  I split the work up into separate PRs to hopefully make the review process easier.  I updated the PR description to include this info.

src/cloudant/views.py
alfinkel: Added more detail to the docstring for this method.  Done in b235b950d4efc80ddce6d75cfdf0ee93e48126f7.

src/cloudant/database.py
alfinkel: Fixed in 5740aa251ca744ad62d53bb36986bf68e2da4340.

src/cloudant/database.py
alfinkel: Added an index_constants module in 21c9d5dd5d95925c23a67fe9318f26c8a10e07b3 and changed all references in the database module to reference the index constants in 5740aa251ca744ad62d53bb36986bf68e2da4340.

src/cloudant/design_document.py
alfinkel: I actually added some more validation to `add_view`, `update_view`, and `delete_view` that raises an exception if a QueryIndexView is found and added more detail to the docstrings as well.  Done in 2707ad460db815447a917e39d8b0d67536a7ea3f.

src/cloudant/design_document.py
alfinkel: Done in 2707ad460db815447a917e39d8b0d67536a7ea3f.

src/cloudant/indexes.py
alfinkel: Fixed in daf19a6dd600acb796fb87dc7f78a0431b5940dc.

src/cloudant/indexes.py
alfinkel: Done in 21c9d5dd5d95925c23a67fe9318f26c8a10e07b3 and daf19a6dd600acb796fb87dc7f78a0431b5940dc.

src/cloudant/indexes.py
alfinkel: After reading your comment and giving some more thought to the actual purpose of this class, I decided to refactor it slightly.  Essentially the purpose of a SpecialIndex in this library is for use by the database API method `get_all_indexes()`.  'get_all_indexes()` retrieves the indexes in the database through the _index endpoint and wraps the primary "special" index as a SpecialIndex object as well as wrapping JSON indexes as Index objects and TEXT indexes as SearchIndex objects and returns the index list to the caller.

Since you cannot really do anything with a SpecialIndex object other than instanciating it, it didn't really make much sense to have any kind of validation as was originally being done.  So the validation was removed and the method signature was changed to match that if the parent Index class.  This change provides us with flexibility if we (Cloudant) decide to change the "special" index definition server side at any point down the road (unlikely but still possible), this library implementation will still be able to handle it.

This is done in daf19a6dd600acb796fb87dc7f78a0431b5940dc and 5740aa251ca744ad62d53bb36986bf68e2da4340.

src/cloudant/indexes.py
```diff
@@ -0,0 +1,260 @@
...
+        self._type = JSON_INDEX_TYPE
+        self._def = kwargs
+
+    @property
+    def index_url(self):
+        """
+    	Constructs and returns the index URL.
+
+    	:returns: Index URL
+    	"""
```
evansde77: indent

evansde77:
LGTM, although there are a couple of indent/spacing things, might be worth running pep8 over it, see what it tells you.

alfinkel:
I ran pep8 and all should be good with indentation and spacing now.  It looks like I a mixture of tabs and spaces in Sublime Text was the cause.  Obviously I need to fix my indenting configuration in the editor.  Thanks for catching that.

emlaver:
+1

Issue #76, opened 2016-01-04
Add index management unit tests for design_document and views modules


alfinkel:
In support of #40.

PR #75, opened 2016-01-04
Add unit tests for the indexes module

alfinkel:
_What:_

Add unit tests for the indexes module in support of #40.

_Why:_

To ensure that the indexes module logic is thoroughly tested.

_How:_

- Add an indexes_tests.py module.
- Add IndexTests class with tests for the Index class.
- Add SearchIndexTests class with tests for the SearchIndex class.
- Add SpecialIndexTests class with tests for the SpecialIndex class.

_Issues:_

- #74 in support of #40 

reviewer: @evansde77 
reviewer: @brynh

PR #77, opened 2016-01-04
Add index management unit tests for the design_documents and views modules

alfinkel:
_What:_

Add index management unit tests for the design_documents and views modules in support of #40.

_Why:_

To ensure that the updated design_document and views index management logic is thoroughly tested.

_How:_

- Update the design_document_tests module with index management tests.
- Update the views_tests module with index management tests.
- Add a QueryIndexViewTests class to test QueryIndexView functionality.

_Issues:_

- #76 in support of #40 

reviewer: @evansde77 
reviewer: @ricellis

### DIFF
--- a/tests/unit/db/design_document_tests.py
+++ b/tests/unit/db/design_document_tests.py
@@ -24,9 +24,10 @@ module docstring.
 
 import unittest
 
+from cloudant.document import Document 
 from cloudant.design_document import DesignDocument
-from cloudant.views import View
-from cloudant.errors import CloudantArgumentError
+from cloudant.views import View, QueryIndexView
+from cloudant.errors import CloudantArgumentError, CloudantException
 
 from unit_t_db_base import UnitTestDbBase
 
@@ -114,6 +115,21 @@ class DesignDocumentTests(UnitTestDbBase):
                 'View view001 already exists in this design doc'
             )
 
+    def test_adding_query_index_view(self):
+        """
+        Test that adding a query index view fails as expected.
+        """
+        ddoc = DesignDocument(self.db, '_design/ddoc001')
+        ddoc['language'] = 'query'
+        with self.assertRaises(CloudantException) as cm:
+            ddoc.add_view('view001', {'foo': 'bar'})
+        err = cm.exception
+        self.assertEqual(
+            str(err),
+            'Cannot add a MapReduce view to a '
+            'design document for query indexes.'
+        )
+
     def test_update_a_view(self):
         """
         Test that updating a view updates the contents of the correct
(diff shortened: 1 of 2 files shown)


tests/unit/db/design_document_tests.py
```diff
@@ -164,35 +213,261 @@ def test_delete_a_view(self):
...
+                            'options': {'def': {'fields': ['name', 'age']},
+                                        'w': 2}
+                            }
+                    }
+        }
+        self.db.create_document(data)
+        ddoc = DesignDocument(self.db, '_design/ddoc001')
+        ddoc.fetch()
+        try:
+            ddoc.delete_view('view001')
```
evansde77: you might be better off testing these with the self.assertRaises statement. 
See: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises

evansde77:
LGTM, but would suggest using the assertRaises stuff to verify exceptions rather than the try:except loops.

tests/unit/db/design_document_tests.py
alfinkel: I replaced every instance of exception checking with assertRaises in the code that was added/updated in PR #75, PR #77 (this PR), and PR #79.  I also created Issue #80 so that we can circle back and make this change in all of our unit tests.

tests/unit/db/design_document_tests.py
```diff
@@ -164,35 +211,251 @@ def test_delete_a_view(self):
...
+            ddoc.save()
+        err = cm.exception
+        self.assertEqual(
+            str(err),
+            'View view001 must be of type QueryIndexView.'
+        )
+
+    def test_save_passes_query_view_validation(self):
+        """
+        Ensure that all views are query views when language is set to query.
```
ricellis: I don't understand this comment because it looks to me like this test is saving a MR view and the language is not set to query (unless it is by default)?

tests/unit/db/design_document_tests.py
```diff
@@ -164,35 +211,251 @@ def test_delete_a_view(self):
...
+
+        with self.assertRaises(CloudantException) as cm:
+            del ddoc['language']
+            ddoc.save()
+        err = cm.exception
+        self.assertEqual(str(err), 'View view001 must be of type View.')
+
+    def test_save_passes_view_validation(self):
+        """
+        Ensure that all views are not query views when language is not set to
```
ricellis: It looks like the language is set to `query` in this one.

tests/unit/db/views_tests.py
```diff
@@ -265,5 +284,144 @@ def test_custom_result_context_manager(self):
...
+    is not necessary.
+    """
+
+    def setUp(self):
+        """
+        Set up test attributes
+        """
+        self.ddoc = mock.Mock()
+        self.ddoc.r_session = 'mocked-session'
+        self.ddoc.document_url = 'http://mocked.url.com/my_db/_design/ddoc001'
```
ricellis: Although we don't make a request, may I suggest using one of the reserved testing/documentation domain names or IP addresses for this mock url? `url.com` appears to be a real thing.
`example.com` is reserved, but if that doesn't suit I'm sure you can find others in the relevant IETF RFCs.

tests/unit/db/design_document_tests.py
alfinkel: These "save" tests were poorly named.  I renamed all four of the "save" tests to more appropriate names as well as changed the docstring comments for each.  They should make more sense now.  Done in https://github.com/cloudant/python-cloudant/commit/83e8033f06441ef1485365876480bbfb86b95298.

tests/unit/db/views_tests.py
alfinkel: Changed to use example.com.  Done in https://github.com/cloudant/python-cloudant/commit/cb9828a55aaa18a1e9b05db4eed94b94dcabaa3b.

ricellis:
+1

Issue #41, opened 2015-12-08
Fix pylint errors after switching to pylint 1.5.0


alfinkel:
As part of the Travis CI build the default version of pylint is now 1.5.0.  This switch from 1.4.4 appears to have additional code analysis restrictions that are causing our pylint test that runs as part of the build to return less than a 10 out of 10 which causes the entire build to fail.

- We need to refactor the library code based on the pylint suggestions so that the code base returns 10 out of 10 again using pylint 1.5.0.
- Currently the pylint test is being skipped so that the build does not fail.

PR #81, opened 2016-01-05
Update library code based on pylint 1.5.0 findings

alfinkel:
_What:_

Update the library code so that it will successfully pass the pylint 1.5.0 unit test.

_Why:_

Pylint helps us keep the code clean and following a standard.  The pylint 1.5.0 release introduced a few wrinkles that were not in place with the previous pylint 1.4.4 version we were using.  To keep current with pylint we needed to make these changes to the library.

_How:_

- Removed the assertion in the pylint unit test that checked STD_ERR for output.  This was a redundant check and pylint was issuing informational warnings there so we were getting test failure when there was nothing wrong.
- Update the pylintrc configuration as needed to work with 1.5.0.
- Refactored code based on pylint 1.5.0 findings.
- Set the pylint unit test to no longer be skipped as part of the CI.

_Issues:_

- #41 

reviewer @ricellis
reviewer @rhyshort

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -63,9 +63,10 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 
+# Disable "redefined-variable-type" refactor warning messages
 # Disable "too-many-..." and "too-few-..." refactor warning messages
 # Disable "locally-disabled" message
-disable=R0901,R0902,R0903,R0904,R0913,R014,R0915,locally-disabled
+disable=R0204,R0901,R0902,R0903,R0904,R0913,R0914,R0915,locally-disabled
 
 
 [REPORTS]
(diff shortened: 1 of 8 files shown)


pylintrc
```diff
@@ -65,7 +65,7 @@ confidence=
 
 # Disable "too-many-..." and "too-few-..." refactor warning messages
 # Disable "locally-disabled" message
-disable=R0901,R0902,R0903,R0904,R0913,R014,R0915,locally-disabled
+disable=R0204,R0901,R0902,R0903,R0904,R0913,R014,R0915,locally-disabled
```
ricellis: It appears that `R014` does not exist was it intended to be `R0914`?

ricellis:
+1 once `R014` is fixed.

rhyshort:
+1

Issue #50, opened 2015-12-14
Fix database __iter__() to wrap Documents correctly


alfinkel:
The CouchDatabase.\__iter__() method haa a bug when it comes to wrapping the yielded documents as Document and DesignDocument.  Currently, documents are wrapped as Document(self) and DesignDocument(self).  They need to be wrapped as Document(self, doc['id']) and DesignDocument(self, doc['id']).  The current way creates Document objects without a `_document_id` attribute therefore without a valid `document_url`.  This leaves these objects useless when it comes to things related to remote processing.

The "for" block at the end of the __iter__() method should read like this after the fix:

```python
for doc in docs:
    # Wrap the doc dictionary as the appropriate
    # document object before yielding it.
    document = {}
    if doc['id'].startswith('_design/'):
        document = DesignDocument(self, doc['id'])
    else:
        document = Document(self, doc['id'])
    document.update(doc['doc'])
    super(CouchDatabase, self).__setitem__(
        doc['id'],
        document
    )
    yield document
```

This problem is referenced here:

https://github.com/cloudant/python-cloudant/issues/32#issuecomment-159668909

Obviously this needs to be thoroughly tested and additional unit tests added to check these conditions.

PR #82, opened 2016-01-06
Fix database iterator to wrap documents correctly

alfinkel:
_What:_

Fix a bug with the database iterator so that the documents that it returns have a valid document_url property.  Previously it was not being set.

_Why:_

WIthout a document_url property the Document/DesignDocument cannot be used to update or delete the corresponding document in the remote database.

_How:_

When instantiating the Document/DesignDocument object in the database iterator we now pass in the document id to the Document/DesignDocument constructor so that the _document_id private variable is set correctly thereby ensuring the document_url will also be set correctly.

_Issues:_

- #50 

reviewer: @emlaver 
reviewer: @brynh

### DIFF
--- a/tests/unit/db/database_tests.py
+++ b/tests/unit/db/database_tests.py
@@ -414,6 +414,44 @@ class DatabaseTests(UnitTestDbBase):
             self.assertEqual(doc['name'], 'julia')
             self.assertEqual(doc['age'], int(id[len(id) - 3 : len(id)]))
 
+    def test_document_iteration_returns_valid_documents(self):
+        """
+        This test will check that the __iter__ method returns documents that are
+        valid Document or DesignDocument objects and that they can be managed
+        remotely.  In this test we will delete the documents as part of the test
+        to ensure that remote management is working as expected and confirming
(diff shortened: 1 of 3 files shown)


tests/unit/db/database_tests.py
```diff
@@ -414,6 +414,43 @@ def test_document_iteration_over_fetch_limit(self):
...
+                )
+                docs.append(doc)
+            doc.delete()
+
+        # Confirm successful deletions
+        for doc in self.db:
+            self.fail('All documents should have been deleted!!!')
+
+        # Confirm that both Document and DesignDocument objects were returned
+        self.assertEqual(len(docs), 3)
```
emlaver: Is it worth adding a comment in the docstring about how many documents and design documents your expecting to return?

alfinkel: I think you mean just adding a comment (not docstring).  Done in a7f9cfa39aed93ee2996499f98961c37d53e25f4

emlaver:
+1

brynh:
+1

Issue #56, opened 2015-12-17
Unable to fetch document


ashutoshh:
I am saving documents in the cloudant with url as the value of _id field. I am able to save the documents successfully but while retrieving its giving me key error.
Example:
client = Cloudant(USERNAME, PASSWORD,url=cloudantURL)
client.connect()
data={'_id':'https://www.mssdfs.com'}
db = client['dbname']
db.create_document(data)
name='https://www.mssdfs.com'
doc = db[name]
if doc.exists():
    print True
else:
    print False

this gives me key error.

alfinkel:
This is a known bug.  You can find the details and a work around here #32.

alfinkel:
Your simpliest route to take would be to use urllib and replace the line `doc = db[name]` with `doc = Document(db, urllib.quote(name, safe='')`.  Then your check for doc.exists() should work.  

You could also change the line to be `doc = db[urllib.quote(name, safe='')]` and that also should get your doc.exists() check working but it may have an unwanted side effect in that it will add the urlencoded key to the locally cached `db` dictionary.  So you would have 2 keys for the same document, one would be 'https://www.mssdfs.com' and the other would be 'https%3A%2F%2Fwww.mssdfs.com'.

Anyway, this is what I would change your code to for the time being until we get the bug fixed and pushed out in an upcoming release:

```python
import urllib
from cloudant.document import Document
client = Cloudant(USERNAME, PASSWORD,url=cloudantURL)
client.connect()
data={'_id':'https://www.mssdfs.com'}
db = client['dbname']
db.create_document(data)
name='https://www.mssdfs.com'
doc = Document(db, urllib.quote(name, safe=''))
if doc.exists():
    print True
else:
    print False
```

alfinkel:
This bug is resolved and the fix should be made available with the 2.0.0b1 release.

Issue #32, opened 2015-11-21
Populating _id with a url results in document that cannot be deleted


ewann:
I don't know if this is a bug or feature request. Either way, to repro:

```python
from cloudant.account import CouchDB
import base64
client = CouchDB('admin', 'admin', url='http://127.0.0.1:5984')
client.connect()

try:
    db = client['some_test_db']
except Exception,e:
    db = client.create_database('some_test_db')

unencoded = 'http://www.google.com'

encoded = base64.b64encode(unencoded)

#succeeds
data = {
    '_id': encoded,
    'someKey': 'someValue'
    }
db.create_document(data)
db[encoded]
db[encoded].delete()

#failure on document delete:
data = {
    '_id': unencoded,
    'someKey': 'someValue'
    }
db.create_document(data)
db[unencoded]
db[unencoded].delete()
```

alfinkel:
@ewann, This is definitely a bug.  I added an internal FogBugz case to track it.  It will likely be fixed for the beta release.  For now you can get around the problem by manually fetching the document while encoding the `_id` and then deleting.  As in:

```python
import urllib
doc = db[urllib.quote(unencoded, safe='')]
doc.fetch()
doc.delete()
```

ewann:
thanks. In case anyone follows along behind, here is I suspect the same bug, differently manifest:

#succeeds:
import urllib
my_document = my_database[urllib.quote('Mm5kMTQ0ODQxNTY0MC45OA==', safe='')]
#my_document = my_database['Mm5kMTQ0ODQxNTY0MC45OA==']
print my_document
from cloudant.document import Document
my_document['title'] = 'Jules'
my_document['text'] = 6
my_document.save()

#fails:
my_document = my_database['Mm5kMTQ0ODQxNTY0MC45OA==']
print my_document
from cloudant.document import Document
my_document['title'] = 'Jules'
my_document['text'] = 6
my_document.save()

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/cloudant/document.py", line 148, in save
    self.create()
  File "/usr/lib/python2.7/site-packages/cloudant/document.py", line 110, in create
    resp.raise_for_status()
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 837, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 409 Client Error: Conflict for url: http://127.0.0.1:5984/db

[...]

PR #84, opened 2016-01-07
Fix document url so that it is appropriately encoded

alfinkel:
_What:_

Ensure that if the document id of a document contains special characters that the document url generated from that document id is encoded appropriately.

_Why:_

A bug was discovered #32 that sites that you cannot use special characters such as `/` in the document id because it creates a problem in the document id because it is not being encoded correctly.  This fix resolves that issue.

_How:_

- Handle the url for a document by encoding the private `_document_id` attribute before using it to generate the `document_url`.
- Handle the url for a design document by stripping `_design/` from the start of `_document_id` and then encoding the remainder of the `document_id` before using it to generate the `document_url`.  This is being done in the Document class rather than the DesignDocument class because the Document class can be used to create a DesignDocument if desired by the user and the DesignDocument extends  the Document.

_Tests:_

- Document tests added to check for url encoding.
- Document tests added to check create, read, update, delete of documents where document id has special characters and the url needs to be encoded.
- DesignDocument tests added to check create, read, update, delete of design documents where document id has special characters and the url needs to be encoded.

_Issues:_

- #32 
- #56 

reviewer @emlaver
reviewer @rhyshort

### DIFF
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -76,10 +76,21 @@ class Document(dict):
         """
         if self._document_id is None:
             return None
+
+        # handle design document url
+        if self._document_id.startswith('_design/'):
+            return posixpath.join(
+                self._database_host,
+                urllib.quote_plus(self._database_name),
+                '_design',
+                urllib.quote(self._document_id[8:], safe='')
+            )
+
+        # handle document url
         return posixpath.join(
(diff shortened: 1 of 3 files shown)


emlaver:
LGTM +1

rhyshort:
+1

Issue #86, opened 2016-01-16
Don't make the fields argument required for a Query


JanekLehr:
Currently a python-cloudant query requires the user to pass in a list of fields in the `fields` parameter, but Cloudant does not require this parameter to be passed. Cloudant will default to returning all fields of a document if `fields` is not specified. Could python-cloudant be changed to mimic this behavior?

https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/query.py#L185

alfinkel:
Looks like that is a bug for sure.  We'll get that fixed.  Thanks.

JanekLehr:
@alfinkel 
I'm writing a fix for this but I don't know how I should be running the test suites. I'm trying to run just the unit tests but I'm running into issues because I don't have a CouchDB instance running locally. Are the unit tests making real connections to a DB? Here's the stack trace from the `setUpClass`:
```
Error
Traceback (most recent call last):
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/Users/janeklehr/Development/python-cloudant/tests/unit/db/unit_t_db_base.py", line 87, in setUpClass
    data='"{0}"'.format(os.environ['DB_PASSWORD'])
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/requests/api.py", line 120, in put
    return request('put', url, data=data, **kwargs)
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/requests/api.py", line 53, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/janeklehr/Development/python-cloudant/venv/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
[...]

PR #89, opened 2016-01-21
Remove the fields parameter from required Query parameters

JanekLehr:
_What:_

Remove the enforcement of the `fields` as a required parameter for making a query.

_Why:_

Cloudant supports not passing in the `fields` when making a query. If no fields are provided then all the fields of each document are returned. This PR makes sure that python-cloudant mimics that behavior.

_How:_

- In `CloudantDatabase.get_query_result()` the `fields` parameter is checked for existence before instantiating the `Query`.
- In `Query.__call__` I removed the check for `fields` being `None` or empty.

_Tests:_

- The query callable tests that test for `fields` being empty or `None` were removed. 
- The tests were updated to treat `fields` as a `**kwargs` argument.
- Test that calling `db.get_query_result` handles a nonexistent `fields` parameter.

_Issues:_
- #86 

reviewer @alfinkel 
reviewer @ricellis

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -919,7 +919,8 @@ class CloudantDatabase(CouchDatabase):
             raise CloudantArgumentError(msg)
         index.delete()
 
-    def get_query_result(self, selector, raw_result=False, **kwargs):
+    def get_query_result(self, selector, fields=None, raw_result=False,
+                         **kwargs):
         """
         Retrieves the query result from the specified database based on the
         query parameters provided.  By default the result is returned as a
@@ -987,7 +988,10 @@ class CloudantDatabase(CouchDatabase):
         :returns: The result content either wrapped in a QueryResult or
             as the raw response JSON content
         """
-        query = Query(self, selector=selector)
+        if fields:
+            query = Query(self, selector=selector, fields=fields)
+        else:
+            query = Query(self, selector=selector)
         if raw_result:
             return query(**kwargs)
         if kwargs:
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -919,7 +919,7 @@ class CloudantDatabase(CouchDatabase):
             raise CloudantArgumentError(msg)
         index.delete()
 
-    def get_query_result(self, selector, fields=None, raw_result=False, **kwargs):
+    def get_query_result(self, selector, raw_result=False, **kwargs):
         """
         Retrieves the query result from the specified database based on the
         query parameters provided.  By default the result is returned as a
@@ -987,7 +987,7 @@ class CloudantDatabase(CouchDatabase):
         :returns: The result content either wrapped in a QueryResult or
             as the raw response JSON content
         """
-        query = Query(self, selector=selector, fields=fields)
+        query = Query(self, selector=selector)
         if raw_result:
             return query(**kwargs)
         if kwargs:
--- a/tests/unit/db/database_tests.py
+++ b/tests/unit/db/database_tests.py
@@ -696,6 +696,57 @@ class CloudantDatabaseTests(UnitTestDbBase):
             ['julia001', 'julia002', 'julia003', 'julia004']
         )
 
+    def test_get_query_result_without_fields(self):
+        """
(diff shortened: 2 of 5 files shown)


JanekLehr:
This PR is incomplete because I do not know how to go about running the tests suites locally. Any help with this would be appreciated.

JanekLehr:
So apparently the checks passed but I imagine some tests still need to be changed?

JanekLehr:
fixes #86

alfinkel:
Yes you are correct that some tests will need to be changed.  Those tests are Cloudant specific tests that don't run as part of the Travis CI.  See https://github.com/cloudant/python-cloudant/issues/86#issuecomment-173692830 for details on running Cloudant tests.

JanekLehr:
@alfinkel: The unit tests should be updated now. I still ran into issues with the replication tests locally but my changes don't affect replication, and I couldn't get them passing on the `master` branch either.

alfinkel:
@JanekLehr - Could you update your initial comment to match the pattern that we use for pull requests.  You can use https://github.com/cloudant/python-cloudant/pull/84 as a reference.  Tag me as a reviewer and I'll look to getting one other to review after I have an initial look through.

src/cloudant/database.py
```diff
@@ -919,7 +919,7 @@ def delete_index(self, design_document_id, index_type, index_name):
             raise CloudantArgumentError(msg)
         index.delete()
 
-    def get_query_result(self, selector, fields=None, raw_result=False, **kwargs):
+    def get_query_result(self, selector, raw_result=False, **kwargs):
```
alfinkel: I think that rather than introducing this breaking change you should return the original `get_query_result` method signature and then handle the case where `fields` is None or an empty list with `if` logic below (Line 990).

alfinkel: I would also add 2 tests to the `database_tests.py` that ensure that when passing either None or an empty list to `fields` returns all of the fields expected.

alfinkel:
Can you add a [FIX] entry to the CHANGES.rst for this bug fix?

JanekLehr:
@alfinkel I addressed your comments

alfinkel:
@JanekLehr, I'll get someone else from the team to take a look but it will likely not be until tomorrow (1/26).

Looks good.  +1 from me.

tests/unit/db/database_tests.py
```diff
@@ -696,6 +696,57 @@ def test_get_query_result_without_kwargs(self):
             ['julia001', 'julia002', 'julia003', 'julia004']
         )
 
+    def test_get_query_result_without_fields(self):
+        """
+        Test that retrieving the QueryResult for a query works as expected when
+        executing a query
```
ricellis: I think it would be nice if this comment said what the expected behaviour is. Maybe add something like:
> Assert that the QueryResult docs include all expected fields when no fields parameter is provided.

tests/unit/db/database_tests.py
```diff
@@ -696,6 +696,57 @@ def test_get_query_result_without_kwargs(self):
...
+            self.assertEqual(doc_fields, expected_fields)
+        self.assertEqual(
+            [doc['_id'] for doc in result],
+            ['julia001', 'julia002', 'julia003', 'julia004']
+        )
+
+    def test_get_query_result_with_empty_fields_list(self):
+        """
+        Test that retrieving the QueryResult for a query works as expected when
+        executing a query
```
ricellis: Maybe add:
> Assert that the QueryResult docs include all expected fields when an empty field list is provided.

ricellis:
@JanekLehr thanks for fixing this. Al asked me to take a look and I've just added a couple suggestions for improving the test comments.
Otherwise I'm +1 for this.

alfinkel:
@JanekLehr, it looks like you've added the comments to the tests that Rich requested.  Can you squash your commits down to the original two and then let me know so that I can merge this PR into master.

JanekLehr:
@alfinkel - I've squashed the commits.

@ricellis -  No problem. I am happy to contribute.

Issue #23, opened 2015-10-21
Plans for Python 3 support


FvD:
Could you specify in the readme if you have any plans for Python 3 support in the (near) future?. 

Perhaps it would make sense to start a list with items to fix for Py3 support such as the use of [basestring](https://github.com/oxplot/fysom/issues/1)  and [type.NoneType](https://bugs.python.org/issue19438).

alfinkel:
Support for Python 3 is on our radar, for sure.  In all likelihood we will look at that once the library is in a stable release (2.0.0).

alfinkel:
We need to test the library in a Python 3 environment and react accordingly.  Any bugs found or changes that need to be made will be handled as separate GH issues.

Additionally as part of this effort we should also add to the Travis CI testing in multiple environments including Python 3.

PR #91, opened 2016-01-27
py3 support

rredburn:
_What:_

Add Python 3 compatibility.

_Why:_

Python 3, now in it's fifth revision, has obtained enough critical mass in the developer community to make supporting it worthwhile.  Bluemix supports Python 3 Cloud Foundry applications as a built in community runtime and makes attaching a Cloudant service an easy and desirable option.  Lack of a Python 3 compatible API inhibits adoption of the Cloudant service. 

_How:_

A new private module, _2to3.py, was added containing compatibility methods.  py2 or py3 unique types and methods were replaced throughout the code base with equivalents from the compatibility module.  Python 2 is treated as the special case given it's unique semantics are "deprecated" and py3 and onward is the new normal (all hail the BDFL).  

_Tests:_

All test cases under the tests/ directory were updated to function under py2 and py3 with similar compatibility methods.  All unit tests (tests/unit/*) were run and pass with py2.7 and py3.5 using a local couchdb instance.  The 85 cloudant specific test cases have not yet been run to my knowledge.  The Travis CI environment only executes a py2.7 test -- a py3.5 test should be added before a final merge is complete (I'm not sure how to do it).

_Issues:_

#23 

reviewer @alfinkel
reviewer @evansde77

### DIFF
--- a/src/cloudant/_2to3.py
+++ b/src/cloudant/_2to3.py
@@ -12,67 +12,72 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-python 2 to 3 compatibility methods
+Python 2 to 3 compatibility methods
+
+The philosophy employed here is to treat py2 as the special case vs. py3 as
+future Python releases presumably will retain new semamtics in py3.
 """
 import sys
 
 PY2 = sys.version_info[0] < 3
 ENCODING = 'utf-8'
 NONETYPE = type(None)
+STRTYPE = basestring if PY2 else str  # pylint: disable=undefined-variable
+UNITYPE = unicode if PY2 else str  # pylint: disable=undefined-variable
 
-# pylint: disable=undefined-variable
-STRTYPE = basestring if PY2 else str
+if PY2:
+    def iteritems_(adict):
+        """
+        iterate dict key, value tuples in a py2 and 3 compatible way
 
-# pylint: disable=undefined-variable
-UNITYPE = unicode if PY2 else str
+        :param dict adict:
+        :return: iterator of (key, value) tuples
+        """
+        return adict.iteritems()
 
+    def next_(itr):
+        """
+        return next item from an iterable is a py2 and 3 compatible way
 
-def iteritems_(adict):
-    """
-    py2 to py3 helper
+        :param Iterable itr:
+        :return: the next item in itr
+        """
+        return itr.next()
+else:
+    def iteritems_(adict):
+        """
+        iterate dict key, value tuples in a py2 and 3 compatible way
 
-    :param dict adict:
-    :return:
-    """
-    return adict.iteritems() if PY2 else adict.items()
+        :param dict adict:
+        :return: iterator of (key, value) tuples
+        """
+        return adict.items()
 
+    def next_(itr):
+        """
+        return the next item in an iterable in a py2 and 3 compatible way
 
-def unicode_(astr):
-    """
-    py2 to py3 helper
-
-    :param str astr:
-    :return:
-    """
-    # pylint: disable=undefined-variable
-    return unicode(astr) if PY2 else astr
+        :param Iterable itr:
+        :return: the next item in itr
+        """
+        return next(itr)
 
 
 def bytes_(astr):
     """
-    py2 to py3 helper
+    return a bytes representation of astr in a py2 and 3 compatible way
 
     :param str astr:
-    :return:
+    :return: bytes object
     """
     return astr.encode(ENCODING) if hasattr(astr, 'encode') else astr
 
 
-def str_(astr):
+def unicode_(astr):
     """
-    py2 to py3 helper
+    return a unicode string representation of astr in a py2 and 3 compatible way
 
     :param bytes astr:
-    :return:
+    :return: unicode string
     """
     return astr.decode(ENCODING) if hasattr(astr, 'decode') else astr
-
-
-def next_(itr):
-    """
-    py2 to py3 helper
-
-    :param itr:
-    :return:
-    """
-    return itr.next() if PY2 else next(itr)
--- a/src/cloudant/_2to3.py
+++ b/src/cloudant/_2to3.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2015 IBM. All rights reserved.
+# Copyright (c) 2016 IBM. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,10 +22,19 @@ import sys
 PY2 = sys.version_info[0] < 3
 ENCODING = 'utf-8'
 NONETYPE = type(None)
-STRTYPE = basestring if PY2 else str  # pylint: disable=undefined-variable
-UNITYPE = unicode if PY2 else str  # pylint: disable=undefined-variable
+
+# pylint: disable=undefined-variable
+STRTYPE = basestring if PY2 else str
+
+# pylint: disable=undefined-variable
+UNITYPE = unicode if PY2 else str
+
 
 if PY2:
+    # pylint: disable=wrong-import-position,no-name-in-module,import-error,unused-import
+    from urllib import quote as url_quote, quote_plus as url_quote_plus
+    from ConfigParser import RawConfigParser
+
     def iteritems_(adict):
         """
         iterate dict key, value tuples in a py2 and 3 compatible way
@@ -44,6 +53,9 @@ if PY2:
         """
         return itr.next()
 else:
+    from urllib.parse import quote as url_quote, quote_plus as url_quote_plus  # pylint: disable=wrong-import-position,no-name-in-module,import-error,ungrouped-imports
+    from configparser import RawConfigParser  # pylint: disable=wrong-import-position,no-name-in-module,import-error
+
     def iteritems_(adict):
         """
         iterate dict key, value tuples in a py2 and 3 compatible way
--- a/src/cloudant/_2to3.py
+++ b/src/cloudant/_2to3.py
@@ -17,6 +17,7 @@ python 2 to 3 compatibility methods
 import sys
 
 PY2 = sys.version_info[0] < 3
+ENCODING = 'utf-8'
 NONETYPE = type(None)
 
 # pylint: disable=undefined-variable
@@ -54,7 +55,17 @@ def bytes_(astr):
     :param str astr:
     :return:
     """
-    return astr.encode('utf-8') if hasattr(astr, 'encode') else astr
+    return astr.encode(ENCODING) if hasattr(astr, 'encode') else astr
+
+
+def str_(astr):
+    """
+    py2 to py3 helper
+
+    :param bytes astr:
+    :return:
+    """
+    return astr.decode(ENCODING) if hasattr(astr, 'decode') else astr
 
 
 def next_(itr):
--- a/src/cloudant/changes.py
+++ b/src/cloudant/changes.py
@@ -19,7 +19,7 @@ changes-like feeds.
 
 import json
 
-from ._py2to3 import next_
+from ._2to3 import next_, str_
 
 class Feed(object):
     """
@@ -96,7 +96,7 @@ class Feed(object):
         if len(line.strip()) == 0:
             return {}
         try:
-            data = json.loads(line)
+            data = json.loads(str_(line))
         except ValueError:
             data = {"error": "Bad JSON line", "line": line}
 
--- a/src/cloudant/changes.py
+++ b/src/cloudant/changes.py
@@ -21,7 +21,6 @@ import json
 
 from ._2to3 import next_, unicode_
 
-
 class Feed(object):
     """
     Provides an infinite iterator for consuming database feeds such as
--- a/src/cloudant/changes.py
+++ b/src/cloudant/changes.py
@@ -19,7 +19,8 @@ changes-like feeds.
 
 import json
 
-from ._2to3 import next_, str_
+from ._2to3 import next_, unicode_
+
 
 class Feed(object):
     """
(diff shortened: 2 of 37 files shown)


rredburn:
ok, so pylint with py2.7 fails on py3.5 conditional imports.  I was able to get pylint/3.5 to ignore the py2 imports with local overrides, but pylint/2.7 won't take them for some reason.  
 

> $ pylint --rcfile=../pylintrc cloudant
> ************* Module cloudant.credentials
> E: 28, 4: Unable to import 'configparser' (import-error)
> ************* Module cloudant.database
> E: 31, 4: No name 'parse' in module 'urllib' (no-name-in-module)
> E: 31, 4: Unable to import 'urllib.parse' (import-error)
> ************* Module cloudant.document
> E: 30, 4: No name 'parse' in module 'urllib' (no-name-in-module)
> E: 30, 4: Unable to import 'urllib.parse' (import-error)

For example, configparser in credentials is overridden but seemingly ignored by pylint/2.7.  Anyone know the secret sauce?

```python
# pylint: disable=wrong-import-order
from ._2to3 import PY2
if PY2:
    # pylint: disable=wrong-import-order
    from ConfigParser import RawConfigParser
else:
    # pylint: disable=wrong-import-order, import-error
    from configparser import RawConfigParser
`

**Comment by rredburn** [comment]
found a fix with line local vs. block pylint overrides.  First time I've seen Travis integration on a Python library, pretty slick.   You'll have to add a py3 check to it.

**Comment by alfinkel** [comment]
@rredburn - At first glance this looks great.  I just sent you an internal email outlining what needs to happen in order for you to be able to contribute to this repo.  It is a fairly simple process.

**Comment by alfinkel** [comment]
@rredburn:  Can you edit your original comment to follow our PR submission process initial comment.  Take a look at https://github.com/cloudant/python-cloudant/pull/89 as a reference.  In your edit reference issue #23, which is the Python3 support issue number, in your comment and add me as a reviewer.  Once I have a look and provide comments, I'll get one other member from our team to take a look as well.

File: src/cloudant/document.py [hunk 1]
```diff
@@ -17,13 +17,19 @@
 """
 import json
 import posixpath
-import urllib
 import requests
-
 from requests.exceptions import HTTPError
 
+# pylint: disable=wrong-import-order
```
evansde77: given this stanza appears a couple of times and I think string quoting is going to be used in other places, may be cleaner to set up a quote_plus_ style symbol in the _2to3 module.

evansde77:
This looks good to me, would suggest dropping a python 3 build in the .travis.yml on this branch and letting travis give it a wallop as well. 
(See https://github.com/cloudant/python-cloudant/blob/master/.travis.yml#L4) 
Eg: 
https://github.com/evansde77/pluggage/blob/develop/.travis.yml#L5

The suggestion about defining a quote_plus_ in _2to3 is at your discretion, but I would prefer to keep the compat stuff in one place and avoid copypasta in other modules. 

+1

alfinkel:
Add the python version `- "3.5"` after the `- "2.7"`entry in the travis.yml (https://github.com/cloudant/python-cloudant/blob/master/.travis.yml#L4) as @evansde77 suggests.  As in:

```
python:
  - "2.7"
  - "3.5"
```

This way we can see how Travis likes it.

alfinkel:
It would also be great if you could confirm that the Cloudant specific tests also work in Python 3.  You can follow instructions [here](https://github.com/cloudant/python-cloudant/blob/master/tests/unit/db/unit_t_db_base.py#L22).  Basically it is a matter of setting 4 environment variables and kicking off the tests as you would normally.

src/cloudant/_2to3.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2015 IBM. All rights reserved.
```
alfinkel: Change to 2016

src/cloudant/changes.py
```diff
@@ -19,6 +19,9 @@
 
 import json
 
+from ._2to3 import next_, unicode_
+
+
```
alfinkel: Remove one blank line.

src/cloudant/database.py
```diff
@@ -18,9 +18,17 @@
...
+
 from requests.exceptions import HTTPError
 
+# pylint: disable=wrong-import-order
+from ._2to3 import PY2
+if PY2:
+    # pylint: disable=wrong-import-order,no-name-in-module
+    from urllib import quote_plus
+else:
+    from urllib.parse import quote_plus  # pylint: disable=import-error,no-name-in-module
```
alfinkel: Can you put the pylint disable comment above the statement?  This more closely matches the coding standard that we've adopted.

src/cloudant/database.py
```diff
@@ -586,11 +590,8 @@ def missing_revisions(self, doc_id, *revisions):
         url = posixpath.join(self.database_url, '_missing_revs')
         data = {doc_id: list(revisions)}
 
-        resp = self.r_session.post(
-            url,
-            headers={'Content-Type': 'application/json'},
-            data=json.dumps(data)
-        )
+        resp = self.r_session.post(url, headers={'Content-Type': 'application/json'},
+                                   data=json.dumps(data))
```
alfinkel: Can you return the statement to its original form?  Again, the coding standard we've adopted.

src/cloudant/database.py
```diff
@@ -615,11 +616,8 @@ def revisions_diff(self, doc_id, *revisions):
         url = posixpath.join(self.database_url, '_revs_diff')
         data = {doc_id: list(revisions)}
 
-        resp = self.r_session.post(
-            url,
-            headers={'Content-Type': 'application/json'},
-            data=json.dumps(data)
-        )
+        resp = self.r_session.post(url, headers={'Content-Type': 'application/json'},
+                                   data=json.dumps(data))
```
alfinkel: Return statement to original form.

src/cloudant/document.py
```diff
@@ -17,13 +17,19 @@
...
+from ._2to3 import PY2, unicode_
+if PY2:
+    from urllib import quote, quote_plus  # pylint: disable=no-name-in-module
+else:
+    from urllib.parse import quote, quote_plus  # pylint: disable=import-error,no-name-in-module
```
alfinkel: Both pylint disable comments should be above their referenced statements.

src/cloudant/replicator.py
```diff
@@ -163,14 +163,11 @@ def update_state():
...
-                state = repl_doc.get('_replication_state')
+                arepl_doc = self.database[repl_id]
+                arepl_doc.fetch()
+                return arepl_doc, arepl_doc.get('_replication_state')
             except KeyError:
-                repl_doc = None
-                state = None
-
-            return repl_doc, state
+                return None, None
```
alfinkel: Any reason why you chose to rewrite this try block?  It doesn't appear to have any py2 v py3 issues.  Also why the renaming of `repl_doc` to `arepl_doc`?

src/cloudant/result.py
```diff
@@ -360,10 +360,11 @@ def __getitem__(self, key):
 
         :returns: Rows data in JSON format
         """
-        if 'skip' in self.options or 'skip' in self._ref.keys():
+        refkeys = set(self._ref.keys())
+        if 'skip' in self.options or 'skip' in refkeys:
             msg = 'Cannot use skip parameter with QueryResult slicing.'
             raise CloudantArgumentError(msg)
-        if 'limit' in self.options or 'limit' in self._ref.keys():
+        if 'limit' in self.options or 'limit' in refkeys:
```
alfinkel: Why the introduction of `refkeys` in this series of if statements?  This does not seem related to adding py3 support.

tests/unit/db/account_tests.py
```diff
@@ -232,14 +234,14 @@ def test_delete_remote_db_via_delitem(self):
...
-            db = self.client.create_database(dbname)
+            _ = self.client.create_database(dbname)
             self.assertIsNotNone(self.client.get(dbname))
             self.client.__delitem__(dbname, remote=True)
             # Removed from local cache
             self.assertIsNone(self.client.get(dbname))
             # Database removed remotely as well
             try:
-                db = self.client[dbname]
+                _ = self.client[dbname]
```
alfinkel: Why change `db` to `_`?

tests/unit/db/database_tests.py
```diff
@@ -315,7 +317,7 @@ def test_get_non_existing_doc_via_getitem(self):
         Test __getitem__ when retrieving a non-existing document
         """
         try:
-            doc = self.db['no_such_doc']
+            _ = self.db['no_such_doc']
```
alfinkel: Same question as before.  Why `doc` to `_`?

tests/unit/db/unit_t_db_base.py
```diff
@@ -113,24 +117,22 @@ def setUp(self):
...
-        		self.pwd,
-        		url=self.url,
-        		x_cloudant_user=self.account
-        		)
+            self.account = os.environ.get('CLOUDANT_ACCOUNT')
+            self.user = os.environ.get('DB_USER')
+            self.pwd = os.environ.get('DB_PASSWORD')
+            self.url = os.environ.get(
+                    'DB_URL',
+                    'https://{0}.cloudant.com'.format(self.account))
```
alfinkel: I guess these are also over indented.

tests/unit/db/unit_t_db_base.py
```diff
@@ -113,24 +117,22 @@ def setUp(self):
...
+            self.client = Cloudant(
+                    self.user,
+                    self.pwd,
+                    url=self.url,
+                    x_cloudant_user=self.account)
```
alfinkel: Same, over indented.

src/cloudant/database.py
rredburn: That does not work with pylint 1.6/python 2.7, or at least, I could not get it to work.  pylint would only pass with the override on the same line (a "line" override) vs. above (a "block" override).  The preferred "block" style override works fine with Python 3.5.  I noted this in prior comments -- if you know how to make it work I'm happy to try.  I've moved all PY2 conditionals to _2to3 based on Dave's feedback so this exact instance is gone at least.

src/cloudant/database.py
rredburn: The problem is we either have to relax your pylintrc constraint of 1000 lines per file or, as I've done here, remove newlines.  I did not want to arbitrarily relax your testing standards so opted to flatten a few method calls as noted in my original comments.  Please advise.

src/cloudant/document.py
rredburn: agreed, all PY2 conditional imports have been moved to _2to3 module as suggested.

src/cloudant/replicator.py
rredburn: Pycharm and I found the scope a little unnerving as repl_doc is used in the outer parent function (review the larger context).  Changing the inner function variable to arepl_doc removes potential confusion on if you intend to manipulate the outer variable or create a new one.  I will undo the edit if desired.

src/cloudant/result.py
rredburn: in py3, keys() returns an iterator which would then have to be forced into list form for "in" testing.  Rather than do that twice back to back, I chose to do it once and make it a set to improve performance, ie. fewer list constructions and an O(1) vs. O(n) membership test.

tests/unit/db/account_tests.py
rredburn: ah, remove complaints about unused variables to declutter pycharm warnings.  I can revert or simply remove the assignment altogether, let me know.

src/cloudant/database.py
alfinkel: Fair enough.  pylint is often a pain.

src/cloudant/database.py
alfinkel: Sorry, I had no idea that we exceeded 1000 lines.  I think it would be fine if you upped the `max-module-lines` to 2000 in the [pylintrc](https://github.com/cloudant/python-cloudant/blob/master/pylintrc#L209).  Most of the lines in the modules are docstring anyway.

src/cloudant/replicator.py
alfinkel: Good point.  Leave it as you have it.

src/cloudant/result.py
alfinkel: Thanks for the clarification.

tests/unit/db/account_tests.py
alfinkel: We've been a bit more lax with the test code as far as standards go.  I had left the `db` in there for it to be a bit clearer as to the purpose of the test.  

I guess my main concern initially was that some changes that were being made in this PR did not seem like they were py2 to py3 related. After reading your responses to some of the earlier comments I realize now that that is not the case. If you ran pycharm on the entire testing code base then obviously you made this type of "fix" everywhere pycharm had a similar issue so I am fine with you leaving this as it is throughout.

alfinkel:
@rredburn:  Thanks for the clear responses.  I think we're pretty close.  We have a few topics left to address:

- Some minor formatting comments.
- Adding the Python 3 test to Travis (https://github.com/cloudant/python-cloudant/pull/91#issuecomment-178658173)
- Running the tests against Cloudant. (https://github.com/cloudant/python-cloudant/pull/91#issuecomment-178659909)
- Upping the limit on module size for pylint (https://github.com/cloudant/python-cloudant/pull/91#discussion_r51758450)

rredburn:
@alfinkel, @evansde77  :  I believe I've addressed all comments to date.  The latest commit includes:

* all requested code edits
* Python 3.5 Travis check,  passes both 2.7 and 3.5
* Cloudant tests run and pass both 2.7 and 3.5.  One thing to note, I did get infrequent, seemingly random, unrepeatable fails when testing against a free Cloudant 30 day trial account.  Such fails occurred in both 2.7 and 3.5 and have no relation to Python code issues that I see.  I believe they are due to race conditions occurring with unreliable network or shared server performance variations (replication tests in particular seem to pass or hang/timeout randomly between consecutive runs).  It certainly wouldn't hurt to run the tests in a dedicated and reliable environment to confirm no issues.

Please review and mark additional edits as needed.  Thanks.

alfinkel:
I think the ball is in my court now.  I have a few tests that I want to run before moving forward.

alfinkel:
+1...

I've merged the contents of the commits from this PR branch over top of a rebased branch in this repo and issued PR #97.  Closing this one.

Issue #52, opened 2015-12-15
Cannot update _security document using Cloudant python


mikerhodes:
You cannot update a _security document using python-cloudant. If you run the following script:

```python
from cloudant.account import Cloudant

client = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME)
client.connect()
session = client.session()
database = client[DATABASE_NAME]
security_document = database[DOC_NAME]
print "Got security_document: ", security_document

# Update the security_document with the API key permissions
security_document[u'cloudant'][u'bouninamendouldnimendepa'] = [u'_reader']

security_document.save()

client.disconnect()
```

You get a problem because `_security` documents do not have `_rev` fields (they are not versioned in the same way):

```
Traceback (most recent call last):
  File "main.py", line 23, in <module>
    security_document.save()
  File "/Users/mike/scratch/python-cloudant/issue-49/lib/python2.7/site-packages/cloudant/document.py", line 172, in save
    super(Document, self).__setitem__('_rev', data['rev'])
KeyError: 'rev'
```

I should've spotted this issue; it's probably an issue elsewhere too.

alfinkel:
@mikerhodes we are not wrapping `_security` documents as `Document` for that very reason.  For now we are using the share_database and unshare_database methods to manage permissions for users and API keys.  So the `_security` document would be updated that way.  I suppose that I see your point in that you want to be able to handle the `_security` document as a Document so we can let this issue track that enhancement.  As for the "rev" finding it should not be an issue other than in this case but again the `_security` document was not meant to be wrapped as a Document when this library was initially written.

alfinkel:
@mikerhodes - I spoke to @evansde77 about setting permissions and the sharing databases.  I've summarized the major points of our conversation below:

- Using the _api/v2/db/$db/_security endpoint (as is used in the share_database method of this library) accomplishes two things when sharing a database.
  1. Sets permissions in the database _security document for the user being granted "shared" access.
  2. Adds the database to the list of shared databases for the user to that user's user document (done via wilson).  Note:  We believe this to be the case.  I will confirm and make a note here if this is NOT the case.  Basically, this is how the user gets alerted to the fact that a database has been shared with them via the dashboard.
- Using the _security endpoint would only perform the setting of permissions in the _security document.  This means a couple of things to us:
[...]

PR #93, opened 2016-02-19
allowed replicator permission to be set in share_database()

jlstack:


alfinkel:
@jlstack, thanks for putting in the effort but unfortunately this change is part of a larger effort detailed in issue #52.  The security apparatus in this library is currently very immature and needs to be beefed up.  This includes adding in all of the roles documented [here](https://docs.cloudant.com/authorization.html#roles) along with a number of other things.  I think that adding the change in this PR to the library does not completely address this library's security question as is detailed in issue #52.  So based on that I think it makes sense to close this PR and let us handle the entire security question which will likely effect how the `CloudantDatabase.share_database` method functions going forward.  

My suggestion in the interim, while we get security sorted out, is to use the _security end point directly as is documented in the Cloudant docs [here](https://docs.cloudant.com/authorization.html) along with this library's [end point access](http://python-cloudant.readthedocs.org/en/latest/getting_started.html#end-point-access) section of the Getting Started portion of this library's documentation.

Issue #85, opened 2016-01-08
Fix DesignDocument so that an empty view sub-document is not created


alfinkel:
When instantiating a DesignDocument, an empty views sub-document is created by default.  If no view is added to the DesignDocument and that DesignDocument is then created/saved to the remote server, that empty views sub-document is written to the DesignDocument.  This should not be the case.  For example:

```python
with DesignDocument(db, '_design/ddoc001x') as ddoc:
    ddoc['foo'] = 'bar'
print db['_design/ddoc001x']
```

will output:

`{u'_rev': u'1-95205e2e4d80983f11db5f6018adb430', u'_id': u'_design/ddoc001x', u'foo': u'bar', u'views': {}}`

This means that the design document saved to the database contains `'views': {}` which it should not.

- Fix the DesignDocument.save() method to check for an empty views sub-document and to remove it before saving it.
- One thing to keep in mind is that removing the "views" key from the DesignDocument object could cause unwanted side effects and introduce new bugs, so a poosible solution would be to copy the DesignDocument content and update that instead as part of the save() method and then updating the original DesignDocument with the new "_rev"
- This bug is somewhat related to #66 and should be resolved before or as part of that bug fix.

PR #99, opened 2016-02-25
Fix DesignDocument to handle empty views appropriately

alfinkel:
## What

Updated the DesignDocument class to ensure that empty views are handled consistently.

## Why

The previous behavior of a DesignDocument was not consistent w.r.t. empty views.  For example when constructing a DesignDocument an empty view dictionary is added to the DesignDocument object.  This dictionary is meant for use by view management within the DesignDocument.  However, if that DesignDocument was saved to the remote database and that views dictionary was still empty then that empty view dictionary would wind up as part of the design document in the remote database which was incorrect functionality.

## How

- Updated the DesignDocument.save() method to strip an empty views dict from the DesignDocument before performing the remote save and then returning the empty views dict after the save returned.
- Updated the DesignDocument.fetch() method to add an empty views dict if one was not found as part of the fetch from remote.

## Testing

- Fixed database, index, and design document tests to work with an empty views dict.
- Added design document tests to verify the save and fetch changes.

reviewer @ricellis 
reviewer @rhyshort 

## Issues

- #85


ricellis:
+1

rhyshort:
+1

Issue #66, opened 2015-12-24
Fix CouchDatabase.create_document to handle DesignDocuments appropriately


alfinkel:
- Currently `database.CouchDatabase.create_document()` creates a Document regardless of whether the document is a `Document` or a `DesignDocument`.  

- We need to add logic to ensure that if the "_id" field passed in as part of the data starts with '_design/' then this method should create a `DesignDocument` object and that DesignDocument object should be attached to the locally cached database dictionary.

PR #101, opened 2016-02-26
66 fix create doc to handle ddocs

alfinkel:
## What

Fix the CouchDatabase.create_document method to handle design documents as well as documents.

## How

Add a check to see if the data being used to create the document contains an `_id` that begins with `_design/`.  If it does then a DesignDocument object is passed back by the method and is attached to the locally cached database object.  Otherwise a Document object will be passed back and cached.

## Testing

- Add assertions to existing tests to verify that a Document object is returned and locally cached when expected.
- Add a test to verify that a DesignDocument object is returned and locally cached when expected.

## Reviewers

reviewer @rhyshort
reviewer @emlaver 

## Issues

- #66


emlaver:
+1

rhyshort:
+1

Issue #103, opened 2016-03-02
Refactor SearchIndex as TextIndex


alfinkel:
When adding Cloudant Query functionality we mistakenly used the term for a search index to refer to a Cloudant Query text index.  In reality, a search index refers to a completely different entity and support for  Cloudant search (_search endpoint) will need to be added as a separate issue #104.

This issue is to track the refactoring of all code referencing a SearchIndex to a TextIndex.

PR #105, opened 2016-03-02
Refactor Cloudant Query index class and DB methods

alfinkel:
## What

We need to refactor the current SearchIndex class to be named TextIndex since it really refers to Cloudant Query text index and not a Cloudant Search index.  Also, we need to provide clearer naming for the database convenience methods of get_all_indexes, create_index, and delete_index to specifically reference the fact that these methods are meant for Cloudant Query (CQ) index management only.

## Why

The current name of SearchIndex is inaccurate as it really handles Cloudant Query text indexes and not actual Cloudant Search indexes and the database convenience methods used to manage the Cloudant Query indexes need clearer naming for the same reason.

## How

- Rename SearchIndex to TextIndex and update all references to the class.
- Rename the get_all_indexes method of the CloudantDatabase class to be get_cq_indexes since the returned set of indexes are specific to Cloudant Query and do not necessarily include "all" indexes in the database.
- Rename the create_index method of the CloudantDatabase class to be create_cq_index since this method can only be used to create a Cloudant Query index.
- Rename the delete_index method of the CloudantDatabase class to be delete_cq_index since this method can only be used to delete a Cloudant Query index.

## Testing

- Renamed all references affected by the refactoring.  No new tests needed.

## Reviewers

reviewer: @emlaver 
reviewer @rhyshort
## Issues

- #103

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,7 @@
 - [FIX] Fixed the handling of empty views in the DesignDocument.
 - [BREAKING] Fixed CloudantDatabase.share_database to accept all valid permission roles.  Changed the method signature to accept roles as a list argument.
 - [FIX] The CouchDatabase.create_document method now will handle both documents and design documents correctly.  If the document created is a design document then the locally cached object will be a DesignDocument otherwise it will be a Document.
-- [BREAKING] Refactored the SearchIndex class to now be the TextIndex class.  Also renamed the CloudantDatabase convenience methods of get_all_indexes, create_index, and delete_index as get_cq_indexes, create_cq_index, and delete_cq_index respectively.  These changes were made to clarify that the changed class and the changed methods were specific to Cloudant Query index processing only.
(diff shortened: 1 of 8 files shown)


emlaver:
LGTM +1

rhyshort:
With CQ now contributed to Apache CouchDB as Mango, I wonder if we should call it something generic like Query like [ObjectiveCloudant](https://github.com/cloudant/objective-cloudant), or make direct references to mango. Although the library is for Cloudant, we do claim compatibility with CouchDB after all.

alfinkel:
Changed method names from get_cq_indexes, create_cq_index, and delete_cq_index to get_query_indexes, create_query_index, and delete_query_index.  Done in https://github.com/cloudant/python-cloudant/commit/eb3ff2fadc25f9eed670d24cb09428cd3f158981.

rhyshort:
+1

Issue #61, opened 2015-12-24
Add support for CouchDB Admin Party mode


alfinkel:
Currently you must have security set up in CouchDB in order to use the lib with CouchDB.  Since often CouchDB is set up with "Admin Party" privileges it would be good for the lib to support that as well.

PR #107, opened 2016-03-07
Add support for CouchDB Admin Party mode

alfinkel:
## What

Add support for CouchDB Admin Party mode where everyone is an Admin.

## Why

The default install of CouchDB uses Admin Party mode where everyone is Admin.  We needed to enable this library to work in this mode.

## How

- Set CouchDB constructor user and auth_token arguments to default to None.
- Add an admin_party property to the CouchDB class that returns true if both user and auth_token are not defined. False otherwise.
- Add logic to handle processing differently for Admin Party in the CouchDB class. Methods include connect(), session(), session_cookie(), session_login(), session_logout(), and basic_auth_str().
- Add logic to handle processing differently for Admin Party in the CouchDatabase class, creds() method.
- Add logic to handle processing differently for Admin Party in the Replicator class, create_replication(...) method.

## Testing

- Update tests to handle both Admin Party and normal security.  No new tests added.
- Added the ADMIN_PARTY environment variable to the .travis.yml so that tests can now run against both Python 2.7 and Python 3.5 using CouchDB in Admin Party mode and CouchDB using standard security.

## Reviewers

reviewer: @emlaver 
reviewer: @evansde77 

## Issues

- #61

### DIFF
--- a/src/cloudant/account.py
+++ b/src/cloudant/account.py
@@ -46,33 +46,23 @@ class CouchDB(dict):
     :param str user: Username used to connect to CouchDB.
     :param str auth_token: Authentication token used to connect to CouchDB.
     :param str url: URL for CouchDB server.
+    :param bool admin_party: Setting to allow the use of Admin Party mode in
+        CouchDB.  Defaults to ``False``.
     :param str encoder: Optional json Encoder object used to encode
         documents for storage.  Defaults to json.JSONEncoder.
     """
     _DATABASE_CLASS = CouchDatabase
 
-    def __init__(self, user=None, auth_token=None, **kwargs):
+    def __init__(self, user, auth_token, admin_party=False, **kwargs):
         super(CouchDB, self).__init__()
         self._cloudant_user = user
         self._cloudant_token = auth_token
         self._cloudant_session = None
         self.cloudant_url = kwargs.get('url')
         self._cloudant_user_header = None
+        self.admin_party = admin_party
         self.encoder = kwargs.get('encoder') or json.JSONEncoder
         self.r_session = None
-
-    @property
-    def admin_party(self):
-        """
-        If neither a username or a password have been provided during object
-        construction then CouchDB Admin Party mode is assumed.
-
-        :returns: Boolean value showing whether a username and password have
-            been provided
-        """
-        if self._cloudant_user is None and self._cloudant_token is None:
-            return True
-        return False
 
     def connect(self):
         """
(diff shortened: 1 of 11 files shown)


src/cloudant/replicator.py
```diff
@@ -77,31 +80,30 @@ def create_replication(self, source_db=None, target_db=None,
         if not data.get('source'):
             if source_db is None:
                 raise CloudantException(
-                    "You must specify either a source_db Database "
-                    "object or a manually composed 'source' string/dict."
+                    'You must specify either a source_db Database '
+                    'object or a manually composed \'source\' string/dict.'
+                )
+            data['source'] = {'url': source_db.database_url}
+            if not source_db.cloudant_account.admin_party:
```
evansde77: I think I would maybe proxy this property with a property on the database object:

```python
@property
 def admin_party(self):
       return self.cloudant_account.admin_party 

```

save a bit of boilerplate

src/cloudant/account.py
```diff
@@ -61,16 +61,30 @@ def __init__(self, user, auth_token, **kwargs):
...
+    @property
+    def admin_party(self):
+        """
+        If neither a username or a password have been provided during object
+        construction then CouchDB Admin Party mode is assumed.
+
+        :returns: Boolean value showing whether a username and password have
+            been provided
+        """
+        if self._cloudant_user is None and self._cloudant_token is None:
```
evansde77: so what happens if the user just fails at passing user/auth_token here? It will assume the db is in admin party mode, which it may be but it may not be. 
I think perhaps the ctor args would be better done by still requiring user and auth_token and adding an admin_party=False. That way the user is clear about setting admin_party and makes it a special case, rather than making is possible for a new user to do the wrong thing because they arent passing any arguments. 

If the user=None, auth_token=None, admin_party=True is too ugly (yeah, its pretty gross) add a helper context:
```python
with couchdb_admin_party(args here):
     # do stuff
```

evansde77:
couple of suggestions:
1. Make admin party more of a special case, rather than normal use (which it is) and reduce the potential for mistakes by not passing user/auth_token and getting a weird admin party/auth error. I threw out a couple of ideas
2. add an admin_party prop to the database class so that you reduce boiler plate. You access it in a few places so this makes sense. 

Everything else LGTM, would really like to see 1. get addressed though

src/cloudant/replicator.py
alfinkel: Done in d14d22a.

src/cloudant/account.py
alfinkel: Fixed in d14d22a.  Also added context manager for couchdb_admin_party.

emlaver:
Looks good +1

Issue #90, opened 2016-01-22
Change view(ddoc,view, keys=[key1,key2]) to use JSON data for keys instead of encoding in URL


bradwbonn:
Right now, the number of keys you can get from a view in the library is limited by the maximum length of the URL, which is woefully small.  Can this be changed to execute a GET or POST with -d {keys: [key1,key2...]} ?

alfinkel:
Sure this makes sense.  We need to add logic [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/views.py#L246) to check for the existence of a "keys" parameter.  If one exists then execute the request as a POST otherwise exercise the request as a GET.

bradwbonn:
Should probably also make sure all other methods pass the JSON data instead of append keys in the URL.  Such as database.all_docs()

alfinkel:
Good point - we'll need to add the same logic [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L379) as is described [above](https://github.com/cloudant/python-cloudant/issues/90#issuecomment-173986454) and also check the the entire library code base for these sort of GET calls to ensure that this is covered throughout.  I don't think that there are many/any other instances of this but I could be wrong.

PR #108, opened 2016-03-10
Use POST request with queries using the keys parameter

alfinkel:
## What

Add logic so that any query using the `keys` parameter is handled via a http POST request instead of a http GET request.

## How

- Update result module python_to_couch function so that it handles the keys argument separately.
- Update the CouchDatabase all_docs method logic to handle any call containing the keys argument as a http POST request.
- Update the Views \_\_call\_\_ method logic to handle any call containing the keys argument as a http POST request.

## Testing

- Add tests to both database_tests.py and view_tests.py to check that the POST request processing is working as expected.  Tests were added to include a list of keys that would normally cause a similar GET request to fail due to url size limitations.
- The current pylint min-similarity-lines setting was increased from 4 to 5 since the 4 setting was incorrectly causing the pylint test to fail due to the recent code commits.  4 similar lines is too few lines to check for so we are relaxing the check to 5 lines. 

## Reviewers

reviewer: @emlaver 
reviewer: @ricellis 

## Issues

#90

### DIFF
--- a/tests/unit/db/database_tests.py
+++ b/tests/unit/db/database_tests.py
@@ -299,19 +299,47 @@ class DatabaseTests(UnitTestDbBase):
         self.assertIsInstance(raw_rslt, dict)
         self.assertEqual(len(raw_rslt.get('rows')), 100)
 
-    def test_all_docs(self):
+    def test_all_docs_post(self):
         """
-        Test the all_docs functionality
+        Test the all_docs POST request functionality using keys param
+        """
+        # Create 200 documents with ids julia000, julia001, julia002, ..., julia199
+        self.populate_db_with_documents(200)
+        # Generate keys list for every other document created
+        # with ids julia000, julia002, julia004, ..., julia198
+        keys_list = ['julia{0:03d}'.format(i) for i in range(0, 200, 2)]
+        self.assertEqual(len(keys_list), 100)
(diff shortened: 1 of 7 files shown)


emlaver:
LGTM +1

tests/unit/db/database_tests.py
```diff
@@ -299,19 +299,28 @@ def test_retrieve_raw_view_results(self):
         self.assertIsInstance(raw_rslt, dict)
         self.assertEqual(len(raw_rslt.get('rows')), 100)
 
-    def test_all_docs(self):
+    def test_all_docs_post(self):
         """
-        Test the all_docs functionality
+        Test the all_docs POST request functionality
+        """
+        keys_list = list(map(lambda x: x['id'], self.populate_db_with_documents()))
```
ricellis: IIUC this keys list is effectively all 100 test documents. I think this means the assertions that this test makes would work even if the keys argument was completely ignored.
It might be worth having an additional test (or changing the list of keys in this test) to use a sub-set to validate that the keys argument it is actually being used.

alfinkel: - Now using a subset of keys for the documents in the test database.
- Also added tests for POST with `keys` alone and POST with `keys` and other arguments.

Done in 82e9646

ricellis:
+1

Issue #43, opened 2015-12-08
Combine get_view_result and get_view_raw_result into one method


alfinkel:
We should merge the database method `get_view_raw_result` into `get_view_result` and pass in a parameter of `raw_result=False` as default to replicate the behavior we recently implemented for query in the database convenience method `get_query_result`.  See https://github.com/cloudant/python-cloudant/blob/51282-add-query-support/src/cloudant/database.py#L795

PR #118, opened 2016-03-18
Combine get_view_result and get_view_raw_result into one method

emlaver:
## What

Combine `get_view_result` and `get_view_raw_result` into one method.

## How

The default setting of get_view_result is `raw_result=False` and will return a Result object. 
If the user requires manually changing the paging and iteration, then they would use the setting `raw_result=True` and pass in the skip and/or limit parameter. 

## Testing

The existing test case `test_retrieve_raw_view_results` that used the method `get_view_raw_result` now uses  `get_view_result` with `raw_result=True`.
Tests added to assert that the result contains the correct number of documents using `raw_result=True` and limit, skip, or both parameters.

## Reviewers

reviewer: @alfinkel
reviewer @ricellis

## Issues
#43

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -212,16 +212,19 @@ class CouchDatabase(dict):
 
         return ddoc
 
-    def get_view_result(self, ddoc_id, view_name, **kwargs):
+    def get_view_result(self, ddoc_id, view_name, raw_result=False, **kwargs):
         """
-        Retrieves a Result object based on the design document
-        and view name.  If you intend to iterate through the
-        result, do not use ``skip`` and/or ``limit`` in the kwargs as
-        that is handled in the Result.  If you would like to
-        manage paging and iteration manually over the result
-        then use the
-        :func:`~cloudant.database.CouchDatabase.get_view_raw_result`
-        method instead.
+        Retrieves the view result based on the design document
+        and view name.  By default the result is returned as a
+        :class:`~cloudant.result.Result` which uses the ``skip`` and
+        ``limit`` query parameters internally to handle slicing and iteration
+        through the view result collection.  Therefore ``skip`` and ``limit``
+        cannot be used as arguments to get the view result when
+        ``raw_result=False``.  However, by setting ``raw_result=True``, the
+        result will be returned as the raw JSON response content for the view
+        requested.  Using this setting requires the developer to manage their
+        own slicing and iteration.  Therefore ``skip`` and ``limit`` are valid
+        arguments in this instance.
 
         For example to retrieve the default Result object based on a
         design document view do:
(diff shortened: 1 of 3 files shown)


src/cloudant/database.py
```diff
@@ -238,73 +235,29 @@ def get_view_result(self, ddoc_id, view_name, **kwargs):
...
-        method instead.
-
-        For example to retrieve the raw JSON response content based on a
+        To retrieve the raw JSON response content based on a
         design document view do:
 
         .. code-block:: python
 
-            db.get_view_raw_result('_design/ddoc_id_001', 'view_001')
+            db.get_view_result('_design/ddoc_id_001', 'view_001', True)
```
alfinkel: Change this to `db.get_view_result('_design/ddoc_id_001', 'view_001', raw_result=True)`.  Yes you can set it directly as you have it but since this is an example I think it makes sense to explicitly use the `raw_result` argument name to underscore what is actually happening.

src/cloudant/database.py
```diff
@@ -238,73 +235,29 @@ def get_view_result(self, ddoc_id, view_name, **kwargs):
...
 
         But to retrieve the raw JSON response content based on a set of
         parameters on the same design document view do something like:
 
         .. code-block:: python
 
-            db.get_view_raw_result('_design/ddoc_id_001', 'view_001',
+            db.get_view_result('_design/ddoc_id_001', 'view_001', True,
```
alfinkel: Change this to `db.get_view_result('_design/ddoc_id_001', 'view_001', raw_result=True,`.  Same reason as above.

tests/unit/db/database_tests.py
```diff
@@ -295,10 +295,55 @@ def test_retrieve_raw_view_results(self):
         self.db.create_document(data)
         self.populate_db_with_documents()
 
-        raw_rslt = self.db.get_view_raw_result('_design/ddoc01', 'view01')
+        raw_rslt = self.db.get_view_result('_design/ddoc01', 'view01', True)
```
alfinkel: Use `raw_result=True`.  This will help others in the future to better understand what `True` is referring to when looking at the test and also ensures that the code we suggest in the example where explicitly use `raw_result=True` is correct.

alfinkel:
Can you go back compare the docstring for `get_view_result` with the docstring for `get_query_result`.  From a documentation consistency POV, it would be great if they both sort of mirrored each other.  For example, in `get_query_result` we have:

```
:returns: The result content either wrapped in a QueryResult or
            as the raw response JSON content
```

whereas in `get_view_result` we have:

```
:returns: The result content wrapped in a Result object.
            If ``raw_result=True``, the raw JSON response content for the
            query requested.
```

It would be great if things like that looked consistent across the two methods since they are so similar.  Use `get_query_result` as the template.

alfinkel:
I don't think that I agree with the need to add these three tests: `test_retrieve_raw_view_with_limit_results`, `test_retrieve_raw_view_with_skip_results`, `test_retrieve_raw_view_with_limit_and_skip_results`.

You should be writing tests that test the code that you added/changed directly.  These tests check that the view query execution is correct, which should be/and is handled in another set of tests found in `view_execution_tests.py`.

So I would remove these tests from this PR.

alfinkel:
You missed a reference to the now non-existent `get_view_raw_result` method in the docstring of `View.__call__`. https://github.com/cloudant/python-cloudant/blob/51282-add-query-support/src/cloudant/views.py#L205.

src/cloudant/database.py
emlaver: Fixed in 977492b.

src/cloudant/database.py
emlaver: Fixed in 977492b.

tests/unit/db/database_tests.py
emlaver: Fixed in 977492b.

emlaver:
Commit 977492b contains changes for above comments:
- Fix the reference of `get_view_raw_result` in `View.__call__` 
- Documentation improvements were made to mirror `get_query_result`.
- Test cases were removed.

alfinkel:
+1

ricellis:
+1

Issue #112, opened 2016-03-16
Add query view execution tests


alfinkel:
Add unit tests that check that the translated parameters generated by the python_to_couch function provide the expected query results when used as part of a view query.

Related to #109.

PR #119, opened 2016-03-18
Add view execution tests using translated parameters

alfinkel:
## What

Add view_execution_tests.py module to test that views using converted Python to CouchDB parameters return valid values.

## How

Add tests for each valid parameter contained within the `ARG_TYPES` dictionary to ensure that the converted parameters yield the expected view query result set of data.

## Testing

- Add the view_execution_tests.py module
- Add tests for each valid parameter and value to validate view query functionality using the translated parameters.

## Reviewers

reviewer: @emlaver 
reviewer @ricellis

## Issues

#112

### DIFF
--- a/tests/unit/db/view_execution_tests.py
+++ b/tests/unit/db/view_execution_tests.py
@@ -31,6 +31,7 @@ class QueryParmExecutionTests(UnitTestDbBase):
         """
         super(QueryParmExecutionTests, self).setUp()
         self.db_set_up()
+        self.populate_db_with_documents()
         self.ddoc = DesignDocument(self.db, 'ddoc001')
         self.ddoc.add_view(
             'view001',
@@ -79,7 +80,6 @@ class QueryParmExecutionTests(UnitTestDbBase):
         {'key': 'julia002', 'id': 'julia002', 'value': 1},
         ...
         """
-        self.populate_db_with_documents(100)
         actual = self.view001(descending=True)['rows']
         expected = [{'key': 'julia{0:03d}'.format(x), 
                      'id': 'julia{0:03d}'.format(x),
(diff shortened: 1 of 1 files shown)


tests/unit/db/view_execution_tests.py
```diff
@@ -0,0 +1,628 @@
...
+        Test view query using descending parameter set to True.
+
+        The view used here will generate rows of data where each key will equal
+        the document id.  Such as:
+        {'key': 'julia000', 'id': 'julia000', 'value': 1},
+        {'key': 'julia001', 'id': 'julia001', 'value': 1},
+        {'key': 'julia002', 'id': 'julia002', 'value': 1},
+        ...
+        """
+        self.populate_db_with_documents(100)
```
emlaver: Since `setUp` is run before each test case, why not move `self.populate_db_with_documents` into it?
```
When a setUp() method is defined, the test runner will run that method prior to each test.
```
Supporting link: https://docs.python.org/2.6/library/unittest.html

alfinkel: Done in 509e9da65b329e639c37e9a060d74ddc06c222ef

tests/unit/db/view_execution_tests.py
```diff
@@ -0,0 +1,602 @@
...
+        Test view query using group_level parameter.
+
+        The view used here along with group_level=1 will generate rows of
+        data where each key will equal the document id.  Such as:
+        {'key': 'julia000', 'value': 1},
+        {'key': 'julia001', 'value': 1},
+        {'key': 'julia002', 'value': 1},
+        ...
+        """
+        actual = self.view002(group_level=1)['rows']
```
ricellis: Should this be using `view005`? I was under the impression that `group_level` only applied to complex keys.

alfinkel: I see your point that using view002 is not a very compelling example and it probably does not test the true nature of the `group_level` parameter.  view005 would not work since it does not contain a reduce function.  I added a view006 which is a copy of view005 with a reduce to `_count` and changed this test to use that view with `group_level=1`.  Fixed in https://github.com/cloudant/python-cloudant/pull/119/commits/9b90772fadf62077492db923a95ee032332b568d.

ricellis:
+1

emlaver:
+1

Issue #59, opened 2015-12-24
Fix Cloudant._usage_endpoint() to only accept both year and month or neither


alfinkel:
In the account module, the Cloudant._usage_endpoint() method should only accept either both a year and a month or nothing.  Currently an exception is raised when a month is provided but a year is not.  We need to update the logic so that an exception is raised if either year or month is missing and of course to not raise an exception if both or neither are provided.

The following curl statements confirm this:

acurl -X GET https://ACCT.cloudant.com/_api/v2/bill/2015
acurl -X GET https://ACCT.cloudant.com/_api/v2/usage/data_volume/2015
acurl -X GET https://ACCT.cloudant.com/_api/v2/usage/requests/2015

The returned response for each was:

```son
{
  "error": "The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again."
}
```

Issue #80, opened 2016-01-04
Use assertRaises in unit tests to test exceptions


alfinkel:
Currently we are using the pattern:

``` python
try:
    do_something()
    self.assertFail('The above line should fail')
except SomeException, err:
    self.assertEqual(str(err), 'expected error message.')
```

to check that an exception or error is raised as expected in our unit tests.

Instead we should be using:

``` python
with self.assertRaises(SomeException) as cm:
    do_something()
err = cm.exception
self.assertEqual(str(err), 'expected error message.')
```

This is the case throughout the unit test code base and should be fixed everywhere.

PR #120, opened 2016-03-22
Fix Cloudant._usage_endpoint() to only accept both year and month or neither

emlaver:
## What

In the account module, the Cloudant._usage_endpoint() method should only accept either both a year and a month or nothing.

## How

If both year and month exist, create endpoint request with both parameters. 
If either year or month do not exist, raise a CloudantException.

## Testing

Additional test cases to assert that CloudantException is raised when only year parameter exists for `bill`, `requests_usage`, and `volume_usage`. 

## Reviewers
reviewer @alfinkel 
reviewer @mikerhodes 

## Issues

#59

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [FIX] Added validation to Cloudant.bill, Cloudant.volume_usage, and Cloudant.requests_usage methods to ensure that a valid year/month combination or neither are used as arguments.
 - [FIX] Fixed the handling of empty views in the DesignDocument.
 - [BREAKING] Fixed CloudantDatabase.share_database to accept all valid permission roles.  Changed the method signature to accept roles as a list argument.
 - [FIX] The CouchDatabase.create_document method now will handle both documents and design documents correctly.  If the document created is a design document then the locally cached object will be a DesignDocument otherwise it will be a Document.
--- a/src/cloudant/account.py
+++ b/src/cloudant/account.py
@@ -25,7 +25,7 @@ import requests
 from ._2to3 import bytes_, unicode_
 from .database import CloudantDatabase, CouchDatabase
 from .changes import Feed
-from .errors import CloudantException
+from .errors import CloudantException, CloudantArgumentError
 
 _USER_AGENT = 'python-cloudant/{0} (Python, Version {1}.{2}.{3})'.format(
     sys.modules['cloudant'].__version__,
@@ -392,32 +392,45 @@ class Cloudant(CouchDB):
         optional year and month URL elements.
 
         :param str endpoint: Cloudant usage endpoint.
-        :param int year: Year to query against.  Defaults to None.
-        :param int month: Month to query against.  Defaults to None.
+        :param int year: Year to query against.  Optional parameter.
+            Defaults to None.  If used, it must be accompanied by ``month``.
+        :param int month: Month to query against that must be an integer
+            between 1 and 12. Optional parameter. Defaults to None.
+            If used, it must be accompanied by ``year``.
         """
-        if year is not None:
-            endpoint = posixpath.join(endpoint, str(year))
-        if month is not None:
-            if year is None:
-                raise CloudantException(
-                    (
-                        "must supply both year and month "
-                        "to usage endpoint: {0}"
-                    ).format(endpoint)
-                )
-            endpoint = posixpath.join(endpoint, str(month))
-        resp = self.r_session.get(endpoint)
-        resp.raise_for_status()
-        return resp.json()
+        err = None
+        if year is None and month is None:
+            resp = self.r_session.get(endpoint)
+        else:
+            try:
+                if int(year) > 0 and int(month) in range(1, 13):
+                    resp = self.r_session.get(
+                        posixpath.join(
+                            endpoint, str(int(year)), str(int(month)))
+                    )
+                else:
+                    err = ('Invalid year and/or month supplied.  '
+                           'Found: year - {0}, month - {1}').format(year, month)
+            except (ValueError, TypeError):
+                err = ('Invalid year and/or month supplied.  '
+                       'Found: year - {0}, month - {1}').format(year, month)
+
+        if err:
+            raise CloudantArgumentError(err)
+        else:
+            resp.raise_for_status()
+            return resp.json()
 
     def bill(self, year=None, month=None):
         """
-        Retrieves Cloudant billing data, optionally for a given year/month.
+        Retrieves Cloudant billing data, optionally for a given year and month.
 
-        :param int year: Year to query against, for example 2014.  Defaults to
-            None.
-        :param int month: Month to query against, for example a number from
-            1 to 12.  Defaults to None.
+        :param int year: Year to query against, for example 2014.
+            Optional parameter.  Defaults to None.  If used, it must be
+            accompanied by ``month``.
+        :param int month: Month to query against that must be an integer
+            between 1 and 12.  Optional parameter.  Defaults to None.
+            If used, it must be accompanied by ``year``.
 
         :returns: Billing data in JSON format
         """
@@ -426,12 +439,15 @@ class Cloudant(CouchDB):
 
     def volume_usage(self, year=None, month=None):
         """
-        Retrieves Cloudant volume usage data, optionally for a given year/month.
+        Retrieves Cloudant volume usage data, optionally for a given
+        year and month.
 
-        :param int year: Year to query against, for example 2014.  Defaults to
-            None.
-        :param int month: Month to query against, for example a number from
-            1 to 12.  Defaults to None.
+        :param int year: Year to query against, for example 2014.
+            Optional parameter.  Defaults to None.  If used, it must be
+            accompanied by ``month``.
+        :param int month: Month to query against that must be an integer
+            between 1 and 12.  Optional parameter.  Defaults to None.
+            If used, it must be accompanied by ``year``.
 
         :returns: Volume usage data in JSON format
         """
@@ -443,12 +459,14 @@ class Cloudant(CouchDB):
     def requests_usage(self, year=None, month=None):
         """
         Retrieves Cloudant requests usage data, optionally for a given
-        year/month.
+        year and month.
 
-        :param int year: Year to query against, for example 2014.  Defaults to
-            None.
-        :param int month: Month to query against, for example a number from
-            1 to 12.  Defaults to None.
+        :param int year: Year to query against, for example 2014.
+            Optional parameter.  Defaults to None.  If used, it must be
+            accompanied by ``month``.
+        :param int month: Month to query against that must be an integer
+            between 1 and 12.  Optional parameter.  Defaults to None.
+            If used, it must be accompanied by ``year``.
 
         :returns: Requests usage data in JSON format
         """
--- a/tests/unit/db/account_tests.py
+++ b/tests/unit/db/account_tests.py
@@ -31,7 +31,7 @@ from datetime import datetime
 
 from cloudant import cloudant, couchdb, couchdb_admin_party
 from cloudant.account import Cloudant, CouchDB
-from cloudant.errors import CloudantException
+from cloudant.errors import CloudantException, CloudantArgumentError
 
 from .unit_t_db_base import UnitTestDbBase
 from ... import bytes_, str_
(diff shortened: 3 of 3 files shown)


src/cloudant/account.py
```diff
@@ -392,32 +392,34 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
+        """
+        if (year is not None) and (month is not None):
+            endpoint = posixpath.join(endpoint, str(year), str(month))
+        elif (year is not None) or (month is not None):
+            raise CloudantException(
+                (
+                    "must supply both year and month "
+                    "to usage endpoint: {0}"
+                ).format(endpoint)
+            )
```
mikerhodes: It's hard to understand the logic here, it's not clear what happens for "neither year nor month".

src/cloudant/account.py
```diff
@@ -392,32 +392,34 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
-                    ).format(endpoint)
-                )
-            endpoint = posixpath.join(endpoint, str(month))
+        :param int year: Year to query against.  Optional parameter.
+        Defaults to None.  If used, it must be accompanied by the ``month``.
+        :param int month: Month to query against. Optional parameter.
+        Defaults to None.  If used, it must be accompanied by the ``year``.
+        """
+        if (year is not None) and (month is not None):
+            endpoint = posixpath.join(endpoint, str(year), str(month))
```
mikerhodes: `posixpath` isn't the right thing to use here as it's for file systems. As you are constructing a URL which always uses forward slashes in its path part, it's safe and clear to just do `"{0}/{1}/{2}".format(endpoint, year, month)`.

src/cloudant/account.py
```diff
@@ -392,32 +392,34 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
-                        "to usage endpoint: {0}"
-                    ).format(endpoint)
-                )
-            endpoint = posixpath.join(endpoint, str(month))
+        :param int year: Year to query against.  Optional parameter.
+        Defaults to None.  If used, it must be accompanied by the ``month``.
+        :param int month: Month to query against. Optional parameter.
+        Defaults to None.  If used, it must be accompanied by the ``year``.
+        """
+        if (year is not None) and (month is not None):
```
mikerhodes: `if year and month` is more idiomatic.

src/cloudant/account.py
```diff
@@ -426,12 +428,15 @@ def bill(self, year=None, month=None):
...
         """
-        Retrieves Cloudant volume usage data, optionally for a given year/month.
+        Retrieves Cloudant volume usage data, optionally for a given
+        year and month.
 
-        :param int year: Year to query against, for example 2014.  Defaults to
-            None.
+        :param int year: Year to query against, for example 2014.
+            Optional parameter.  Defaults to None.  If used, it must be
+            accompanied by the ``month``.
```
mikerhodes: "must be accompanied by ``month``" reads better as it's a parameter name.

tests/unit/db/account_tests.py
```diff
@@ -498,6 +522,19 @@ def test_requests_usage_data(self):
...
             self.client.disconnect()
 
+    def test_set_year_without_month_for_requests_usage_data(self):
+        """
+        Test raising an exception when retrieving requests usage data with only
+        year parameter
+        """
+        try:
+            self.client.connect()
+            year = datetime.now().year
```
mikerhodes: Is there any reason not just to use a hard-coded value in these tests? I'd rather if we were able to do that as it removes variability in the tests.

src/cloudant/account.py
```diff
@@ -443,12 +448,14 @@ def volume_usage(self, year=None, month=None):
...
         :param int month: Month to query against, for example a number from
-            1 to 12.  Defaults to None.
+            1 to 12.  Optional parameter.  Defaults to None.
```
mikerhodes: "for example" isn't appropriate here, just use "must be an integer between 1 and 12".

src/cloudant/account.py
alfinkel: @mikerhodes: I think using `posixpath` is OK to use in [constructing URLs](http://effbot.org/librarybook/posixpath.htm).  I've also checked the posixpath.py docstring in both Python 2.7 and 3.5 and they both contain:

```
...
Some of this can actually be useful on non-Posix systems too, e.g.
for manipulation of the pathname component of URLs.
```

src/cloudant/account.py
emlaver: Fixed in be827cb.

src/cloudant/account.py
emlaver: Fixed in be827cb.

src/cloudant/account.py
emlaver: Fixed in be827cb.

tests/unit/db/account_tests.py
emlaver: Fixed in be827cb.

src/cloudant/account.py
emlaver: Fixed in be827cb.

src/cloudant/account.py
```diff
@@ -392,32 +392,38 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
+                posixpath.join(endpoint, str(year), str(month))
+            )
+        elif not year and not month:
+            resp = self.r_session.get(endpoint)
+        else:
+            raise CloudantException(
+                (
+                    "must supply both year and a valid month "
+                    "between 1 and 12 to usage endpoint: {0}"
+                ).format(endpoint)
```
alfinkel: I think this error message is misleading and the exception should be a `CloudantArgumentError` in this case.  I know it was from the original code but it doesn't seem to work here.  I would change this to:

```python
msg = 'Invalid year and/or month supplied.  Found: year - {0}, month - {1}'.format(year, month)
```

tests/unit/db/account_tests.py
```diff
@@ -450,6 +450,31 @@ def test_billing_data(self):
...
 
+    def test_set_year_without_month_for_billing_data(self):
+        """
+        Test raising an exception when retrieving billing data with only
+        year parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            self.assertRaises(CloudantException, self.client.bill, year)
```
alfinkel: Change this to use the assertRaises context manager and then to assert that the error that is returned is the desired message.  For example:

```python
with self.assertRaises(CloudantArgumentError) as cm:
    self.client.bill(year)
expected = ('Invalid year and/or month supplied.  Found: year - 2016, month - None')
self.assertEqual(str(cm.exception), expected)
```

This also works into our plans to update all of our exception testing to use this pattern.  See #80.

tests/unit/db/account_tests.py
```diff
@@ -450,6 +450,31 @@ def test_billing_data(self):
...
+    def test_set_year_with_invalid_month_for_billing_data(self):
+        """
+        Test raising an exception when retrieving billing data with an
+        invalid month parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            month = 13
+            self.assertRaises(CloudantException, self.client.bill, year, month)
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/120/files#r57163130

src/cloudant/account.py
```diff
@@ -392,32 +392,38 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
-                )
-            endpoint = posixpath.join(endpoint, str(month))
-        resp = self.r_session.get(endpoint)
+        :param int year: Year to query against.  Optional parameter.
+            Defaults to None.  If used, it must be accompanied by ``month``.
+        :param int month: Month to query against that must be an integer
+            between 1 and 12. Optional parameter. Defaults to None.
+            If used, it must be accompanied by ``year``.
+        """
+        if year and month in range(1, 13):
```
alfinkel: Since we're validating arguments here, we should validate that the year is an integer.

tests/unit/db/account_tests.py
```diff
@@ -450,6 +450,31 @@ def test_billing_data(self):
...
+        finally:
+            self.client.disconnect()
+
```
alfinkel: I think it might also be useful to add a tests for:

1. When the month is supplied but the year is not
2. When year is invalid (like when a string is passed in)

alfinkel: Same comment applies to volume usage and requests usage.

tests/unit/db/account_tests.py
```diff
@@ -474,6 +499,33 @@ def test_volume_usage_data(self):
...
 
+    def test_set_year_without_month_for_volume_usage_data(self):
+        """
+        Test raising an exception when retrieving volume usage data with only
+        year parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            self.assertRaises(CloudantException, self.client.volume_usage, year)
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/120/files#r57163130

tests/unit/db/account_tests.py
```diff
@@ -474,6 +499,33 @@ def test_volume_usage_data(self):
...
+        Test raising an exception when retrieving volume usage data with an
+        invalid month parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            month = 13
+            self.assertRaises(
+                CloudantException, self.client.volume_usage, year, month
+            )
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/120/files#r57163130

tests/unit/db/account_tests.py
```diff
@@ -498,6 +550,34 @@ def test_requests_usage_data(self):
...
+    def test_set_year_without_month_for_requests_usage_data(self):
+        """
+        Test raising an exception when retrieving requests usage data with an
+        invalid month parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            self.assertRaises(
+                CloudantException, self.client.requests_usage, year)
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/120/files#r57163130

tests/unit/db/account_tests.py
```diff
@@ -498,6 +550,34 @@ def test_requests_usage_data(self):
...
+        Test raising an exception when retrieving requests usage data with only
+        year parameter
+        """
+        try:
+            self.client.connect()
+            year = 2016
+            month = 13
+            self.assertRaises(
+                CloudantException, self.client.requests_usage, year, month
+            )
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/120/files#r57163130

alfinkel:
Don't forget to add an entry to CHANGES.rst.

src/cloudant/account.py
emlaver: @alfinkel Which of these exception messages is more clear for when users pass a type that is not `int`:
```
            msg = 'Invalid year and/or month supplied.  ' \
                  'Check that both year and month are of type integer.' \
                  'Found: year - {0}, month - {1}'\
                .format(year, month)
```
Or print out the type of year and month:
```
            msg = 'Invalid year and/or month supplied. ' \
                  'Found: year - {0} of {1}, month - {2} of {3}' \
                .format(year, type(year), month, type(month))
```

alfinkel: I think the trick here is to implement validation that ensures that the message described (here)[https://github.com/cloudant/python-cloudant/pull/120/files/be827cbda13b61a1024377db42807ecfd5fac058#r57162205] works.

What I mean is that if the year, for example, is a string of '2016', it should not be rejected by this validation because 2016 is a valid year but if the year is 'foo' then it should not pass the validation.  I think the validation logic can be implemented so that the message: `'Invalid year and/or month supplied.  Found: year - {0}, month - {1}'.format(year, month)` works for any condition.

emlaver:
Added entry to CHANGES.rst in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

src/cloudant/account.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

tests/unit/db/account_tests.py
emlaver: Fixed in in d16de58.

src/cloudant/account.py
emlaver: Fixed in in d16de58.

CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [FIX] Fixed Cloudant._usage_endpoint to only accept both year and month or neither.
```
alfinkel: Since `._usage_endpoint` is private, I think it would be better to frame the change in terms of the API so that when users look at this file they will know how the change affects them.

Something like: `[FIX] Added validation to Cloudant.bill(...), Cloudant.volume_usage(...), and Cloudant.requests_usage(...) to ensure that a valid year/ month combination or neither are used as arguments.`

src/cloudant/account.py
```diff
@@ -393,22 +393,34 @@ def _usage_endpoint(self, endpoint, year=None, month=None):
...
+                        posixpath.join(
+                            endpoint, str(int(year)), str(int(month)))
+                    )
+                else:
+                    err = 'Invalid year and/or month supplied.  ' \
+                          'Found: year - {0}, month - {1}'.format(year, month)
+            except ValueError:
+                err = 'Invalid year and/or month supplied.  ' \
+                      'Found: year - {0}, month - {1}'.format(year, month)
+
```
alfinkel: I think the point behind using the try/except block was to simplify the `if` logic by catching exceptions from the call to the `int()` built-in.  That being the case, you should change the except to catch not only `ValueError` but also `TypeError` which is what gets raised if either year or month are None.  As in `except (ValueError, TypeError):`.

Also for line continuation on statements don't use `\`.  Instead use parentheses.  The `if` statement should now fit on a single line once you remove the explicit check for `year` and `month` but you can change the `err` setting statements to: 

```python
err = ('Invalid year and/or month supplied.  '
    'Found: year - {0}, month - {1}').format(year, month)
```

tests/unit/db/account_tests.py
alfinkel: I don't think this is fixed.

tests/unit/db/account_tests.py
alfinkel: Also not fixed.

tests/unit/db/account_tests.py
alfinkel: Not fixed.

tests/unit/db/account_tests.py
alfinkel: Not fixed.

tests/unit/db/account_tests.py
alfinkel: Nope.

src/cloudant/account.py
emlaver: Fixed in d16de58.

CHANGES.rst
emlaver: Fixed in d16de58.

tests/unit/db/account_tests.py
emlaver: Yea, I missed several of the tests in the last commit.  The 5 test cases that needed the assertRaises context manager are now fixed in d16de58.

alfinkel:
+1

src/cloudant/account.py
mikerhodes: I'd still say it's an odd use, and unnecessary indirection when the string munging works fine.

alfinkel: If you're set on having this changed then I think it should be done as part of a separate PR since this is how we compose URLs throughout the library, so the change should probably be applied to the entire repo.  But fwiw, it doesn't seem like an odd use to me since the docs back it up and this way of constructing URLs was inherited from the original library written by the SAPI folks.  But it's your call @mikerhodes.

alfinkel: I dug into this a bit more...

I think if we change from using `posixpath.join` to using `urljoin` it should probably makes things clearer from a code readability standpoint and we'd then be using a library that is meant for all things URL related.  See [Py2 - urlparse.urljoin](https://docs.python.org/2/library/urlparse.html#urlparse.urljoin) and [Py3 - urllib.parse.urljoin](https://docs.python.org/3.2/library/urllib.parse.html#urllib.parse.urljoin).

I've added a separate issue #127 to track this, so that we can handle this at the repo level.

mikerhodes:
+1

Issue #42, opened 2015-12-08
Add Result class validation logic


alfinkel:
We need to add unit tests that test Result class functionality against CouchDB and Cloudant database instances.  Once these Result unit tests are in place we should have a full suite of unit tests that test this library against database instances.  We should then consider removing the mocked unit test versions as they may be redundant and at times ineffective w.r.t. testing of functionality.  The removal of the mocked tests may be spun off as a separate issue.

alfinkel:
As part of this testing we will also need to fix any bugs encountered during the testing as well as adding in validation routines within the Result class specifically w.r.t. key, slice and iteration processing and how each process relates to the query parameters passed in specifically key, keys, start_key, end_key, startkey_docid, endkey_docid, limit, and skip.

alfinkel:
Splitting parts of this issue to #109 and #110 for better manageability.

This issue will now only track the addition of validation to the Result class as described https://github.com/cloudant/python-cloudant/issues/42#issuecomment-195046790 as well as the addition of Result class unit tests that verify the item retrieval, slicing, and iteration.

alfinkel:
Splitting out tests and related documentation updates to separate issues for better manageability and review.  See #121 and #122.

This issue will now only apply to the Result class validation logic changes.

PR #124, opened 2016-03-25
Add Result class validation

alfinkel:
## What:

Add result collection key/index access, key/index slicing and iteration validation logic to the Result class and update supporting documentation.

## How:

- Add a ResultException class to handle all Result specific exceptions.
- Add logic to Result `__getitem__()` to handle data access via index.  As in `result[10]` will return the data in position 10 of the result collection.
- Add validation logic to key slicing to ensure that invalid requests are caught.
- Update iteration logic validation to use ResultException.
- Update "Dealing with Results" section of the "Getting Started" documentation.
- Update all affected docstring content in database, views, and result modules.

## Testing:

All tests have been written and the new tests all pass successfully under Python 2.7 and 3.5 targeting both CouchDB as well as Cloudant database instances.  The tests will be submitted in a follow up PR to keep the review process manageable.

## Reviewers:

reviewer: @emlaver 
reviewer: @brynh 

## Issues:

- #42

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -214,51 +214,40 @@ class CouchDatabase(dict):
 
     def get_view_result(self, ddoc_id, view_name, raw_result=False, **kwargs):
         """
-        Retrieves the view result based on the design document
-        and view name.  By default the result is returned as a
-        :class:`~cloudant.result.Result` which uses the ``skip`` and
-        ``limit`` query parameters internally to handle slicing and iteration
-        through the view result collection.  Therefore ``skip`` and ``limit``
-        cannot be used as arguments to get the view result when
-        ``raw_result=False``.  However, by setting ``raw_result=True``, the
-        result will be returned as the raw JSON response content for the view
-        requested.  Using this setting requires the developer to manage their
-        own slicing and iteration.  Therefore ``skip`` and ``limit`` are valid
-        arguments in this instance.
+        Retrieves the view result based on the design document and view name.
+        By default the result is returned as a
+        :class:`~cloudant.result.Result` object which provides a key
+        accessible, sliceable, and iterable interface to the result collection.
+        Depending on how you are accessing, slicing or iterating through your
+        result collection certain query parameters are not permitted.  See
+        :class:`~cloudant.result.Result` for additional details.
 
-        For example to retrieve the default Result object based on a
-        design document view do:
+        However, by setting ``raw_result=True``, the result will be returned as
+        the raw JSON response content for the view requested.  With this setting
+        there are no restrictions on the query parameters used but it also
+        means that the result collection key access, slicing, and iteration is
+        the responsibility of the developer.
 
-        .. code-block:: python
-
-            db.get_view_result('_design/ddoc_id_001', 'view_001')
-
-        But to retrieve a customized Result object based on the
-        same design document view do something like:
+        For example:
 
         .. code-block:: python
 
-            db.get_view_result('_design/ddoc_id_001', 'view_001',
+            # get Result based on a design document view
+            result = db.get_view_result('_design/ddoc_id_001', 'view_001')
+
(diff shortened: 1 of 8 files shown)


src/cloudant/errors.py
```diff
@@ -49,3 +49,30 @@ class CloudantArgumentError(CloudantException):
     """
     def __init__(self, msg, code=400):
         super(CloudantArgumentError, self).__init__(msg, code)
+
+class ResultException(CloudantException):
+    """
+    Provides a way to issue Cloudant Python client library result specific
+    exceptions.
+
+    :param int code: An code value used to identify the result exception.
```
emlaver: I would remove 'An code' and just have:
```
:param int code: Value used to ...
```

alfinkel: Rather than switching to `Value used to...`, I made it `A code value used to...` to keep consistent with the other two Cloudant exception classes.  Fixed in https://github.com/cloudant/python-cloudant/pull/124/commits/2bb4fc32c7434e1c22c1bd89875d5e186669d919.

src/cloudant/result.py
```diff
@@ -176,99 +217,193 @@ class Result(object):
...
-        raise CloudantArgumentError(msg)
+        data = None
+        if isinstance(arg, int):
+            data = self._handle_result_by_index(arg)
+        elif isinstance(arg, STRTYPE) or isinstance(arg, list):
+            data = self._handle_result_by_key(arg)
+        elif isinstance(arg, ResultByKey):
+            data = self._handle_result_by_key(arg())
+        elif isinstance(arg, slice):
+            # slice is entire result set - no additional processing required
```
emlaver: Did you mean `slice the entire result set` or `slice entire result set`?

alfinkel: What this comment means is that the slice requested is the entire result set.  So we don't need to do any additional processing to the request as we do for key slicing or index slicing.

src/cloudant/database.py
```diff
@@ -214,51 +214,40 @@ def get_design_document(self, ddoc_id):
...
+            # get raw response content based on a design document view
+            result = db.get_view_result('_design/ddoc_id_001', 'view_001',
                 raw_result=True)
 
-        But to retrieve the raw JSON response content based on a set of
-        parameters on the same design document view do something like:
-
-        .. code-block:: python
-
+            # get customized the raw response content for a design document view
```
emlaver: `get customized the raw response content` sounds weird. 
Maybe something like:
`get raw response with customized content based on a design doc view`

emlaver:
+1

src/cloudant/result.py
```diff
@@ -129,40 +130,80 @@ def type_or_none(typerefs, value):
     """
     return isinstance(value, typerefs) or value is None
 
+class ResultByKey(object):
+    """
+    Provides a wrapper for a value used to retrieving records from a result
```
brynh: I think `to retrieving` should be changed to either `to retrieve` or `for retrieving`.

src/cloudant/result.py
```diff
@@ -176,99 +217,193 @@ class Result(object):
         for i in result:
             print i
 
-        # Iterate over the entire result collection,
-        # including documents and in batches of a 1000.
+        # Iterate over the entire result collection in batches of a 1000, including documents.
```
brynh: `a 1000` should be just `1000`.

src/cloudant/result.py
alfinkel: Fixed in 65b05169c7a7b08f87cd0177266bdba88b5744fb

src/cloudant/result.py
alfinkel: Fixed in 65b05169c7a7b08f87cd0177266bdba88b5744fb

brynh:
+1

Issue #123, opened 2016-03-25
Move credentials module to tests folder


alfinkel:
- remove credentials module from this library since its functionality is out of scope of this library's functionality.
- remove from documentation as well.

alfinkel:
Moving credentials.py under `tests` folder for possible later use when resolving https://github.com/cloudant/python-cloudant/issues/70 and https://github.com/cloudant/python-cloudant/issues/114.

Issue #114, opened 2016-03-16
Clean up integration tests


alfinkel:
Clean up the integration tests found in the `tests/integration` folder.  They have not been kept current with the current code base and likely do not function currently.  We may need to look to adding additional tests here as well and finally we want to consider adding these tests to be part of the Travis CI process.

Issue #70, opened 2015-12-24
Use configuration file for tests targeting a database


alfinkel:
See PR 26 comment:  https://github.com/cloudant/python-cloudant/pull/26#discussion-diff-43795872

Ref:  https://docs.python.org/2/library/configparser.html

PR #125, opened 2016-03-28
Move credentials module under tests

alfinkel:
## What:

Moved the credentials.py module under the tests folder.

## Why:

The functionality provided by the credentials.py module does not fit into the feature set of this library.  It may however be useful in the event that we change our unit testing process to use a configuration file rather than environment variables as is described in issue #70.

## How:

- Move credentials.py from `src/cloudant` to `tests` folder.
- Remove reference to credentials in the documentation.

## Testing:

- Updated the end to end integration test to point to the new location of the credentials module.  The integration test functionality will be revisited as part of issue #114.

## Reviewers:

reviewer: @rhyshort 

## Issues:

- #123


rhyshort:
+1

emlaver:
+1

Issue #110, opened 2016-03-11
Support QueryResult key by index and relax skip/limit restriction


alfinkel:
Add unit tests to ensure that the QueryResult class functions as expected.

This is part of splitting #42 into manageable work items.

alfinkel:
Changed the scope of this issue somewhat to adding support for using a single integer key when looking for a QueryResult and also relaxing the validation logic that prohibits the use of skip and limit with a QueryResult.

alfinkel:
The only validation restrictions on a QueryResult should be that the argument passed to `__getitem__()` should be an `int` or a slice where the start and stop values are both `type_or_none()`.

So these should all be valid:

``` python
qr = QueryResult(callable)
qr[10]
qr[10:20]
qr[10: ]
qr[: 20]
qr[: ]
```

alfinkel:
- QueryResult iteration will still be handled through the `Result.__iter__()` and the skip/limit restrictions there will also apply to QueryResult.
- If a `skip` or a `limit` are found in the `self._ref` of a QueryResult object then those parameters should be copied over to `self.options`, if there is no `skip` and/or `limit`already in the `self.options`.  This will ensure that `__getitem()__` logic in the Result will be able to handle skip/limit processing when dealing with results by key index and index slicing.

PR #130, opened 2016-04-01
Add QueryResult data access by index value

alfinkel:
## What:

Added the ability to access data from a QueryResult by using an index value.  As in:

```python
result = QueryResult(...)
result[0]             # Retrieves the 1st document in the query result
result[9]             # Retrieves the 10th document in the query result
```

Also relaxed skip/limit restrictions on QueryResult slicing.  Now skip and/or limit can be used in the original Query or the customized QueryResut.

## How:

- Added logic to handle the processing of a single integer value as the key of a QueryResult.
- Removed logic that raised exceptions when either `skip` or `limit` were encountered in the Query or the QueryResult custom options.

## Testing:

- Refactored and added additional QueryResult tests to ensure that all functionality is working as expected.

## Reviewers:

reviewer: @emlaver 
reviewer: @brynh 

## Issues

- #110

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,6 @@
 2.0.0 (Unreleased)
 ==================
-- [FIX] Added validation to Cloudant.bill, Cloudant.volume_usage, and Cloudant.requests_usage methods to ensure that a valid year/month combination or neither are used as arguments.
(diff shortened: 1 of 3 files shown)


emlaver:
LGTM +1

brynh:
+1

Issue #44, opened 2015-12-08
Change account module to client module


alfinkel:
Consider changing the module name from account.py to clients.py.  clients.py seems to make more sense since the module contains the CouchDB and Cloudant classes that dictate which client connection to perform.  While the notion of account is valid with Cloudant, it is less so clear with CouchDB.

alfinkel:
Removed line item from original description as it is handled in Issue #58:

alfinkel:
Removed line item from original description as it is handled in Issue #116

PR #136, opened 2016-04-05
Rename account.py to client.py

alfinkel:
## What:

Rename the account module to be the client module.

## Why:

The notion of an account is clear when it comes to Cloudant but less so when it comes to CouchDB.  Since this library serves both, we are switching the name of the account module to client so that it makes sense for both Cloudant and CouchDB.

## How:

- Rename account.py as client.py
- Update all import references in code and tests
- Update documentation .rst files
- Update docsting comments as necessary to reference client instead of account

## Testing:

- No additional tests but all tests using account module now use client module
- Comments updated as necessary

## Reviewers:

reviewer: @rhyshort 
reviewer: @emlaver 

## Issues

- #44

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -15,10 +15,10 @@ Connections
 In order to manage a connection you must first initialize the connection by 
 constructing either a ``Cloudant`` or ``CouchDB`` client.  Since connecting to 
 the Cloudant managed service provides extra end points as compared to a CouchDB 
-instance, we provide the two different client implementations in order to 
+server, we provide the two different client implementations in order to 
 connect to the desired database service.  Once the client is constructed, 
-you follow that up by connecting to the account, performing your tasks, and then 
-disconnecting from the account.
+you follow that up by connecting to the server, performing your tasks, and
+then disconnecting from the server.
 
 Later in the `Context managers`_ section we will see how to 
 simplify this process through the use of the Python *with* statement.
@@ -29,16 +29,16 @@ Connecting with a client
 .. code-block:: python
 
     # Use CouchDB to create a CouchDB client
-    # from cloudant.account import CouchDB
+    # from cloudant.client import CouchDB
     # client = CouchDB(USERNAME, PASSWORD, url='http://127.0.0.1:5984')
 
     # Use Cloudant to create a Cloudant client using account
-    from cloudant.account import Cloudant
+    from cloudant.client import Cloudant
     client = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME)
     # or using url
     # client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com')
     
-    # Connect to the account
+    # Connect to the server
     client.connect()
 
     # Perform client tasks...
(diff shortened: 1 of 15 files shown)


docs/getting_started.rst
```diff
@@ -17,8 +17,8 @@ constructing either a ``Cloudant`` or ``CouchDB`` client.  Since connecting to
 the Cloudant managed service provides extra end points as compared to a CouchDB 
 instance, we provide the two different client implementations in order to 
 connect to the desired database service.  Once the client is constructed, 
-you follow that up by connecting to the account, performing your tasks, and then 
-disconnecting from the account.
+you follow that up by connecting to the instance, performing your tasks, and
```
rhyshort: Using instance here doesn't quite sound right, perhaps it should be `server` rather than `instance`?

docs/getting_started.rst
```diff
@@ -29,16 +29,16 @@ Connecting with a client
...
 
     # Use Cloudant to create a Cloudant client using account
-    from cloudant.account import Cloudant
+    from cloudant.client import Cloudant
     client = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME)
     # or using url
     # client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com')
     
-    # Connect to the account
+    # Connect to the instance
```
rhyshort: I think it should be `server` here too

src/cloudant/client.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015 IBM. All rights reserved.
+# Copyright (c) 2016 IBM. All rights reserved.
```
rhyshort: I'm no copyright expert but I thought this would read `2015 - 2016` ?

alfinkel: I figured that since it was technically a "new" file, that it should be 2016.

src/cloudant/client.py
```diff
@@ -325,7 +325,7 @@ def __setitem__(self, key, value, remote=False):
         Note:  The only way to override the default for the ``remote`` argument
         setting it to True is to call __setitem__ directly.  A much simpler
         approach is to use
-        :func:`~cloudant.account.CouchDB.create_database` instead, if your
+        :func:`~cloudant.clients.CouchDB.create_database` instead, if your
```
emlaver: `clients` should be `client`.

docs/getting_started.rst
alfinkel: Fixed in 6a7b2a0

docs/getting_started.rst
alfinkel: Fixed in 6a7b2a0

src/cloudant/client.py
alfinkel: Changed to `2015, 2016` in 93e123b

src/cloudant/client.py
alfinkel: Fixed in 93e123b

rhyshort:
+1

emlaver:
+1

Issue #132, opened 2016-04-04
Change views module to view module


alfinkel:
Rename views.py to view.py to match naming pattern for modules within this library.

Issue #113, opened 2016-03-16
Remove mocked tests and move db tests to unit folder


alfinkel:
Once all of the database unit tests are complete we should remove the mocked tests since they will be redundant at this point.

Database unit tests will be complete once #42, #109, #110, #112 have been closed.

alfinkel:
Be sure to add some mocked tests for the `client.db_updates` method before deleting all mocked tests since testing against a db is somewhat prohibitive.

See: https://github.com/cloudant/python-cloudant/blob/master/tests/unit/mocked/client_test.py#L49

alfinkel:
_db_updates tests will be now be handled in #147

PR #138, opened 2016-04-06
Rename views.py module to view.py

emlaver:
## What:

Rename views.py to view.py to match naming pattern for modules within this library.
#132 

## How:

- Rename views.py to view.py
- Update all import references in code and tests
- Update Sphinx documentation .rst files
- Update docsting comments

## Testing:

- No additional tests
- All tests using views module now use view module

## Reviewers:

reviewer @alfinkel 
reviewer @ricellis

### DIFF
--- a/tests/unit/mocked/view_test.py
+++ b/tests/unit/mocked/view_test.py
@@ -13,21 +13,21 @@
 # See the License for the specific language governing permissions and 
 # limitations under the License.
 """
-_views_test_
+_view_test_
 
-views module unit tests
+view module unit tests
 
 """
 import mock
 import unittest
 
-from cloudant.views import Code, View
+from cloudant.view import Code, View
 from cloudant.design_document import DesignDocument
 from cloudant.result import Result
 
 
 class CodeTests(unittest.TestCase):
-    """tests for cloudant.views.Code, not much to test here yet..."""
+    """tests for cloudant.view.Code, not much to test here yet..."""
     def test_code(self):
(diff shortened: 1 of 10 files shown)


tests/unit/mocked/view_test.py
```diff
@@ -13,21 +13,21 @@
...
 
-from cloudant.views import Code, View
+from cloudant.view import Code, View
 from cloudant.design_document import DesignDocument
 from cloudant.result import Result
 
 
 class CodeTests(unittest.TestCase):
-    """tests for cloudant.views.Code, not much to test here yet..."""
+    """tests for cloudant.view.Code, not much to test here yet..."""
```
ricellis: Can probably update this comment, looks like there are some tests.

alfinkel:
+1

tests/unit/mocked/view_test.py
alfinkel: I think the comment is correct.  There's just the one test.

ricellis:
I couldn't find any other references to the `views` module so +1

tests/unit/mocked/view_test.py
alfinkel: Looking at the code, the one "test" does not even assert anything so there is actually < 1 tests :).

In any event, these mocked tests will soon be removed anyway.

emlaver: @alfinkel Yes, I agree, this class will be removed in #113.

Issue #46, opened 2015-12-09
Remove make_result from Query and View


alfinkel:
The `make_result` method in View and Query classes is redundant and confusing.  It should be removed.    A Result/QueryResult object can be stood up that essentially would perform the same functionality.  Likewise the database convenience methods of `get_view_results` and `get_query_results` also perform the same functionality.

- database module
  - change `get_view_results` to standup a new Result object passing it **kwargs.
  - change `get_query_results` to standup a new QueryResult object passing it **kwargs.
- result module
  - Move parameter docs from `make_result` methods to Result/QueryResult class docstring.
  - Remove references to `make_result` in comments.
- view module
  - Remove `make_result`.
  - Refactor `custom_result` to use either the `result` attribute on View or simply standup a new Result object passing it **kwargs.
  - Add `result` attribute to the class docstring.
- query module
  - Remove `make_result`.
  - Refactor `custom_result` to use either the `result` attribute on Query or simply standup a new QueryResult object passing it **kwargs.
    - Add `result` attribute to the class docstring.
- Tests
  - Ensure current tests still work.
  - Remove `make_result` tests.

PR #144, opened 2016-04-08
Remove make_result from Query and View

emlaver:
## What

The make_result method in Query and View classes is redundant and confusing. It should be removed. A Result/QueryResult object can be stood up that essentially would perform the same functionality. Likewise the database convenience methods of get_view_results and get_query_results also perform the same functionality.

## How

In database.py module:
- change get_view_results to standup a new Result object passing it **kwargs.
- change get_query_results to standup a new QueryResult object passing it **kwargs.

In result.py module:
- Move parameter docs from make_result methods to Result/QueryResult class docstring.
- Remove references to make_result in comments.

In view.py module:
- Remove make_result.
- Refactor custom_result to use a new Result object passing it **kwargs.
- Add QueryIndexView.custom_result to override and disable the base method View.custom_result

In query.py module:
- Remove make_result.
- Refactor custom_result to use a new QueryResult object passing it **kwargs.

## Testing

No additional tests.
Removed make_result tests.

## Reviewers

reviewer @alfinkel 
reviewer @ricellis

## Issues

#46

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [BREAKING] Removed the make_result method from View and Query classes.  If you need to make a query or view result, use ``CloudantDatabase.get_query_result``, ``CouchDatabase.get_view_result``, or the ``View.custom_result`` context manager.  Additionally, the ``Result`` and ``QueryResult`` classes can be called directly to construct a result object.
 - [BREAKING] Renamed modules account.py, errors.py, indexes.py, views.py, to client.py, error.py, index.py, and view.py.
 - [FIX] Added validation to ``Cloudant.bill``, ``Cloudant.volume_usage``, and ``Cloudant.requests_usage`` methods to ensure that a valid year/month combination or neither are used as arguments.
 - [FIX] Fixed the handling of empty views in the DesignDocument.
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -236,10 +236,38 @@ class Result(object):
     | Iteration                     | ``limit``, ``skip`` not permitted                         |
     +-------------------------------+-----------------------------------------------------------+
 
-    :param method_ref: A reference to the method or callable that returns
+    :param str method_ref: A reference to the method or callable that returns
         the JSON content result to be wrapped as a Result.
-    :param options: See :func:`~cloudant.view.View.make_result` for a
-        list of valid result customization options.
+    :param bool descending: Return documents in descending key order.
+    :param endkey: Stop returning records at this specified key.
+        Not valid when used with key access and key slicing.
+    :param str endkey_docid: Stop returning records when the specified
+        document id is reached.
+    :param bool group: Using the reduce function, group the results to a
+        group or single row.
+    :param group_level: Only applicable if the view uses complex keys: keys
+        that are JSON arrays. Groups reduce results for the specified number
+        of array fields.
+    :param bool include_docs: Include the full content of the documents.
+    :param bool inclusive_end: Include rows with the specified endkey.
+    :param key: Return only documents that match the specified key.
+        Not valid when used with key access and key slicing.
+    :param list keys: Return only documents that match the specified keys.
+        Not valid when used with key access and key slicing.
+    :param int limit: Limit the number of returned documents to the
+        specified count.  Not valid when used with key iteration.
+    :param int page_size: Sets the page size for result iteration.
+    :param bool reduce: True to use the reduce function, false otherwise.
+    :param int skip: Skip this number of rows from the start.
+        Not valid when used with key iteration.
+    :param str stale: Allow the results from a stale view to be used. This
+        makes the request return immediately, even if the view has not been
+        completely built yet. If this parameter is not given, a response is
+        returned only after the view has been built.
+    :param startkey: Return records starting with the specified key.
+        Not valid when used with key access and key slicing.
+    :param str startkey_docid: Return records starting with the specified
+        document ID.
     """
     def __init__(self, method_ref, **options):
         self.options = options
@@ -404,9 +432,6 @@ class Result(object):
         a batch of data from the result collection and then yields each
         element.
 
-        See :func:`~cloudant.view.View.make_result` for a list of valid
-        result customization options.
-
         See :class:`~cloudant.result.Result` for Result iteration examples.
 
         :returns: Iterable data sequence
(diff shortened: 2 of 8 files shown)


CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [BREAKING] Removed make_result from query.py and view.py. Replaced instances of make_result with a Result or QueryResult object.
```
alfinkel: This is confusing.  Information about the module is not as important as is the Class that the method if being removed from.  Perhaps change this to just one sentence:  `Removed make_result method from View and Query classes`.

src/cloudant/result.py
```diff
@@ -236,10 +236,47 @@ class Result(object):
     | Iteration                     | ``limit``, ``skip`` not permitted                         |
     +-------------------------------+-----------------------------------------------------------+
 
+    List of valid Result object customization options:
+    :param bool descending: Return documents in descending key order.
+    :param endkey: Stop returning records at this specified key.
+        Not valid when used with :class:`~cloudant.result.Result` key
```
alfinkel: Since we are already in the Result class there is no need for a link to the ~cloudant.result.Result documentation.  I think it is enough to just say "Not valid when used with key access and key slicing.

You should apply this here and to lines 254, 257, 261, 265, and 271.

src/cloudant/result.py
```diff
@@ -236,10 +236,47 @@ class Result(object):
...
+        makes the request return immediately, even if the view has not been
+        completely built yet. If this parameter is not given, a response is
+        returned only after the view has been built.
+    :param startkey: Return records starting with the specified key.
+        Not valid when used with :class:`~cloudant.result.Result` key
+        access and key slicing.
+    :param str startkey_docid: Return records starting with the specified
+        document ID.
+
+
```
alfinkel: Remove these empty lines

CHANGES.rst
emlaver: Fixed in 3514bad.

src/cloudant/result.py
```diff
@@ -236,10 +236,47 @@ class Result(object):
...
     :param method_ref: A reference to the method or callable that returns
```
alfinkel: I'm not sure why this is down here after a couple of empty lines.  The `method_ref` param should be at the top of the params list.  Also, it is a string so it should read `:param str method_ref`.

src/cloudant/result.py
```diff
@@ -236,10 +236,47 @@ class Result(object):
...
         the JSON content result to be wrapped as a Result.
-    :param options: See :func:`~cloudant.view.View.make_result` for a
-        list of valid result customization options.
+    :param options: A list of valid Result customization options.
```
alfinkel: Since we are now explicitly specifying all of the valid options we no longer need an "options" catch all param.  Remove this.

src/cloudant/result.py
```diff
@@ -404,7 +441,7 @@ def __iter__(self):
         a batch of data from the result collection and then yields each
         element.
 
-        See :func:`~cloudant.view.View.make_result` for a list of valid
+        See :class:`~cloudant.result.Result` for a list of valid
         result customization options.
```
alfinkel: I think you can now remove this "See:" reference.

src/cloudant/result.py
```diff
@@ -236,10 +236,47 @@ class Result(object):
     | Iteration                     | ``limit``, ``skip`` not permitted                         |
     +-------------------------------+-----------------------------------------------------------+
 
+    List of valid Result object customization options:
```
alfinkel: Remove this line.

src/cloudant/result.py
```diff
@@ -503,10 +540,32 @@ class QueryResult(Result):
...
+    :param str use_index: Identifies a specific index for the query to run
+        against, rather than using the Cloudant Query algorithm which finds
+        what it believes to be the best index.
+
+
     :param query: A reference to the query callable that returns
         the JSON content result to be wrapped.
-    :param options: See :func:`~cloudant.query.Query.make_result` for a
-        list of valid query result customization options.
+    :param options: A list of valid QueryResult customization options.
```
alfinkel: Apply the same comments that I made on the Result docstring changes here as well.

src/cloudant/view.py
```diff
@@ -451,10 +399,10 @@ def __call__(self, **kwargs):
             'use the database \'get_query_result\' convenience method.'
         )
 
-    def make_result(self, **options):
+    def custom_result(self, **options):
```
alfinkel: You should add a test for this.

alfinkel: Nevermind, there already is.

alfinkel: I would, however, change the CloudantException message that is being raised by this method to something that tells the developer that the context manager is invalid rather than you cannot make a result.

Something like 'Cannot create a custom result context manager using a QueryIndexView.  If you wish to execute a query use the database "get_query_result" convenience method instead.'

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Fixed in 5222af0.

src/cloudant/view.py
emlaver: Fixed in 5222af0.

src/cloudant/result.py
emlaver: Removed the link ~cloudant.result.Result in 5222af0 and f5d6f34

alfinkel:
+1

CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [BREAKING] Removed make_result method from View and Query classes.
```
ricellis: I think it might be good to direct people to the appropriate methods to use here in case anyone was using `make_result`.

ricellis:
+1 after small addition to CHANGES.md

CHANGES.rst
emlaver: @ricellis @alfinkel How does this sound:
```
[BREAKING] Removed make_result method from View and Query classes and moved its functionality into ``CloudantDatabase.get_query_result``, ``CouchDatabase.get_view_result``, and ``View.custom_result``.
```

emlaver: UPDATE:
```
[BREAKING] Removed make_result method from View and Query classes. Use one of the methods ``CloudantDatabase.get_query_result``, ``CouchDatabase.get_view_result``, or ``View.custom_result`` for equivalent functionality.
```

ricellis: +1

alfinkel: I suppose if we're going to 100% accurate about what to do now for a developer to "make a result" then I think this should read something like this:

```
[BREAKING] Removed the make_result method from View and Query classes.  If you need to make a query or view result now, use ``CloudantDatabase.get_query_result``, ``CouchDatabase.get_view_result``, or the ``View.custom_result`` context manager, or you can use the ``Result`` and ``QueryResult`` classes directly and construct a result object that way.
```

Issue #45, opened 2015-12-08
Investigate Feed implementation


alfinkel:
- We need to investigate how the Feed class is implemented.  Specifically, how a continuous feed works.  It appears that the endpoint parameters (for a continuous feed) are not being used to manage how a `_changes` or a `_db_updates` call is handled.
- Got feedback from user of 0.5.9 cloudant-python lib.
  - He uses the changes feed and saw that the changes feed in this library doesn't function the same as before.  We need to ensure that we are providing an optimal feed solution.

See: https://github.com/cloudant-labs/cloudant-python/issues/31 and https://github.com/cloudant-labs/cloudant-python/pull/33

alfinkel:
A request from @willholley on Slack:

It would be very useful to be able to fetch _changes in batches rather than an infinite iterator in python-cloudant

I’m batching changes from one database and inserting them into another for the insert, I need to batch updates together but with an iterator, I can’t know when I’ve reached the end

alfinkel:
Rename changes module to feed.

- changes.py -> feed.py

evansde77:
FWIW the iterator has a non-continuous mode which should allow fetching in batches using since etc:
https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/changes.py#L51
Looks like it is hard coded into params though:
https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/changes.py#L51

alfinkel:
Yeah, that hardcoding into the params appears to be the root cause of the problem.  However, if I override it by setting the `_params` to `{'feed': 'normal'}` when setting `continuous=False` I do get the iterator to complete but there is still some issue with the processing of the feed content.  I don't think it is a major issue but with that said I'll still have to dig a little deeper into it before coming up with the full resolution.  Thanks @evansde77 for pointing that out.

evansde77:
Looks like I did a shitty job testing it :)

alfinkel:
You and me both.

alfinkel:
The continuous and non-continuous feeds return different response stream content.  A continuous feed returns change results whereas a non-continuous feed returns change results in a "result" array along with a "last_seq" number.  The result array and the last_seq number make up a dictionary object.  Because of this difference, we need to handle the response slightly differently for non-continuous feeds.  Currently the logic only handles the continuous feed cleanly.

Ref: https://docs.cloudant.com/database.html#get-changes

[...]

PR #145, opened 2016-04-14
Fix feed implementation

alfinkel:
## What:

Fix the feed implementation to accurately utilize the `changes` and `_db_updates` endpoints and their respective options.

## How:

- Rename changes.py to feed.py
- Change the implementation to be true to the _changes and _db_updates endpoint usage
- Add feed option validation
- Add a stop() method
- Add the ability to provide formatted as well as raw feed content
- Update documentation

## Testing:

- Update existing tests accordingly
- Create new feed specific tests that target a database

## Reviewers:

reviewer: @emlaver 
reviewer: @rhyshort

## Issues

- #45


src/cloudant/client.py
```diff
@@ -210,27 +210,54 @@ def delete_database(self, dbname):
...
+            for db_update in db_updates:
+                print(db_update)
+            print(db_updates.last_seq)
+
+            # Iterate over a "continuous" _db_updates feed with additional options
+            db_updates = db.db_updates(feed='continuous', since='now', descending=True)
+            for db_update in db_updates:
+                print(db_update)
+
+        :param bool raw_data: Dictates whether the data returned by the feed is
```
emlaver: Maybe change to something like:
```
True if data returned is formatted, false if returned as a raw response.  Default is False.
```

tests/unit/db/feed_tests.py
```diff
@@ -0,0 +1,544 @@
...
+                feed.stop()
+        expected = set(['julia{0:03d}'.format(i) for i in range(100)])
+        self.assertSetEqual(set([x['id'] for x in changes]), expected)
+        self.assertIsNone(feed.last_seq)
+        # Compare continuous with normal
+        normal = Feed(self.session, self.changes_url)
+        self.assertSetEqual(
+            set([x['id'] for x in changes]), set([n['id'] for n in normal]))
+
+    def test_get_longpoll_feed(self):
```
emlaver: Should `heartbeat` be added to the test name?

alfinkel: No because the heartbeat is just a way for us to drop out of the continuous feed and complete the test.  The actual heartbeat test is a few tests down.

tests/unit/db/feed_tests.py
```diff
@@ -0,0 +1,544 @@
...
+        self.assertIsInstance(feed._r_session, Session)
+        self.assertTrue(feed._raw_data)
+        self.assertDictEqual(feed._options, {'feed': 'continuous'})
+        self.assertEqual(feed._chunk_size, 1)
+
+    def test_constructor_db_updates(self):
+        """
+        Test constructing a _db_updates feed
+        """
+        feed = Feed(self.session, self.db_updates_url, feed='continuous',
```
emlaver: This is the only test against the `db_updates` endpoint and it doesn't test getting back a response. Shouldn't we have more tests similar to the ones below for `db_updates`?

tests/unit/db/feed_tests.py
```diff
@@ -0,0 +1,544 @@
...
+                changes.append(change)
+                beats = 0
+            else:
+                beats += 1
+            if beats == 15 or len(changes) == 3:
+                feed.stop()
+        expected = set(['julia003', 'julia004', 'julia005'])
+        self.assertSetEqual(set([x['id'] for x in changes]), expected)
+
+    def test_get_feed_using_timeout(self):
```
emlaver: What about another test case to assert that ``heartbeat`` supersedes ``timeout``?

src/cloudant/changes.py
```diff
@@ -1,112 +0,0 @@
-#!/usr/bin/env python
```
rhyshort: Can you do this a `git mv` so we retain the history of the file?

src/cloudant/client.py
```diff
@@ -210,27 +210,54 @@ def delete_database(self, dbname):
...
+            number.  Default is no limit.
+        :param since: Start the results from changes after the specified
+            sequence identifier. In other words, using since excludes from the
+            list all changes up to and including the specified sequence
+            identifier. If since is 0 (the default), or omitted, the request
+            returns all changes. If it is ``now``, only changes made after the
+            time of the request will be emitted.
+        :param int timeout: Number of milliseconds to wait for data before
+            terminating the response. ``heartbeat`` supersedes ``timeout`` if
+            both are supplied.
```
rhyshort: Shouldn't this also support using a filter function? or hasn't that been implemented yet?

rhyshort: oh wait no i see, this isn't changes I miss read my bad

tests/unit/db/feed_tests.py
alfinkel: I don't see the point to test that since that seems to me like server functionality.  The purpose of these tests are specifically to ensure that the logic in returning the iterated stream of data is sound and that the argument options are validated and translated from Python to couch/cloudant request correctly.  I specifically did not put any validation in place to block anyone from using heartbeat and timeout together so there's no test for it.

tests/unit/db/feed_tests.py
alfinkel: My plan with `_db_updates` tests was to add them in a follow up PR since this one is already quite bloated.  We won't make a release without the tests being in place.  I am tracking them with #147.

emlaver: Ok, sounds good.

src/cloudant/client.py
alfinkel: Fixed in both client and database.  Done in d28550d22672b0558ed7dfafe75839b299bc3484.

src/cloudant/changes.py
rhyshort: as per slack discussion, lets skip this :)

rhyshort:
+1

emlaver:
LGTM +1

src/cloudant/changes.py
alfinkel: I went ahead and did this, but I had to split it across two commits.  Done in 	e3f598b and 4617505.

Issue #146, opened 2016-04-14
Add an infinite feed implementation


alfinkel:
The original feed implementation's intent seemed to be that of an infinite feed where if a `last_seq` was encountered as part of a continuous feed then another http request would be initiaited to the feed endpoint using the last_seq value as the "since" value thus creating an infinite feed iterator.

Since the feed was rewritten (as part of #45) to be more in line with how the `_changes` and `_db_updates` feed endpoints and their options function, this infinite feed functionality has been somewhat lost.

This issue tracks the re-addition of the infinite feed functionality back into this library.

Consider an InfiniteFeed class??

PR #148, opened 2016-04-18
Add infinite feed implementation

alfinkel:
## What:

Added the ability for a feed (_changes or _db_updates) to return an infinite iterator.

## How:

Anytime a last sequence identifier is encountered for a `continuous` feed another HTTP request stream is made to the server thus creating a perpetually refreshed feed.  This was done by extending the Feed class as an InfiniteFeed class.  The InfiniteFeed has added logic in the `next()` method call to check for a last sequence identifier and if found to re-issue the HTTP request with a new `since` option set to the last sequence identifier.

The InfiniteFeed implementation is only valid for CouchDatabase and CloudantDatabase _changes feeds and the Cloudant _db_updates feed only.  The CouchDB _db_updates feed does not return a last sequence identifier and does not accept `since` as an option so the client db_updates method was placed on the child Cloudant class instead of the parent CouchDB class.  In other words the _db_updates InfiniteFeed implementation is not supported for CouchDB.

## Testing:

- Added an infinite_feed_tests.py module.  These tests test the actual InfiniteFeed functionality.
- Updated client_tests.py module.  These tests specifically verify that an InfiniteFeed object is constructed and that the necessary settings are in place in the object.
- Updated database_tests.py module.  These tests specifically verify that an InfiniteFeed object is constructed and that the necessary settings are in place in the object.

## Reviewers:

reviewer: @brynh 
reviewer: @ricellis

## Issues:

- #146

### DIFF
--- a/src/cloudant/feed.py
+++ b/src/cloudant/feed.py
@@ -77,9 +77,9 @@ class Feed(object):
 
     :param Session session: Requests session used by the Feed.
     :param str url: URL used by the Feed.
-    :param bool raw_data: Dictates whether the streamed data is returned as
-        a decoded Python ``dict`` object or as raw response data.  Defaults to
-        False.
+    :param bool raw_data: If set to True then the raw response data will be
+        streamed otherwise if set to False then JSON formatted data will be
+        streamed.  Default is False.
     """
     def __init__(self, session, url, raw_data=False, **options):
         self._r_session = session
@@ -130,6 +130,11 @@ class Feed(object):
         else:
             msg = 'Could not identify feed based on url: {0}'.format(self._url)
             raise CloudantException(msg)
+        if isinstance(self, InfiniteFeed) and options.get('feed') != 'continuous':
+            msg = (
+                'Invalid infinite feed option: {0}.  Must be set to continuous.'
+            ).format(options.get('feed'))
+            raise CloudantArgumentError(msg)
         translation = dict()
         for key, val in iteritems_(options):
             _validate(key, val, arg_types)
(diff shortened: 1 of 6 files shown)


tests/unit/db/client_tests.py
```diff
@@ -374,6 +375,22 @@ def test_create_db_via_setitem(self):
...
             self.client.disconnect()
 
+    def test_db_updates_feed_call(self):
+        """
+        Test that db_updates() method call constructs and returns a Feed object
+        """
+        try:
+            self.client.connect()
+            db_updates = self.client.db_updates(limit=100)
+            self.assertIsInstance(db_updates, Feed)
```
brynh: I think this would also pass if `db_updates` was an `InfiniteFeed`, but I think (given my limited Python expertise)  if it were written as `assertIs(type(db_updates), Feed)`, it would ensure it's a `Feed` but not an `InfiniteFeed`.

tests/unit/db/database_tests.py
```diff
@@ -601,41 +602,28 @@ def test_view_clean_up(self):
...
-        for change in feed:
-            self.assertEqual(change['id'], doc['_id'])
-            i += 1
-            doc = self.db.create_document(
-                {'_id': 'julia{0:03d}'.format(i), 'name': 'julia', 'age': i})
-            if i == 10:
-                feed.stop()
-        self.assertEqual(i, 10)
+        changes = self.db.changes(limit=100)
+        self.assertIsInstance(changes, Feed)
```
brynh: Again, I think `assertIs(type(changes), Feed)`, is a slightly stronger assertion.

tests/unit/db/client_tests.py
alfinkel: Good point.  Fixed in 1569ce598cc6a00e3939d94cbfccfeb67bf51baa.

tests/unit/db/database_tests.py
alfinkel: Fixed in 1569ce598cc6a00e3939d94cbfccfeb67bf51baa.

brynh:
+1

src/cloudant/feed.py
```diff
@@ -206,3 +206,37 @@ def _process_data(self, line):
...
+class InfiniteFeed(Feed):
+    """
+    Provides an infinite iterator for consuming client and database feeds such
+    as ``_db_updates`` and ``_changes``.  An InfiniteFeed object is constructed
+    with a reference to a client or database Requests Session object and a feed
+    endpoint URL.  Unlike a :class:`~cloudant.feed.Feed` which can be either a
+    ``normal``, ``longpoll``, or ``continuous`` feed an InfiniteFeed can only be
+    ``continuous`` and the iterator will only stream formatted JSON objects.
+    Instead of using this class directly, it is recommended to use the client
+    API :func:`~cloudant.client.CouchDB.db_updates` and the database API
```
ricellis: shouldn't this be `infinite_db_updates`?

src/cloudant/feed.py
```diff
@@ -206,3 +206,37 @@ def _process_data(self, line):
...
+    :func:`~cloudant.database.CouchDatabase.changes`.  Reference those methods
```
ricellis: and `infinite_changes`?

src/cloudant/feed.py
```diff
@@ -206,3 +206,37 @@ def _process_data(self, line):
...
+        Handles the iteration by pulling the next line out of the stream and
+        converting the response to JSON.
+
+        :returns: Data representing what was seen in the feed
+        """
+        if self._last_seq:
+            self._options.update({'since': self._last_seq})
+            self._resp = None
+            self._last_seq = None
+        return super(InfiniteFeed, self).next()
```
ricellis: I wonder if this should be wrapped in some exception swallowing (possibly outputting a warning or something). It seems to me that the likely use case of an infinite feed is to keep processing changes _ad infinitum_. Won't it terminate at the first exception atm? Admittedly different exceptions might warrant different behaviour (e.g. a networking glitch probably shouldn't break the feed, but an unauthorized probably should). With that in mind maybe it is sufficient for a user's code to restart the feed after an exception if we explicitly document that it is not exception resilient (and possibly how to restart from the same place)?

src/cloudant/feed.py
```diff
@@ -206,3 +206,37 @@ def _process_data(self, line):
...
+    :func:`~cloudant.database.CouchDatabase.changes`.  Reference those methods
+    for a valid list of feed options.
+
+    :param Session session: Requests session used by the Feed.
+    :param str url: URL used by the Feed.
+    """
+    def __init__(self, session, url, **options):
+        super(InfiniteFeed, self).__init__(session, url, False, **options)
+        # feed must be set to "continuous" in the case of an infinite feed
+        self._options.update({'feed': 'continuous'})
```
ricellis: I'm inclined to suggest that this should throw if another `feed` type has been supplied. Despite the documentation I think it is potentially confusing to have this blind override behaviour.

src/cloudant/feed.py
alfinkel: Fixed in 2914b998a4b006ec7cef9a0d4ffa047fd90f391d

src/cloudant/feed.py
alfinkel: Fixed in 2914b998a4b006ec7cef9a0d4ffa047fd90f391d

src/cloudant/feed.py
alfinkel: I added a "Note:" in the class docstring as you described.  I also added a GH issue #149 to revisit this and to consider adding in some addtional exception handling and restart logic.  Done in 2914b998a4b006ec7cef9a0d4ffa047fd90f391d

src/cloudant/feed.py
alfinkel: Done in 2914b998a4b006ec7cef9a0d4ffa047fd90f391d.  Also added a test.

ricellis:
+1

Issue #147, opened 2016-04-15
Add more validation logic to the feed module for db_updates


alfinkel:
Add unit tests to test _db_updates functionality

alfinkel:
Changing the scope of this issue from just adding tests for db_updates but also adding additional validation as well.

- The _db_updates implementations for CouchDB and Cloudant differ enough to warrant that we add validation in the InfiniteFeed and/or possibly the Cloudant.db_updates() method to ensure that the CouchDB and Cloudant `_db_updates` are handled based on their set of options.  The `infinite_db_updates` method has already been moved from the CouchDB class to the Cloudant class due to the fact that the CouchDB `_db_updates` endpoint does not rely or return a `last_seq` identifier.  See #146 / #148.
- As part of this work we may want to consider changing the Feed and InfiniteFeed constructor signature to accept a `source` (feed source) instead of a Requests Session and a URL.  This feed source will be either a client or a database object and the Requests Session and URL can be acquired that way instead.  If we choose to go this route then we will likely need to add something to CouchDB, Cloudant, CouchDatabase, and CloudantDatabase to allow the feed to know which type of object the Feed if working with since importing into the feed module is not an option because that would cause a circular dependency.  Will need to research a bit more here...
- Obviously add more tests.

alfinkel:
Bullet #2 in the previous comment is proving more of a hassel than originally expected.  While doing an import of client and database in the feed works as expected, it does however still cause pylint to issue a series of cyclic import warnings.  That being the case, I believe that leaving the feed constructor signature as is and instead extending the Feed with a class(es) that take the source as an argument may be the better way to go.

PR #151, opened 2016-04-25
Cleanup db updates logic and add tests

alfinkel:
## What:

Updated Feed/InfiniteFeed logic to add validation for `_db_updates` feeds for CouchDB and Cloudant.

## How:

- Separated _db_updates argument dictionaries into CouchDB _db_updates, Cloudant _db_updates, and _changes (for both CouchDB and Cloudant.
- Changed Feed/InfiniteFeed constructor to accept a client or a database object and to use that to get the request session and contstruct the appropriate URL.
- Ensure that appropriate validation logic exists for CouchDB _db_updates, Cloudant _db_updates, and CouchDB/Cloudant _changes.
- Updated client and database modules according to Feed/InfiniteFeed changes detailed above.
  - Specifically for client module, we now have a separate `db_updates` method for CouchDB and Cloudant.

## Testing:

- Added a db_updates_tests module
- Renamed feed_tests to changes_tests and moved all _db_updates related tests to the db_updates_tests module.
- Updated all tests to work with new Feed/InfiniteFeed implementation

## Reviewers:

reviewer @emlaver 
reviewer @ricellis

## Issues:

- #147


tests/unit/db/db_updates_tests.py
```diff
@@ -0,0 +1,469 @@
...
+                    feed.stop()
+        changes = [json.loads(unicode_(x)) for x in raw_content]
+        self.assertDictEqual(
+            changes[0], {'db_name': self.new_dbs[0].database_name, 'type': 'created'})
+        self.assertDictEqual(
+            changes[1], {'db_name': self.new_dbs[1].database_name, 'type': 'created'})
+        self.assertDictEqual(
+            changes[2], {'db_name': self.new_dbs[2].database_name, 'type': 'created'})
+
+    def test_get_longpoll_feed_as_default(self):
```
emlaver: What about creating a database using `create_dbs` for longpoll tests?

tests/unit/db/db_updates_tests.py
```diff
@@ -0,0 +1,469 @@
...
+        self.assertIsInstance(feed._r_session, Session)
+        self.assertFalse(feed._raw_data)
+        self.assertDictEqual(feed._options,
+            {'feed': 'continuous', 'heartbeat': False, 'timeout': 2})
+
+    def test_stop_iteration_of_continuous_feed_with_heartbeat(self):
+        """
+        - Test stopping the iteration
+        - Test continuous feed
+        - Test heartbeat
```
emlaver: Probably better to keep the comments consistent by using `and` instead of bullet points.

tests/unit/db/db_updates_tests.py
alfinkel: In CouchDB `_db_updates` works a bit differently from `Cloudant`.  There is no "since" option and a `longpoll` feed always starts from the point from which the request is made to the server.  As such, there is no point to prime the server with any new databases because they would be irrelevant to the `longpoll` request.  Also, the `longpoll` behavior appears to be a "one and done" type situation.  Meaning that a `longpoll` feed is waiting for a single "event" to occur after which it terminates and a heartbeat constitutes an event, so there is no opportunity to create a database as part of the feed loop either since the first event will always be a heartbeat in this test.

tests/unit/db/db_updates_tests.py
alfinkel: Fixed in 28dfc67d5cac4dbfc457a881e49dfe0b2bd89274

ricellis:
+1

emlaver:
LGTM +1

Issue #58, opened 2015-12-24
Remove references to Cloudant in CouchDB classes


alfinkel:
In various parts of the library, we have Couch* classes, but they have attributes with names like 'cloudant_url'. This doesn't make sense and we should change these names. 

This issue seems to be ubiquitous throughout (found in accounts.py, document.py, database.py, result.py, replicator.py) the library and we may need to create subcases for this task. Essentially all of the modules suffer from this either in the code or the documentation. 

There should be some priority for this so that we don't let this bleed out into the API for too long.

PR #153, opened 2016-04-27
Rename cloudant attribute references in couchdb classes

alfinkel:
## What:

A number of the CouchDB specific classes have public and private attributes with "cloudant" as part of their name.  This PR removes "cloudant" references from those attributes to make things clearer from a useability and maintenance point of view.

## How:

- Rename client module CouchDB class attributes referencing Cloudant
- Rename database module CouchDatabase class attributes referencing Cloudant
- Rename document module Document class attributes referencing Cloudant
- Rename feed module Feed/InfiniteFeed class attributes referencing Cloudant
- Rename query module Query class attributes referencing Cloudant

## Testing:

- Rename cloudant references in tests as needed based on above.

## Reviewers:

reviewer: @rhyshort 
reviewer: @brynh 

## Issues

- #58

### DIFF
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -55,11 +55,11 @@ class CouchDB(dict):
 
     def __init__(self, user, auth_token, admin_party=False, **kwargs):
         super(CouchDB, self).__init__()
-        self._cloudant_user = user
-        self._cloudant_token = auth_token
-        self._cloudant_session = None
-        self.cloudant_url = kwargs.get('url')
-        self._cloudant_user_header = None
+        self._user = user
+        self._auth_token = auth_token
+        self._client_session = None
+        self.server_url = kwargs.get('url')
+        self._client_user_header = None
         self.admin_party = admin_party
         self.encoder = kwargs.get('encoder') or json.JSONEncoder
         self.r_session = None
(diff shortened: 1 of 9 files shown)


src/cloudant/client.py
```diff
@@ -55,11 +55,11 @@ class CouchDB(dict):
...
         super(CouchDB, self).__init__()
-        self._cloudant_user = user
-        self._cloudant_token = auth_token
-        self._cloudant_session = None
-        self.cloudant_url = kwargs.get('url')
-        self._cloudant_user_header = None
+        self._user = user
+        self._auth_token = auth_token
+        self._client_session = None
+        self.client_url = kwargs.get('url')
```
rhyshort: wouldn't this make more sense being called `server_url`?

alfinkel: I suppose it depends on how you look at it.  I named it `client_url` because the url attribute continues to come up as you move down the object management chain.  As in, for a database object you would have a `database_url` and for a document object you would have a `document_url`.  So I was following the naming convention pattern in naming it `client_url`.  I think it may be odd to have a public attribute on a client object named `server_url` but I could be convinced otherwise since I do see your point.  After knowing all this, do you still think it should be `server_url`?  Perhaps @brynh can chime in as well.

brynh: I can kind of see both arguments (although I think I'd be biased towards calling it `server_url` rather than `client_url`), but maybe just calling it `url` would work?

alfinkel: I considered just naming it `url` but stopped short of doing that because I felt like then I would need to rename `database_url` and `document_url` to `url` in the database and document classes.

Giving it some more thought I renamed `client_url` to `server_url` as suggested, reasoning that while the attribute is public in the client it is not annotated as a property while the `database_url` and the `document_url` are.  So they are somewhat different and `server_url` makes more sense than `client_url`.  Done in 901e7d6.

brynh:
+1

rhyshort:
+1

PR #154, opened 2016-04-27
60008 429 backoff

ricellis:
## What

Added backoff/retry for 429 status codes.

## How

Use the request library's built in `HTTPAdapter`to provide retry configuration to the underlying urllib3 library. Configure for 10 retries starting at a 250 ms backoff for all CouchDB HTTP methods. Use 0 retries for connection and read so these retries apply only to the `status_forcelist` which includes only the 429 status code, so no other retry behaviour is affected by the change.

## Testing

No new tests, provides configuration to underlying library. Existing tests would fail if configuration syntax was wrong. Testing the actual retry function would require adding a mock server. The pattern in python-cloudant has been to avoid testing underlying library code in this way. I have manually verified that the retry functions. 

## Reviewers

reviewer: @alfinkel
reviewer: @emlaver

## Issues
Internal case 60008

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
 2.0.0 (Unreleased)
 ==================
+- [NEW] Handle HTTP status code ``429 Too Many Requests`` with blocking backoff and retries.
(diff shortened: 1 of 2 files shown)


ricellis:
Added extra back ticks in d13ce63

emlaver:
LGTM +1

Issue #150, opened 2016-04-25
Update User-Agent string


alfinkel:
Update the User-Agent string to match (as much as possible) to the java-cloudant implementation.  See:  https://github.com/cloudant/java-cloudant/pull/246.

- Change user agent content in client.py module to:

```python
import sys
import os
...
_USER_AGENT = '/'.join([
    'python-cloudant',
    sys.modules['cloudant'].__version__,
    'Python',
    '{0}.{1}.{2}'.format(sys.version_info[0], sys.version_info[1], sys.version_info[2]),
    os.uname()[0],
    os.uname()[2]
])
```

PR #156, opened 2016-04-28
Update user agent format

alfinkel:
## What:

Update User-Agent format to be `python-cloudant/<library version>/Python/<Python version>/<OS name>/<OS architecture>`.

So an example User-Agent will look like `python-cloudant/2.0.0/Python/2.7.10/Darwin/x86_64`.

## How:

- Update the _USER_AGENT setting in the client.py

## Testing:

- Update the client test that tests the connection specifics to include checks for the User-Agent string components.

## Reviewers:

reviewer: @ricellis 
reviewer: @emlaver 

## Issues:

- #150

### DIFF
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -20,6 +20,7 @@ import base64
 import json
 import posixpath
 import sys
+import os
 import requests
 
 from ._2to3 import bytes_, unicode_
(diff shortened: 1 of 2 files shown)


emlaver:
+1

ricellis:
+1

Issue #158, opened 2016-04-29
Cloudant search query execution


emlaver:
This is the first part of two work items for Cloudant Search #104:

Handle search query parameter processing and execution in `CloudantDatabase.get_search_result`
Unit tests cases for search query in `CloudantDatabaseTests`

emlaver:
Issue #164 needs to be addressed and merged before handling this item.

emlaver:
Work item title and description updated.

emlaver:
Since DesignDocument search index functions will use the SearchIndex class (#159), we should move the search query execution from the SearchIndex `__call__` function into a `CloudantDatabase.get_search_result` class.  This way, we don’t have any CouchDB/Cloudant restrictions in the future search index management PR (#159) and it's clear that executing a search query can only be done on a CloudantDatabase object.

emlaver:
The query execution is now be handled in `CloudantDatabase.get_search_result`. The `SearchIndex` class is removed as it's no longer required. Updated the work item title and description.

Issue #104, opened 2016-03-02
Add support for Cloudant Search


alfinkel:
We need to add support for creating Search Indexes in a DesignDocument as well as support for performing query searches through the `_search` endpoint targeting a Search Index.

This issue is dependent upon the completion of the bug fix #103.

alfinkel:
ref:  https://docs.cloudant.com/search.html

emlaver:
Item will be split into 3 issues:
1. Search index management
2. Search index query execution
3. Search index query result processing

PR #162, opened 2016-05-04
WIP: Cloudant Search index management

emlaver:
## What

New SearchIndex class in search.py that handles search index management in design documents module.  The search.py module will support future query access through the search endpoint.

## How

- Created search.py module with SearchIndex class.
- New search index accessor property for the locally cached DesignDocument in design_document.py.
- Added functions for search index management (add, update, get, delete, list) in design document module.

## Testing

Test cases to assert new search index management functions in design_document.py.
Added `indexes` key with empty dictionary value to locally cached design documents in existing test case assertions.

## Reviewers

reviewer: @alfinkel

## Issues

#158 
#104


src/cloudant/design_document.py
```diff
@@ -26,15 +27,12 @@ class DesignDocument(Document):
...
-    functions, search indexes, etc.  When instantiating a DesignDocument or
+    provides an API to view management, search index management, list and show
+    functions, etc.  When instantiating a DesignDocument or
     when setting the document id (``_id``) field, the value must start with
     ``_design/``.  If it does not, then ``_design/`` will be prepended to
     the provided document id value.
 
-    Note:  Currently only the view management API exists.  Remaining design
-    document functionality will be added later.
-
```
alfinkel: Why did you choose to remove this note?  All design document sub-document management has not been implemented (ex. list, show, ...).

src/cloudant/search.py
```diff
@@ -0,0 +1,61 @@
...
+
+class SearchIndex(dict):
+    """
+    Encapsulates a search index as a dictionary based object, exposing the
+    search index function as an attribute and supporting query/data access
+    through the search endpoint.  A Search object is instantiated with a
+    reference to a DesignDocument and is typically used as part of the
+    :class:`~cloudant.design_document.DesignDocument`
+    search index management API.
+    """
```
alfinkel: The comment here talks about a `Search` object but the class is `SearchIndex`.  I'm also unclear about the section of docstring that talks about "supporting query/data access through the serch endpoint".  I understand that this was a straight copy of the View class but the View class is callable.  Have you decided that the SearchIndex will be callable as well or is the plan to have a separate class for search.  I know that we spoke a lot about this but I just want to be sure about what your plan is for search index management as it related to search query execution.

alfinkel: Also, You might want to consider having the `SearchIndex` class be part of the index module rather than this `Search` module just from an organization standpoint it might make more sense for it to be there.

src/cloudant/search.py
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+API module for composing and executing Cloudant Search indexes.
+"""
+from ._common_util import codify
+from cloudant.error import CloudantException
+
+
```
alfinkel: I would remove the extra blank line

src/cloudant/search.py
```diff
@@ -0,0 +1,61 @@
...
+    reference to a DesignDocument and is typically used as part of the
+    :class:`~cloudant.design_document.DesignDocument`
+    search index management API.
+    """
+    def __init__(
+            self,
+            ddoc,
+            search_index_name,
+            search_index=None,
+            **kwargs
```
alfinkel: What exactly would/could the `kwargs` be in the case of a search index?

src/cloudant/search.py
```diff
@@ -0,0 +1,61 @@
...
+    ):
+        super(SearchIndex, self).__init__()
+        self._design_doc = ddoc
+        self._search_index_name = search_index_name
+        self['index'] = codify(search_index)
```
alfinkel: You don't want to be doing this here in the constructor.  The View handles map and reduce functions in the form of properties with getters and setters.  You should follow that pattern here as well.

src/cloudant/search.py
```diff
@@ -0,0 +1,61 @@
...
+        if not self._search_index_name:
+            return '/'.join([
+                self._design_doc.document_url,
+                '_search',
+                self._search_index_name
+            ])
+        else:
+            msg = 'Search index name is None or was not provided.'
+            raise CloudantException(msg)
+
```
alfinkel: There a few things that need to be mentioned here:

1.  I think your `if/else` logic is reversed.  Which makes me wonder about your tests if this is not failing.
2.  There does not seem to be a point for the `if/else` block since the `search_index_name` is required in the constructor and will fail there.  The only purpose this check serves is if someone specifically entered in a series of blank spaces as a `search_index_name` but we don't do this type of check anywhere else in the library and I think if at some point you want to add this level of validation then it might make sense to make that decision for the entire library.
3.  Finally, I don't see a need for the `url` property here at all unless you plan to make a `SearchIndex` callable.  I don't think that is what we agreed to before but things may have changed since we spoke last.  In either case, this url property doesn't seem to add any value from a search index management stand point.  So I think it makes sense to remove it and include it where appropriate when you implement the search query execution piece of this work in a later PR.

src/cloudant/design_document.py
```diff
@@ -15,6 +15,7 @@
 """
 API module/class for interacting with a design document in a database.
 """
+from cloudant.search import SearchIndex
```
alfinkel: This is a nit but I like to keep the _2to3 module as the first module in our list of imported modules.  Since it is our Py2 to Py3 compatibility module.

src/cloudant/search.py
alfinkel: facet and store?

src/cloudant/design_document.py
```diff
@@ -79,6 +89,22 @@ def add_view(self, view_name, map_func, reduce_func=None, **kwargs):
...
+    def add_search_index(self, search_index_name, search_index, **kwargs):
+        """
+        Appends a Cloudant search index to the locally cached DesignDocument
+        SearchIndex dictionary.
+
+        :param str search_index_name: Name used to identify the SearchIndex.
+        :param str search_index: Javascript search index function.
+        """
+        if self.get_search_index(search_index_name) is not None:
+            msg = "Search index {0} already exists in this design doc"\
```
alfinkel: I think we touched upon this some time before, we are not using the `\` notation anywhere else in the this library.  Consider breaking this `msg` assignment up differently.  maybe on the open parens in the `format`.

This comment applies to the entire PR.

alfinkel: Also, use single quotes not double quotes to match the code style in the rest of the library.

Again, applies throughout the PR.

src/cloudant/design_document.py
```diff
@@ -153,6 +209,18 @@ def fetch(self):
                         view_def.pop('reduce', None),
                         **view_def
                     )
+        if not self.search_indexes:
+            # Ensure indexes dict exists in locally cached DesignDocument.
+            self.setdefault('indexes', dict())
+        if self.search_indexes:
+            for (search_index_name, search_index) \
+                    in iteritems_(self.get('indexes', dict())):
```
alfinkel: Lets work on getting this on one line.  There really shouldn't be a reason for this for statement have to span multiple lines.  Consider shortening the names of the elements in the tuple and perhaps perform the `iteritems_` call or the `self.get(...)` call on a separate line but the `for` logic should really be on one line.

alfinkel:
I don't think that we considered the fact that a search index is Cloudant specific so that somehow needs to be reflected in our implementation of the DesignDocument.  We cannot provide convenience methods in the DesignDocument that only apply to Cloudant yet not put in safe guards to ensure that someone does not try this with a DesignDocument for CouchDB.  I'm not sure whether checks directly inside of the DesignDocument is the right way to go or if possibly splitting out a DesignDocument to separate CouchDB and Cloudant specific ddoc classes as we do with databases and clients is the right way to go.  But one thing is certain, we do have to handle this.

alfinkel:
Also, have you considered the possibility of adding convenience methods to the CloudantDatabase API as well?  Along the lines of how we handle query indexes?

alfinkel:
Just a suggestion at this point but it might make sense to back off of this PR and address the design questions before reviewing any further.

Issue #164, opened 2016-05-06
Cloudant specific design document class to manage query and search indexes


emlaver:
During the re-design of the Cloudant Search index management work (#162), a new class called CloudantDesignDocument (may rename) will manage both query indexes and search indexes. 
Issue #158 will handle implementing search index management in CloudantDesignDocument.

emlaver:
This new class is a breaking change and will not be handled at this time.  I will create a new issue if this design is recommended for a future major release.

PR #168, opened 2016-05-10
CloudantDesignDocument class for managing query indexes

emlaver:
## What

Create a Cloudant specific design document for managing query indexes.

## How

- Add new CloudantDesignDocument class in design_document.py
- Move query index functionality from DesignDocument functions to CloudantDesignDocument functions

## Testing

Move existing query index tests into CloudantDesignDocumentTests.
No new tests created.

## Reviewers

reviewer @ricellis
reviewer @alfinkel

## Issues

#164

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,6 @@
 2.0.1 (Unreleased)
 ==================
 - [FIX] Fixed issue with Windows platform compatibility,replaced usage of os.uname for the user-agent string.
-- [BREAKING] Moved existing query index management into new Cloudant-based design document class.
 
 2.0.0 (2016-05-02)
 ==================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 2.0.1 (Unreleased)
 ==================
 - [FIX] Fixed issue with Windows platform compatibility,replaced usage of os.uname for the user-agent string.
+- [BREAKING] Moved existing query index management into new Cloudant-based design document class.
 
 2.0.0 (2016-05-02)
 ==================
--- a/src/cloudant/design_document.py
+++ b/src/cloudant/design_document.py
@@ -25,18 +25,17 @@ class DesignDocument(Document):
     """
     Encapsulates a specialized version of a
     :class:`~cloudant.document.Document`.  A DesignDocument object is
-    instantiated with a reference to a database and
-    provides an API to view management, list and show
-    functions, etc.  When instantiating a DesignDocument or
-    when setting the document id (``_id``) field, the value must start with
-    ``_design/``.  If it does not, then ``_design/`` will be prepended to
-    the provided document id value.
+    instantiated with a reference to a :class:`~cloudant.database.CouchDatabase`
+    and provides an API to view management, list and show functions, etc.
+    When instantiating a DesignDocument or when setting the document id
+    (``_id``) field, the value must start with ``_design/``.  If it does not,
+    then ``_design/`` will be prepended to the provided document id value.
 
     Note:  Currently only the view management API exists.  Remaining design
     document functionality will be added later.
 
-    :param database: A database instance used by the DesignDocument.  Can be
-        either a ``CouchDatabase`` or ``CloudantDatabase`` instance.
+    :param CouchDatabase database: A ``CouchDatabase`` instance used to
+        manipulate ``DesignDocument`` content.
     :param str document_id: Optional document id.  If provided and does not
         start with ``_design/``, it will be prepended with ``_design/``.
     """
@@ -116,14 +115,13 @@ class DesignDocument(Document):
             self.setdefault('views', dict())
         else:
             for view_name, view_def in iteritems_(self.get('views', dict())):
-                if self.get('language', None) != QUERY_LANGUAGE:
-                    self['views'][view_name] = View(
-                        self,
-                        view_name,
-                        view_def.pop('map', None),
-                        view_def.pop('reduce', None),
-                        **view_def
-                    )
+                self['views'][view_name] = View(
+                    self,
+                    view_name,
+                    view_def.pop('map', None),
+                    view_def.pop('reduce', None),
+                    **view_def
+                )
 
     def save(self):
         """
@@ -135,11 +133,10 @@ class DesignDocument(Document):
         accordingly based on the successful response of the operation.
         """
         if self.views:
-            if self.get('language', None) != QUERY_LANGUAGE:
-                for view_name, view in self.iterviews():
-                    if isinstance(view, QueryIndexView):
-                        msg = 'View {0} must be of type View.'.format(view_name)
-                        raise CloudantException(msg)
+            for view_name, view in self.iterviews():
+                if isinstance(view, QueryIndexView):
+                    msg = 'View {0} must be of type View.'.format(view_name)
+                    raise CloudantException(msg)
         else:
             # Ensure empty views dict is not saved remotely.
             self.__delitem__('views')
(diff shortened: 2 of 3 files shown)


src/cloudant/design_document.py
```diff
@@ -243,3 +207,128 @@ def info(self):
...
+    Encapsulates a Cloudant version of a
+    :class:`~cloudant.design_document.DesignDocument`.
+    A CloudantDesignDocument object is instantiated with a reference
+    to a Cloudant database and provides an API to query index management,
+    list and show functions, etc.
+
+    Note:  Currently only the query view management API exists.  Remaining
+    Cloudant design document functionality will be added later.
+
+    :param database: A ``CloudantDatabase`` instance used by the
```
ricellis: Isn't this the `CloudantDatabase` instance where the design document is saved? Might it be a little clearer to say that? This description is accurate, but it doesn't really help a caller much in understanding what database to provide.

src/cloudant/design_document.py
```diff
@@ -243,3 +207,128 @@ def info(self):
...
+    def save(self):
+        """
+        Saves changes made to the locally cached CloudantDesignDocument object's
+        data structures to the remote database.  If the design document does not
+        exist remotely then it is created in the remote database.  If the object
+        does exist remotely then the design document is updated remotely.  In
+        either case the locally cached CloudantDesignDocument object is also
+        updated accordingly based on the successful response of the operation.
+        """
+        if self.views:
```
ricellis: As discussed I think this needs a language check for `QUERY_LANGUAGE`

CHANGES.rst
```diff
@@ -1,6 +1,7 @@
 2.0.1 (Unreleased)
 ==================
 - [FIX] Fixed issue with Windows platform compatibility,replaced usage of os.uname for the user-agent string.
+- [BREAKING] Moved existing query index management into new Cloudant-based design document class.
```
alfinkel: I think we touched upon this on Slack but this definitely can't be part of the 2.0.1 and by the sound of things @ricellis has some reservations about it even being part of a 2.1.0 release.  In any case, you should probably remove this line item from the CHANGES.rst for now until you settle on where this work fits from a release perspective.

src/cloudant/design_document.py
```diff
@@ -27,7 +27,7 @@ class DesignDocument(Document):
     :class:`~cloudant.document.Document`.  A DesignDocument object is
     instantiated with a reference to a database and
```
alfinkel: I think now that this work is splitting CouchDB and Cloudant design document functionality it probably makes sense to be more descriptive in this DesignDocument docstring and rework this docstring appropriately.  At the very least you probably should change the generic "database" reference to a link to the CouchDatabase class.

alfinkel: The param description for [database](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/design_document.py#L38-L39) also needs to be fixed. 

Don't forget to add the class name to the param as well as in `:param CouchDatabase database: ...`

src/cloudant/design_document.py
```diff
@@ -97,37 +85,28 @@ def update_view(self, view_name, map_func, reduce_func=None, **kwargs):
...
     def fetch(self):
         """
         Retrieves the remote design document content and populates the locally
-        cached DesignDocument dictionary.  View content is stored either as
-        View or QueryIndexView objects which are extensions of the ``dict``
+        cached DesignDocument dictionary.  View content is stored as
+        a View object which is an extension of the ``dict``
         type.  All other design document data are stored directly as
         ``dict`` types.
         """
```
alfinkel: Can you explain why the `if` check for `QUERY_LANGUAGE` still exists on [line 119](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/design_document.py#L119)?

src/cloudant/design_document.py
```diff
@@ -169,13 +140,6 @@ def save(self):
                     if isinstance(view, QueryIndexView):
```
alfinkel: Same [question](https://github.com/cloudant/python-cloudant/pull/168/files#r62887515) for line [138](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/design_document.py#L138)

src/cloudant/design_document.py
alfinkel: This description definitely needs more clarity.  See [this comment](https://github.com/cloudant/python-cloudant/pull/168/files#r62885465) and this [other comment](https://github.com/cloudant/python-cloudant/pull/168/files#r62886306).  Be sure that these two cases (DesignDocument and CloudantDesignDocument) are handled similarly.

src/cloudant/design_document.py
```diff
@@ -27,7 +27,7 @@ class DesignDocument(Document):
     :class:`~cloudant.document.Document`.  A DesignDocument object is
     instantiated with a reference to a database and
     provides an API to view management, list and show
```
alfinkel: Also be sure that what you list here as potential upcoming design document API is accurate for CouchDB.  Can you confirm that show and list are valid for CouchDB?

src/cloudant/design_document.py
emlaver: My original thought was to keep this to not allow adding query indexes in DesignDocument.  But after our earlier discussion, this check should be removed from line [line 119](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/design_document.py#L119) and [line 138](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/design_document.py#L138) since CloudantDesignDocument handles all query index logic and other functionality.

src/cloudant/design_document.py
```diff
@@ -243,3 +207,128 @@ def info(self):
         GET databasename/_design/{ddoc}/_info
         """
         raise NotImplementedError("_info not yet implemented")
+
+class CloudantDesignDocument(DesignDocument):
+    """
+    Encapsulates a Cloudant version of a
+    :class:`~cloudant.design_document.DesignDocument`.
+    A CloudantDesignDocument object is instantiated with a reference
+    to a Cloudant database and provides an API to query index management,
```
alfinkel: Change `Cloudant database` to a link to CloudantDatabase use 2 lines above DesignDocument as a reference.

src/cloudant/design_document.py
```diff
@@ -243,3 +207,128 @@ def info(self):
...
+    Note:  Currently only the query view management API exists.  Remaining
+    Cloudant design document functionality will be added later.
+
+    :param database: A ``CloudantDatabase`` instance used by the
+        CloudantDesignDocument.
+    :param str document_id: Optional document id.  If provided and does not
+        start with ``_design/``, it will be prepended with ``_design/``.
+    """
+    def __init__(self, database, document_id=None):
+        super(CloudantDesignDocument, self).__init__(database, document_id)
```
alfinkel: I don't think you need the `__init__` constructor here since you're not performing any additional constructor logic.  The `__init__` should automatically inherit its behavior from DesignDocument.

src/cloudant/design_document.py
```diff
@@ -243,3 +207,128 @@ def info(self):
...
+        :param str view_name: Name used to identify the View.
+        :param str map_func: Javascript map function.
+        :param str reduce_func: Optional Javascript reduce function.
+        """
+        if self.get('language', None) == QUERY_LANGUAGE:
+            msg = ('Cannot add a MapReduce view to a '
+                   'design document for query indexes.')
+            raise CloudantException(msg)
+
+        super(CloudantDesignDocument, self).add_view(view_name, map_func,
```
alfinkel: I think this `super` call should be enclosed in an `else` block.

alfinkel:
I think this PR is missing some functionality specifically where we provide convenience methods like in the database object.

For example:

- [CouchDatabase.create_document](https://github.com/cloudant/python-cloudant/blob/cefd64747f5ea45ef9782f6776f3977edce817ca/src/cloudant/database.py#L123) should probably now have a corresponding `CloudantDatabase.create_document` method.  
- There are more of these types of methods in the database.py module...
- Obviously additional tests will need to be written as well.

You should probably do a search on `DesignDocument` within the library and check every place one is being constructed.  It would be a good bet that wherever that happens additional logic is needed to handle a `CloudantDesignDocument` as well.

src/cloudant/design_document.py
emlaver: After reviewing the link below and creating show/list functions in test design documents, I can access the `_list` and `_show` endpoints in both CouchDB and Cloudant:
http://docs.couchdb.org/en/1.6.1/couchapp/ddocs.html

src/cloudant/design_document.py
emlaver: Improved description in 45edead.

src/cloudant/design_document.py
emlaver: Fixed in 45edead.

CHANGES.rst
emlaver: Fixed in 45edead.

src/cloudant/design_document.py
emlaver: Fixed in 45edead.

src/cloudant/design_document.py
emlaver: Fixed in 45edead.

src/cloudant/design_document.py
emlaver: Fixed in 45edead.

PR #169, opened 2016-05-13
Cloudant SearchIndex class with search query execution support

emlaver:
## What

New `SearchIndex` class in `index.py` that exposes the search index function and analyzer as attributes. 
Next PR will implement search index management using `SearchIndex` in `DesignDocument` functions.

## How

- Add SearchIndex class to index module
- Property accessor for search index function and analyzer
- `CloudantDatabase.get_search_result` function for handling search query parameter processing and execution
- Use Document's encoder attribute when querying search endpoint with POST request

## Testing

Added `SearchIndex` unit tests in `cloudant.index_tests.SearchIndexTests`.
Added unit tests for search query parameter processing and execution in `CloudantDatabaseTests`.

## Issues

#158

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -24,7 +24,6 @@ import json
 
 from ._2to3 import STRTYPE, NONETYPE, UNITYPE, iteritems_
 from .error import CloudantArgumentError
-from cloudant.document import Document
 
 # Library Constants
 
@@ -124,7 +123,6 @@ SEARCH_INDEX_ARGS = {
     'group_sort': list,
     'include_docs': bool,
     'limit': (int, NONETYPE),
-    'q': (STRTYPE, int),
     'query': (STRTYPE, int),
     'ranges': dict,
     'sort': (STRTYPE, list),
@@ -235,26 +233,6 @@ def codify(code_or_str):
     if not isinstance(code_or_str, _Code):
         return _Code(code_or_str)
     return code_or_str
-
-def execute_search_or_find_query(db_or_ddoc, url, data):
-    """
-    Provides a helper function to execute a query against the _find or _search
-    endpoint.
-    """
-    if isinstance(db_or_ddoc, Document):
-        # Set encoder variable if db_or_ddoc is a DesignDocument
-        encoder = db_or_ddoc.encoder
-    else:
-        # Set encoder variable if db_or_ddoc is a CloudantDatabase
-        encoder = db_or_ddoc.client.encoder
-    headers = {'Content-Type': 'application/json'}
-    resp = db_or_ddoc.r_session.post(
-        url,
-        headers=headers,
-        data=json.dumps(data, cls=encoder)
-    )
-    resp.raise_for_status()
-    return resp.json()
 
 # Classes
 
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -24,6 +24,7 @@ import json
 
 from ._2to3 import STRTYPE, NONETYPE, UNITYPE, iteritems_
 from .error import CloudantArgumentError
+from cloudant.document import Document
 
 # Library Constants
 
@@ -113,6 +114,28 @@ QUERY_ARG_TYPES = {
 }
 
 TEXT_INDEX_ARGS = {'fields': list, 'default_field': dict, 'selector': dict}
+
+SEARCH_INDEX_ARGS = {
+    'bookmark': STRTYPE,
+    'counts': list,
+    'drilldown': list,
+    'group_field': STRTYPE,
+    'group_limit': (int, NONETYPE),
+    'group_sort': list,
+    'include_docs': bool,
+    'limit': (int, NONETYPE),
+    'q': (STRTYPE, int),
+    'query': (STRTYPE, int),
+    'ranges': dict,
+    'sort': (STRTYPE, list),
+    'stale': STRTYPE,
+    'highlight_fields': list,
+    'highlight_pre_tag': STRTYPE,
+    'highlight_post_tag': STRTYPE,
+    'highlight_number': (int, NONETYPE),
+    'highlight_size': (int, NONETYPE),
+    'include_fields': list
+}
 
 # Functions
 
@@ -212,6 +235,26 @@ def codify(code_or_str):
     if not isinstance(code_or_str, _Code):
         return _Code(code_or_str)
     return code_or_str
+
+def execute_search_or_find_query(db_or_ddoc, url, data):
+    """
+    Provides a helper function to execute a query against the _find or _search
+    endpoint.
+    """
+    if isinstance(db_or_ddoc, Document):
+        # Set encoder variable if db_or_ddoc is a DesignDocument
+        encoder = db_or_ddoc.encoder
+    else:
+        # Set encoder variable if db_or_ddoc is a CloudantDatabase
+        encoder = db_or_ddoc.client.encoder
+    headers = {'Content-Type': 'application/json'}
+    resp = db_or_ddoc.r_session.post(
+        url,
+        headers=headers,
+        data=json.dumps(data, cls=encoder)
+    )
+    resp.raise_for_status()
+    return resp.json()
 
 # Classes
 
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -234,6 +234,21 @@ def codify(code_or_str):
         return _Code(code_or_str)
     return code_or_str
 
+def search_url(index_name, design_document_url):
+    """
+    Constructs and returns the Cloudant Search URL.
+
+    :param index_name Name of the search index.
+    :param design_document_url Design document URL containing the search index.
+
+    :returns: Search URL
+    """
+    return '/'.join([
+        design_document_url,
+        '_search',
+        index_name
+    ])
+
 # Classes
 
 class _Code(str):
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -234,21 +234,6 @@ def codify(code_or_str):
         return _Code(code_or_str)
     return code_or_str
 
-def search_url(index_name, design_document_url):
-    """
-    Constructs and returns the Cloudant Search URL.
-
-    :param index_name Name of the search index.
-    :param design_document_url Design document URL containing the search index.
-
-    :returns: Search URL
-    """
-    return '/'.join([
-        design_document_url,
-        '_search',
-        index_name
-    ])
-
 # Classes
 
 class _Code(str):
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -21,15 +21,19 @@ import posixpath
 
 from requests.exceptions import HTTPError
 
-from ._2to3 import url_quote_plus
+from ._2to3 import url_quote_plus, iteritems_
+from ._common_util import (
+    JSON_INDEX_TYPE,
+    SEARCH_INDEX_ARGS,
+    SPECIAL_INDEX_TYPE,
+    TEXT_INDEX_TYPE,
+    search_url,
+    python_to_couch
+)
 from .document import Document
 from .design_document import DesignDocument
 from .view import View
 from .index import Index, TextIndex, SpecialIndex
-from ._common_util import JSON_INDEX_TYPE
-from ._common_util import TEXT_INDEX_TYPE
-from ._common_util import SPECIAL_INDEX_TYPE
-from ._common_util import python_to_couch
 from .query import Query
 from .error import CloudantException, CloudantArgumentError
 from .result import Result, QueryResult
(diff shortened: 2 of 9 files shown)


src/cloudant/index.py
```diff
@@ -20,7 +20,7 @@
 import json
 
 from ._2to3 import STRTYPE, iteritems_
-from ._common_util import JSON_INDEX_TYPE
+from ._common_util import JSON_INDEX_TYPE, codify, SEARCH_ANALYZERS
 from ._common_util import TEXT_INDEX_TYPE
 from ._common_util import SPECIAL_INDEX_TYPE
 from ._common_util import TEXT_INDEX_ARGS
```
alfinkel: Stylistically it is not consistent to have multiple imports from a module on the same line and then other imports individually imported from that same module.  I suggest changing L23-26 to:

```python
from ._common_util import (
    JSON_INDEX_TYPE, 
    TEXT_INDEX_TYPE , 
    SPECIAL_INDEX_TYPE , 
    TEXT_INDEX_ARGS, 
    SEARCH_ANALYZERS,
    codify
)
```

I think this looks cleaner.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
         """
         msg = 'Deleting the \"special\" index is not allowed.'
         raise CloudantException(msg)
+
+class SearchIndex(dict):
+    """
+    Encapsulates a SearchIndex as a dictionary based object, exposing the
+    search index function and analyzer as attributes.
+    """
```
alfinkel: I assume you will be adding more content here to mirror that of the [View](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/view.py#L28-L92) docstring?

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+    def __init__(
+            self,
+            ddoc,
+            search_index_name=None,
+            search_index=None
```
alfinkel: I think this has 1 level of indentation too much.  Remove 4 spaces.  I see that the `View.__init__()` has that as well.  Seems like that's where this originated.  It is unfortunate that pylint decided that it was not interested in catching that.  Anyway, can you create a separate PR to fix the `View.__init__()`?

alfinkel: Also, I think we're getting a little long with our variable/parm/arg names.  Let's just call it an `index_name` rather than `search_index_name` and I think `search_index` would be better named as `search_func` to sort mirror the `View` functions.  After all the `search_index` really is a function, right?

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+    ):
+        super(SearchIndex, self).__init__()
+        self.design_doc = ddoc
+        if search_index_name is not None:
+            self._search_index_name = search_index_name
```
alfinkel: I don't understand the purpose of this `if` block or the reason why you default the index name to `None` in the constructor signature.  It does not follow how it is implemented in a `View`.  Can a search index to exist without a name?

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+    search index function and analyzer as attributes.
+    """
+    def __init__(
+            self,
+            ddoc,
+            search_index_name=None,
+            search_index=None
+    ):
+        super(SearchIndex, self).__init__()
+        self.design_doc = ddoc
```
alfinkel: I'm assuming that you chose not to have a `self._r_session` here because you will be adding that as part of a future PR which makes this class callable?

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+        if search_index_name is not None:
+            self._search_index_name = search_index_name
+        if search_index is not None:
+            self['index'] = codify(search_index)
+        self['analyzer'] = codify('standard')
```
alfinkel: Why not allow the developer a chance to set the analyzer as part of the constructor?  You forcing anyone who wants to set the analyzer do it in 2 steps by not providing that option in the constructor.

src/cloudant/index.py
mikerhodes: You could probably even go for `name` and `func` given the namespace of search index is already there :)

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+        Provides an search index analyzer property accessor and setter.
+
+        For example:
+
+        .. code-block:: python
+
+            # Set the SearchIndex index property
+            search.analyzer = 'standard'
+
+            print search.analyzer
```
mikerhodes: You need not have a code example for a simple setter/getter pair. More useful would be directing the user to the possible values.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+
+        :param str index_func: Javascript search index function.
+
+        :returns: Codified search index function
```
mikerhodes: I don't think either of the `param` or `returns` is correct.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+        if search_index_name is not None:
+            self._search_index_name = search_index_name
+        if search_index is not None:
+            self['index'] = codify(search_index)
+        self['analyzer'] = codify('standard')
+
+    @property
+    def analyzer(self):
+        """
+        Provides an search index analyzer property accessor and setter.
```
mikerhodes: This method is just the setter. Docs should describe what it does, something like `Get the analyzer for this index`.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+        :param str index_func: Javascript search index function.
+
+        :returns: Codified search index function
+        """
+        return self.get('analyzer')
+
+    @analyzer.setter
+    def analyzer(self, analyzer=None):
+        """
+        Provides a search index analyzer property setter.
```
mikerhodes: Similar to above, `Set the analyzer for this index`.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+    def index(self):
+        """
+        Provides an search index property accessor and setter.
+
+        For example:
+
+        .. code-block:: python
+
+            # Set the SearchIndex index property
+            search.index = 'function (doc) {\n  index(\"default\", doc._id);\n}'
```
mikerhodes: Definitely keep this example, I think in this case the unusual nature of the value for this property means the example is worthwhilte. You can remove the `print` line though.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+            self['analyzer'] = codify(analyzer)
+        else:
+            raise CloudantArgumentError('The analyzer property must be a '
+                                        'valid string and supported by '
+                                        'Cloudant search')
+
+    @property
+    def index(self):
+        """
+        Provides an search index property accessor and setter.
```
mikerhodes: As above, shorten to `Get/Set the index for this...blah` as appropriate.

src/cloudant/index.py
```diff
@@ -262,3 +262,91 @@ def delete(self):
...
+    def index(self, index_func):
+        """
+        Provides a search index property setter.
+        """
+        self['index'] = codify(index_func)
+
+    @property
+    def index_name(self):
+        """
+        Displays the search index name.
```
mikerhodes: Returns, not displays.

tests/unit/index_tests.py
```diff
@@ -584,5 +585,134 @@ def test_delete_disabled(self):
             'Deleting the \"special\" index is not allowed.'
         )
 
+@unittest.skipUnless(
+    os.environ.get('RUN_CLOUDANT_TESTS') is not None,
```
mikerhodes: You've a double-negative here (unless/is not None). This would be simpler as `skipIf(os.environ.get('RUN_CLOUDANT_TESTS') is None,...`.

tests/unit/index_tests.py
```diff
@@ -584,5 +585,134 @@ def test_delete_disabled(self):
...
+        self.assertIsInstance(search['analyzer'], _Code)
+        self.assertEqual(
+            search['analyzer'],
+            'standard'
+        )
+
+        self.assertEqual(search, {
+            'index': 'function (doc) { index("default", doc._id); }}',
+            'analyzer': 'standard'
+        })
```
mikerhodes: You have a few branches in your setter, is it worth testing the `None` cases for the name and func args?

src/cloudant/index.py
alfinkel: The only reason not to do this is for consistency across the lib.  The `View` class has it as `view_name`, `map_func` and `reduce_func`.  I suppose that from the "function" perspective we could over look the consistency aspect since in the `View` there are two functions that need naming whereas in the `SearchIndex` there is only one so `func` would be fine there but the naming consistency of the view and the search index still exists.  EIther way this only affects the documentation and most likely no one on the planet would notice other than me.

tests/unit/index_tests.py
emlaver: This skip code is used in other test modules.  I would like to handle this change as a separate issue.

src/cloudant/index.py
emlaver: I've removed the `index_name` property.

src/cloudant/index.py
emlaver: Changed:
index property getter: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R323
index property setter:https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R341

src/cloudant/index.py
emlaver: Updated: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R302

src/cloudant/index.py
emlaver: Updated: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R316

src/cloudant/index.py
emlaver: Removed print in code block: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R329

src/cloudant/index.py
emlaver: Updated: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R309

src/cloudant/index.py
emlaver: Removed code block and added link to analyzer doc: https://github.com/cloudant/python-cloudant/pull/169/files#diff-eeec48d80ce666f16517361d09cf63f1R304

src/cloudant/index.py
```diff
@@ -262,3 +267,201 @@ def delete(self):
...
+    Encapsulates a SearchIndex as a dictionary based object, exposing the
+    search index function and analyzer as attributes. A SearchIndex object is
+    instantiated with a reference to a DesignDocument and is typically used as
+    part of the :class:`~cloudant.design_document.DesignDocument`
+    search index management API.
+
+    :param DesignDocument ddoc: DesignDocument instance used in part to
+        identify the search index.
+    :param str search_index_name: Name used in part to identify the index.
+    :param str search_index_func: Javascript search index function.
```
alfinkel: I like the new names (index_name and search_func) but the docstring needs to be fixed to match them.

src/cloudant/_common_util.py
```diff
@@ -213,6 +236,26 @@ def codify(code_or_str):
...
+        encoder = db_or_ddoc.client.encoder
+    headers = {'Content-Type': 'application/json'}
+    resp = db_or_ddoc.r_session.post(
+        url,
+        headers=headers,
+        data=json.dumps(data, cls=encoder)
+    )
+    resp.raise_for_status()
+    return resp.json()
+
```
alfinkel: I'm not a fan of this function.  Perhaps we can discuss this one when we talk later.

src/cloudant/index.py
```diff
@@ -262,3 +267,201 @@ def delete(self):
...
+        :param int highlight_number: A number of fragments returned in
+            highlights. If the search term occurs less often than the number of
+            fragments specified, longer fragments are returned.  Default is 1.
+        :param int highlight_size: A number of characters in each fragment for
+            highlights.  Defaults to 100 characters.
+        :param list include_fields: A list of field names to include in search
+            results. Any fields included must have been indexed with the
+            ``store:true`` option.
+
+        :returns: Search query result data in JSON format
```
alfinkel: I assume you pulled the param content from the Cloudant documentation directly?

src/cloudant/query.py
```diff
@@ -174,14 +173,7 @@ def __call__(self, **kwargs):
...
         # Execute query find
-        headers = {'Content-Type': 'application/json'}
-        resp = self._r_session.post(
-            self.url,
-            headers=headers,
-            data=json.dumps(data, cls=self._encoder)
-        )
-        resp.raise_for_status()
-        return resp.json()
+        return execute_search_or_find_query(self._database, self.url, data)
```
alfinkel: Like I mentioned earlier, I'm not a fan of this function but if you're going to use it then I would think that `self._r_session` would need to be removed.  I'm curious as to why pylint did not catch that.  Anyway, don't make any change to this now.  We can discuss this later as well.

src/cloudant/index.py
emlaver: Yes, I used the content from Cloudant documentation and made some changes to follow the docstring design in the Query and View callable function.

tests/unit/index_tests.py
```diff
@@ -584,5 +585,250 @@ def test_delete_disabled(self):
...
+        for group in resp['groups']:
+            for row in group['rows']:
+                self.assertEqual(row['fields'], {u'name': u'julia'})
+                self.assertTrue(row['id'].startswith('julia0'))
+                self.assertIsNotNone(row['order'])
+            self.assertEqual(group['total_rows'], 1)
+            self.assertIsNotNone(group['by'])
+
+
+
```
alfinkel: too many blank lines

src/cloudant/index.py
```diff
@@ -262,3 +267,201 @@ def delete(self):
...
+                raise CloudantArgumentError(msg)
+            if not isinstance(val, SEARCH_INDEX_ARGS[key]):
+                msg = (
+                    'Argument {0} is not an instance of expected type: {1}'
+                ).format(key, SEARCH_INDEX_ARGS[key])
+                raise CloudantArgumentError(msg)
+        if not kwargs.get('query'):
+            msg = (
+                'Null value or empty lucene search syntax in '
+                'the query parameter. Add a search query and retry.'
```
alfinkel: I don't think `Null` is the appropriate Python terminology.

src/cloudant/index.py
```diff
@@ -262,3 +267,201 @@ def delete(self):
...
+            search.index = 'function (doc) {  index(\"default\", doc._id); }'
+
+        :param str index_func: Javascript search index function.
+
+        :returns: Codified search index function
+        """
+        return self.get('index')
+
+    @index.setter
+    def index(self, index_func):
```
alfinkel: You probably want the `index_func` to be `search_func` to match what it is called in the constructor.

src/cloudant/index.py
emlaver: Fixed in 5de508f.

src/cloudant/_common_util.py
emlaver: Removed common util function in 5de508f.

src/cloudant/query.py
emlaver: Removed in 5de508f.

tests/unit/index_tests.py
emlaver: Fixed in 5de508f.

src/cloudant/index.py
emlaver: Fixed in 5de508f.

emlaver:
Updated description and added comment about `CloudantDatabase.get_search_result` function in #158

src/cloudant/document.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, database, document_id=None):
         self._document_id = document_id
         if self._document_id is not None:
             self['_id'] = self._document_id
-        self._encoder = self._client.encoder
+        self.encoder = self._client.encoder
```
alfinkel: All instances of `self.encoder` should be set back to `self._encoder` in the Document.  We talked about this.

src/cloudant/database.py
```diff
@@ -1098,3 +1102,124 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+                msg = (
+                    'Argument {0} is not an instance of expected type: {1}'
+                ).format(key, SEARCH_INDEX_ARGS[key])
+                raise CloudantArgumentError(msg)
+        # Execute query search
+        headers = {'Content-Type': 'application/json'}
+        resp = ddoc.r_session.post(
+            search_url(index_name, ddoc.document_url),
+            headers=headers,
+            data=json.dumps(data, cls=ddoc.encoder)
```
alfinkel: The database object has access to the client encoder.  Use that so that you don't have to make the Document encoder public.

src/cloudant/_common_util.py
```diff
@@ -213,6 +234,21 @@ def codify(code_or_str):
...
+    :param design_document_url Design document URL containing the search index.
+
+    :returns: Search URL
+    """
+    return '/'.join([
+        design_document_url,
+        '_search',
+        index_name
+    ])
+
```
alfinkel: I truly hate this function.

There, I've said it.  I feel much better now...

Remove this function and put it where it belongs, as a single statement of code in CloudantDatabase.get_search_result.  You tests also, shouldn't be using it.

src/cloudant/database.py
alfinkel: I actually question whether this method should even be accepting a DesignDocument as an argument.  It seems unnecessary and could be replaced by just a ddoc id.  By removing the callable from the SearchIndex you are basically removing the link between search index management (via DesignDocument) and search query functionality.  Which is what we discussed is the desired behavior.  That being the case, there really is no reason to mandate that a developer provide a full DesignDocument here when just a ddoc id will do.

If you are worried that there will be an inconsistency in the way that you generate the ddoc portion of the url, you can certainly just stand up a DesignDocument internally to this method call and pull the ddoc url from it yourself.  As in:

```python
def get_search_result(self, ddoc_id, index_name, query, **kwargs):
...
    ddoc = DesignDocument(self, ddoc_id)
    resp = ddoc.r_session.post(
        '/'.join([ddoc.document_url, '_search', index_name]),
        headers=headers,
        data=json.dumps(data, cls=self.client.encoder)
    )

File: src/cloudant/document.py (hunk 23 above)
Comment (emlaver) [inline src/cloudant/document.py]: Fixed in 2f14ef5.

File: src/cloudant/database.py (hunk 24 above)
Comment (emlaver) [inline src/cloudant/database.py]: Fixed in 2f14ef5.

File: src/cloudant/_common_util.py (hunk 25 above)
Comment (emlaver) [inline src/cloudant/_common_util.py]: Fixed in 2f14ef5.

File: src/cloudant/database.py [hunk 26]
```diff
@@ -1098,3 +1101,124 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+
+        .. code-block:: python
+
+            # Assuming that 'searchindex001' exists as part of the
+            # design document ddoc in the remote database...
+            # Retrieve documents where the Lucene field name is 'name' and
+            # the value is 'julia*'
+            query = 'name:julia*'
+            resp = db.get_search_result(ddoc, 'searchindex001', query,
+                                        include_docs=True)
```
alfinkel: I think your example needs to be reworked slightly now that you are using ddoc_id instead of a full DesignDocument (ddoc).

src/cloudant/database.py
```diff
@@ -1098,3 +1101,124 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+        the ``group_field`` parameter:
+
+        .. code-block:: python
+
+            # Assuming that 'searchindex001' exists as part of the
+            # design document ddoc in the remote database...
+            query = 'julia*'
+            # Retrieve JSON response content, limiting response to 10 documents
+            resp = db.get_search_result(ddoc, 'searchindex001',
+                                        group_field='name', limit=10)
```
alfinkel: same here as well.

src/cloudant/database.py
```diff
@@ -1098,3 +1101,124 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+            highlights.  Defaults to 100 characters.
+        :param list include_fields: A list of field names to include in search
+            results. Any fields included must have been indexed with the
+            ``store:true`` option.
+
+        :returns: Search query result data in JSON format
+        """
+        data = dict(self)
+        data.update(query=query)
+        data.update(kwargs)
```
alfinkel: What exactly is your intention with these 3 lines of code?  This is the a CloudantDatabase class, so `self` in this case would be the CloudantDatabase object.  Why would you be setting `data` to it?  I'm quite confused.

src/cloudant/index.py
```diff
@@ -262,3 +265,79 @@ def delete(self):
         """
         msg = 'Deleting the \"special\" index is not allowed.'
         raise CloudantException(msg)
+
+class SearchIndex(dict):
```
alfinkel: Are you planning on deleting this class and moving the codifying of the index function to the DesignDocument like we discussed?

src/cloudant/database.py
```diff
@@ -1098,3 +1101,124 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+                raise CloudantArgumentError(msg)
+            if not isinstance(val, SEARCH_INDEX_ARGS[key]):
+                msg = (
+                    'Argument {0} is not an instance of expected type: {1}'
+                ).format(key, SEARCH_INDEX_ARGS[key])
+                raise CloudantArgumentError(msg)
+        # Execute query search
+        headers = {'Content-Type': 'application/json'}
+        ddoc = DesignDocument(self, ddoc_id)
+        resp = ddoc.r_session.post(
```
alfinkel: Since you're in the CloudantDatabase, you should use the CloudantDatabase `r_session`.  The example I provided before should have used `self.r_session` instead of `ddoc.r_session`.

src/cloudant/index.py
emlaver: Yes, this will be deleted in the next PR.  I'll add a link here as a reference.

PR #171, opened 2016-05-20
Cloudant search index query

emlaver:
## What

Support Cloudant search query parameter processing and execution.

## How

- Handled search query parameter processing and execution in `CloudantDatabase.get_search_result`
- Added dictionary of search query parameters in common_util

## Testing

- Unit tests for search index query in `CloudantDatabaseTests` class
- Created a search index during test class setup for querying tests with `create_search_index`

## Reviewers

reviewer: @alfinkel 
reviewer @ricellis

## Issues

#158

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -21,15 +21,18 @@ import posixpath
 
 from requests.exceptions import HTTPError
 
-from ._2to3 import url_quote_plus
+from ._2to3 import url_quote_plus, iteritems_
+from ._common_util import (
+    JSON_INDEX_TYPE,
+    SEARCH_INDEX_ARGS,
+    SPECIAL_INDEX_TYPE,
+    TEXT_INDEX_TYPE,
+    python_to_couch
+)
 from .document import Document
 from .design_document import DesignDocument
 from .view import View
 from .index import Index, TextIndex, SpecialIndex
-from ._common_util import JSON_INDEX_TYPE
-from ._common_util import TEXT_INDEX_TYPE
-from ._common_util import SPECIAL_INDEX_TYPE
-from ._common_util import python_to_couch
 from .query import Query
 from .error import CloudantException, CloudantArgumentError
 from .result import Result, QueryResult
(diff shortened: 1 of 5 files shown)


src/cloudant/database.py
```diff
@@ -1098,3 +1101,123 @@ def get_query_result(self, selector, fields=None, raw_result=False,
             return QueryResult(query, **kwargs)
         else:
             return query.result
+
+    def get_search_result(self, ddoc_id, index_name, query, **kwargs):
+        """
+        Retrieves the raw JSON content from the remote database based on the
+        search index on the server, using the kwargs provided as query
+        parameters.
+
```
alfinkel: You mention using kwargs as query parameters.  You might also want to mention using `query` as well.  Just a thought.

tests/unit/database_tests.py
```diff
@@ -1065,5 +1066,108 @@ def test_get_query_indexes(self):
...
+    def test_get_search_result_with_invalid_argument(self):
+        """
+        Test get_search_result by passing in invalid arguments
+        """
+        try:
+            self.db.get_search_result('searchddoc001', 'searchindex001',
+                                      'julia*', foo={'bar': 'baz'})
+            self.fail('Above statement should raise an Exception')
+        except CloudantArgumentError as err:
+            self.assertEqual(str(err), 'Invalid argument: foo')
```
alfinkel: Rather than using the try/except block to test invalid conditions you should use `assertRaises`.  See this #80 for detail.  This applies to all new tests that you added as part of this PR.

tests/unit/database_tests.py
```diff
@@ -1065,5 +1066,108 @@ def test_get_query_indexes(self):
...
+            limit=5,
+            include_docs=True
+        )
+        self.assertEqual(len(resp['rows']), 5)
+        self.assertEqual(resp['total_rows'], 100)
+        for row in resp['rows']:
+            self.assertIsNotNone(row['fields'])
+            self.assertTrue(row['id'].startswith('julia0'))
+            self.assertIsNotNone(row['order'])
+            self.assertIsNotNone(row['doc'])
```
alfinkel: I don't really like stepping through the elements of the `resp` and performing asserts inside of the loop.  Would it be possible to compose an object that looks like what you expect the `resp` to look like based on the query call and then to use that object to do a direct comparison to the `resp`?  This is was what I've tried to do as much as possible in this library and what most of the other client libraries also attempt to do.  Same comment for the group_option test below.

pylintrc
```diff
@@ -235,7 +235,7 @@ notes=FIXME,XXX,TODO
 [SIMILARITIES]
 
 # Minimum lines number of a similarity.
-min-similarity-lines=5
+min-similarity-lines=10
```
alfinkel: Is this change necessary any longer not that we've remove a lot of the SearchIndex stuff?

tests/unit/database_tests.py
```diff
@@ -640,6 +640,7 @@ def setUp(self):
         """
         super(CloudantDatabaseTests, self).setUp()
         self.db_set_up()
+        self.create_search_index()
```
alfinkel: Is it worth having this in the `setUp` for 6 tests?  Maybe move it to the tests themselves?

src/cloudant/database.py
```diff
@@ -1098,3 +1101,123 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+        :param int highlight_size: A number of characters in each fragment for
+            highlights.  Defaults to 100 characters.
+        :param list include_fields: A list of field names to include in search
+            results. Any fields included must have been indexed with the
+            ``store:true`` option.
+
+        :returns: Search query result data in JSON format
+        """
+        data = dict(query=query)
+        data.update(kwargs)
```
alfinkel: Unfortunately, I believe that this creates a vulnerability where someone can provide a `query` as part of the `kwargs` and basically overwrite the `query` provided as a separate argument.  This could cause a lot of confusion for a developer that mistakenly falls into this trap.

I think it would be cleaner/safer to change the method signature to `def get_search_result(self, doc_id, index_name, **query_params)` and to then perform a check for `query` since it is mandatory.  As in:

```python
def get_search_result(self, doc_id, index_name, **query_params):
    ....
    if not query_params.get('query'):
        # raise CloudantArgumentError exception
    data = dict(query_params)
    ....
```

src/cloudant/database.py
```diff
@@ -1098,3 +1101,123 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+                    'Argument {0} is not an instance of expected type: {1}'
+                ).format(key, SEARCH_INDEX_ARGS[key])
+                raise CloudantArgumentError(msg)
+        # Execute query search
+        headers = {'Content-Type': 'application/json'}
+        ddoc = DesignDocument(self, ddoc_id)
+        resp = self.r_session.post(
+            '/'.join([ddoc.document_url, '_search', index_name]),
+            headers=headers,
+            data=json.dumps(data, cls=self.client.encoder)
```
alfinkel: As discussed over Slack, you no longer need `data` since you are not manipulating the query parameters any longer.  You can get rid of the `data` variable and use `query_params` directly here.

src/cloudant/database.py
emlaver: Changed method signature and removed `query` argument in 8e22bed.

tests/unit/database_tests.py
emlaver: Fixed in 8e22bed.

tests/unit/database_tests.py
emlaver: I've composed expected `resp` objects in 8e22bed.

pylintrc
emlaver: I've changed `min-similarity-lines` back to 5 in 8e22bed.

tests/unit/database_tests.py
emlaver: Only two unit tests require `create_search_index` so I've moved it in 8e22bed.

src/cloudant/database.py
emlaver: Changed method signature with updates to docstring in 8e22bed.

src/cloudant/database.py
emlaver: Fixed in 8e22bed.

alfinkel:
+1

src/cloudant/database.py
```diff
@@ -1098,3 +1101,128 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+            optional and defaults to number.
+        :param str stale: Optional string to allow the results from a stale
+            index to be used. This makes the request return immediately, even
+            if the index has not been completely built yet.
+        :param list highlight_fields: Optional list of fields which should be
+            highlighted.
+        :param str highlight_pre_tag: Optional string inserted before the
+            highlighted word in the highlights output.  Defaults to ``<em>``.
+        :param str highlight_post_tag: Optional string inserted after the
+            highlighted word in the highlights output.  Defaults to ``<em>``.
```
ricellis: Shouldn't this be `</em>`?

src/cloudant/database.py
```diff
@@ -1098,3 +1101,128 @@ def get_query_result(self, selector, fields=None, raw_result=False,
...
+                'containing Lucene syntax and retry.')
+
+        # Validate query arguments and values
+        for key, val in iteritems_(query_params):
+            if key not in list(SEARCH_INDEX_ARGS.keys()):
+                msg = 'Invalid argument: {0}'.format(key)
+                raise CloudantArgumentError(msg)
+            if not isinstance(val, SEARCH_INDEX_ARGS[key]):
+                msg = (
+                    'Argument {0} is not an instance of expected type: {1}'
```
ricellis: This message could be a little confusing for the cases where we accept multiple types. For instance `query` and `sort`. If query was the wrong type for example I think the message would read:
>Argument query is not an instance of expected type: `(<type 'basestring'>, <type 'int'>)`

I can see why is preferable to do this check using the tuple as it makes the code much simpler. I'm not sure saying "one of the expected types:" is really any better, but it might be a bit less ambiguous:
>Argument query is not an instance of one of the expected types: `(<type 'basestring'>, <type 'int'>)`

although this looks a bit weird in the singular case:
>Argument stale is not an instance of one of the expected types: `(<type 'basestring'>)`

tests/unit/database_tests.py
```diff
@@ -1065,5 +1065,156 @@ def test_get_query_indexes(self):
...
+    def test_get_search_result_with_invalid_value_types(self):
+        """
+        Test get_search_result by passing in invalid value types for
+        query parameters
+        """
+        test_data = [
+            {'bookmark': 1},                    # Should be a basestring
+            {'counts': 'blah'},                 # Should be a list
+            {'drilldown': 'blah'},              # Should be a list
+            {'group_field': ['blah']},          # Should be an basestring
```
ricellis: Looks like you missed a couple on copy/paste
a basestring

tests/unit/database_tests.py
```diff
@@ -1065,5 +1065,156 @@ def test_get_query_indexes(self):
...
+            {'group_limit': 'int'},             # Should be an int
+            {'group_sort': 'blah'},             # Should be a list
+            {'include_docs': 'blah'},           # Should be an boolean
```
ricellis: a boolean

src/cloudant/database.py
emlaver: Looks like it could be more common for a user to see the singular case knowing that 6 of the 18 query parameters accept multiple types.  If we change this instance of the message to `an instance of one of the expected types`, we would also want to consider changing the other instances in the project.

src/cloudant/database.py
emlaver: This default value was copied directly from the documentation:
https://docs.cloudant.com/search.html#queries

tests/unit/database_tests.py
emlaver: Fixed in 4e706c2.

tests/unit/database_tests.py
emlaver: Fixed in 4e706c2.

src/cloudant/database.py
ricellis: Yeah I see that the singular is more common, but to my mind even the slightly weird singular case:
>Argument stale is not an instance of one of the expected types: `(<type 'basestring'>)`

makes more sense than the reverse:
>Argument query is not an instance of expected type: `(<type 'basestring'>, <type 'int'>)`

I thought most of the other places in the codebase were singular, except when ints are expected so it is `(int, NONETYPE)`.
I guess an option would be to replace all the places these messages are generated with calls to a single utility method to format the message based on the number of types checked. If that is a better option I think that could happen under another issue/PR.

src/cloudant/database.py
ricellis: Surely that doc must be incorrect `<em>match<em>` would not be valid.

src/cloudant/database.py
emlaver: I think it's better to handle this under another issue.

alfinkel: btw - `basestring` does not exist in Python 3.

emlaver: Thanks @alfinkel. From the comment above, it's probably best to change the comments with `basestring` to:
```
# Should be a STRTYPE
```

src/cloudant/database.py
emlaver: You're correct, it's invalid.  Result from a quick curl test:
```
{..., "rows":[...,"highlights":{"name":["<em>aardvark</em>"]}}]}
```
Fixed in 90213c1.

tests/unit/database_tests.py
```diff
@@ -1065,5 +1065,156 @@ def test_get_query_indexes(self):
...
+        err = cm.exception
+        self.assertEqual(str(err), 'Invalid argument: foo')
+
+    def test_get_search_result_with_invalid_value_types(self):
+        """
+        Test get_search_result by passing in invalid value types for
+        query parameters
+        """
+        test_data = [
+            {'bookmark': 1},                    # Should be a basestring
```
ricellis: So is the plan to change these comments to say `a STRTYPE` now?

ricellis:
OK so I think we're agreed:

1. Open a new issue for replacing the type checking messages.
2. Fix the comments referring to `basestring` to `STRTYPE` to suit both python versions

+1

emlaver:
@ricellis Issue opened: https://github.com/cloudant/python-cloudant/issues/176

PR #172, opened 2016-05-23
Cloudant Search index management

### LINKED ISSUE #104: Add support for Cloudant Search

### LINKED ISSUE #159: Cloudant Search index management
This is the second part of three work items for Cloudant Search #104:
- Create search index management in design_document.py
    - Provide indexes accessor
    - add/update/delete/fetch/save/list/get functions using `SearchIndex` object
        - Throw exception if the database instance is CouchDatabase
    - iterate search indexes function

emlaver:
## What

Functions for managing Cloudant search index in design documents module. 

## How

- Added functions for search index management in `design_document.py`
- New indexes property for the locally cached DesignDocument in `design_document.py`

## Testing

Unit test cases to assert new search index management functions in `DesignDocumentTests`
Added `indexes` key with empty dictionary value to locally cached design documents in existing test case assertions.

## Issues

fixes #159 
fixes #104

### DIFF
--- a/src/cloudant/design_document.py
+++ b/src/cloudant/design_document.py
@@ -15,25 +15,25 @@
 """
 API module/class for interacting with a design document in a database.
 """
-from ._2to3 import iteritems_
+from ._2to3 import iteritems_, STRTYPE
+from ._common_util import QUERY_LANGUAGE, codify
 from .document import Document
 from .view import View, QueryIndexView
 from .error import CloudantArgumentError, CloudantException
-from ._common_util import QUERY_LANGUAGE
 
 class DesignDocument(Document):
     """
     Encapsulates a specialized version of a
     :class:`~cloudant.document.Document`.  A DesignDocument object is
     instantiated with a reference to a database and
-    provides an API to view management, list and show
-    functions, search indexes, etc.  When instantiating a DesignDocument or
+    provides an API to view management, index management, list and show
+    functions, etc.  When instantiating a DesignDocument or
     when setting the document id (``_id``) field, the value must start with
     ``_design/``.  If it does not, then ``_design/`` will be prepended to
     the provided document id value.
 
-    Note:  Currently only the view management API exists.  Remaining design
-    document functionality will be added later.
+    Note:  Currently only the view management and search index management API
+    exists.  Remaining design document functionality will be added later.
 
     :param database: A database instance used by the DesignDocument.  Can be
         either a ``CouchDatabase`` or ``CloudantDatabase`` instance.
@@ -45,6 +45,7 @@ class DesignDocument(Document):
             document_id = '_design/{0}'.format(document_id)
         super(DesignDocument, self).__init__(database, document_id)
         self.setdefault('views', dict())
+        self.setdefault('indexes', dict())
 
     @property
     def views(self):
@@ -55,6 +56,17 @@ class DesignDocument(Document):
         :returns: Dictionary containing view names and View objects as key/value
         """
         return self.get('views')
+
+    @property
+    def indexes(self):
+        """
+        Provides an accessor property to the indexes dictionary in the
+        locally cached DesignDocument.
+
+        :returns: Dictionary containing index names and index objects
+            as key/value
+        """
+        return self.get('indexes')
 
     def add_view(self, view_name, map_func, reduce_func=None, **kwargs):
         """
@@ -78,6 +90,26 @@ class DesignDocument(Document):
 
         view = View(self, view_name, map_func, reduce_func, **kwargs)
         self.views.__setitem__(view_name, view)
+
+    def add_search_index(self, index_name, search_func, analyzer=None):
+        """
+        Appends a Cloudant search index to the locally cached DesignDocument
+        indexes dictionary.
+
+        :param str index_name: Name used to identify the search index.
+        :param str search_func: Javascript search index function.
+        :param analyzer: Optional analyzer for this search index.
+        """
+        if self.get_index(index_name) is not None:
+            msg = ('An index with name {0} already exists in this design doc'
+                   .format(index_name))
+            raise CloudantArgumentError(msg)
+        if analyzer is not None:
+            search = {'index': codify(search_func), 'analyzer': analyzer}
+        else:
+            search = {'index': codify(search_func)}
+
+        self.indexes.__setitem__(index_name, search)
 
     def update_view(self, view_name, map_func, reduce_func=None, **kwargs):
         """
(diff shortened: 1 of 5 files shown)


src/cloudant/design_document.py
```diff
@@ -56,6 +57,17 @@ def views(self):
...
+    @property
+    def search_indexes(self):
+        """
+        Provides an accessor property to the indexes dictionary in the
+        locally cached DesignDocument.
+
+        :returns: Dictionary containing search index names and search functions
+            as key/value
+        """
+        return self.get('indexes')
```
emlaver: @ricellis you said: "If this is retrieving the indexes dictionary entry there is no guarantee they will only be search indexes is there? So shouldn't it just be indexes?"

Query indexes are not handled in the `DesignDocument` class.  I think it may be misleading to change this to `indexes` since we are only supporting search indexes.

src/cloudant/design_document.py
```diff
@@ -16,18 +16,18 @@
...
-from ._common_util import QUERY_LANGUAGE
 
 class DesignDocument(Document):
     """
     Encapsulates a specialized version of a
     :class:`~cloudant.document.Document`.  A DesignDocument object is
     instantiated with a reference to a database and
-    provides an API to view management, list and show
-    functions, search indexes, etc.  When instantiating a DesignDocument or
+    provides an API to view management, search index management, list and show
```
emlaver: @ricellis said: "shouldn't this just be index management?
The design document can contain query indexes as well."

Query index management is not handled in `DesignDocument` class. 
Create/delete/etc query index functions are handled in `index` module and the user would call these functions in `CloudantDatabase` (i.e. `CloudantDatabase.create_query_index`)

ricellis:
As discussed I think the index functions on design_document need to be index type agnostic. If we need additional helpers to make it easier to create the index functions to pass into the design_document then those functions should go elsewhere on something search specific.
So for example so if we had `add_search_index(name, func, analyzer)` it should probably be on a search specific utility and create an index object that got passed to `design_document.add_index()`. It may be easier however just to allow the `add_index` method to accept other arguments that would get added as JSON to the parent index object.

src/cloudant/design_document.py
ricellis: How do you guarantee that the `design_document` class is only used for search indexes?
What behaviour are you expecting if I do:
1. `database.create_query_index('myDesignDoc','myIndex' ... )`
2. `database.get_design_document('myDesignDoc').search_indexes`
?

Isn't that is going to return the (query) index called `myIndex` even though I have called a method called `search_indexes`?

emlaver: Renamed `search_indexes` to `indexes` in 39f991e.

ricellis:
I'm wondering about the use of `analyser='standard'` as the default argument for the search index methods. I thought we usually took the approach of not specifying a value for a parameter if it wasn't provided - that is we omit it and allow the server default to apply.

How do you feel about using  `analyzer=None` with a check before adding it to the dict?

src/cloudant/design_document.py
```diff
@@ -225,6 +333,15 @@ def list_views(self):
         """
         return list(self.views.keys())
 
+    def list_indexes(self):
+        """
+        Retrieves a list of available search indexes in the locally
+        cached DesignDocument indexes dictionary.
+
+        :returns: List of search index names
```
ricellis: This doc block has refers to "search indexes" and returns "search index" names. I think the word search should be removed from both these places since it will list any indexes right?

src/cloudant/design_document.py
```diff
@@ -216,6 +306,24 @@ def iterviews(self):
...
+
+        For example:
+
+        .. code-block:: python
+
+            for index_name, search_func in ddoc.iterindexes():
+                # Perform search index processing
+
+        :returns: Iterable containing search index name and associated
+            search function object
```
ricellis: This doc block refers to search indexes. I think the example could still be for search indexes, but the returns statement ought to be corrected.

src/cloudant/design_document.py
```diff
@@ -56,6 +57,17 @@ def views(self):
...
         return self.get('views')
 
+    @property
+    def indexes(self):
+        """
+        Provides an accessor property to the indexes dictionary in the
+        locally cached DesignDocument.
+
+        :returns: Dictionary containing search index names and search functions
+            as key/value
```
ricellis: This returns statement needs updating as it refers to search.

ricellis:
Do you think there is any value in also doing the language check during the `add_search_index` and `update_search_index` methods? If the design doc is of the wrong language type we could fail earlier than `save`. I guess we probably ought to still check during save anyway, only because someone might modify the dict directly without using the add/update functions, so maybe it isn't worth the extra code, but something to consider.

src/cloudant/design_document.py
```diff
@@ -79,6 +91,23 @@ def add_view(self, view_name, map_func, reduce_func=None, **kwargs):
...
+        """
+        Appends a Cloudant search index to the locally cached DesignDocument
+        indexes dictionary.
+
+        :param str index_name: Name used to identify the search index.
+        :param str search_func: Javascript search index function.
+        :param analyzer: Optional analyzer for this search index.
+        """
+        if self.get_index(index_name) is not None:
+            msg = ('Search index {0} already exists in this design doc'
```
ricellis: I think this message should be less specific e.g. `An index with name {0} already...` I don't think we have any guarantee that the existing index is a search index. Possibly wouldn't be necessary if a language check happened first and threw a different exception.

tests/unit/design_document_tests.py
```diff
@@ -626,5 +687,355 @@ def test_get_info(self):
...
+        Ensure that the document fetched from the database returns the
+        DesignDocument format as expected when retrieving a design document
+        containing no search indexes.
+        """
+        ddoc = DesignDocument(self.db, '_design/ddoc001')
+        ddoc.save()
+        ddoc_remote = DesignDocument(self.db, '_design/ddoc001')
+        ddoc_remote.fetch()
+        self.assertEqual(set(ddoc_remote.keys()),
+                         {'_id', '_rev', 'indexes', 'views'})
```
ricellis: It's a bit confusing why the `ddoc_remote.keys()` contains an `indexes` entry if the ddoc has no indexes. Is that the one we create in the locally cached design document that gets stripped on save if it is empty? If so perhaps a comment to that effect in this test might help someone unfamiliar with the code understand this oddity.

src/cloudant/design_document.py
emlaver: Fixed in 45a4667.

tests/unit/design_document_tests.py
emlaver: Fixed in 45a4667.

src/cloudant/design_document.py
emlaver: Fixed in 45a4667.

src/cloudant/design_document.py
emlaver: Fixed in 45a4667.

src/cloudant/design_document.py
emlaver: Fixed in 45a4667.

emlaver:
@ricellis I've updated the `analyzer` argument value to `None` and updated the test cases in a137e9a.

ricellis:
+1

rhyshort:
+1

Issue #161, opened 2016-05-02
Update readthedocs link in README.rst


alfinkel:
The current link to the library documentation in the README.rst is set to `http://python-cloudant.readthedocs.org`.  This resolves to `http://python-cloudant.readthedocs.org/index.html` which is actually our index module documentation.  When a user clicks this link we want it to go to the documentation home page which is currently `default.html`.  We may want to consider either changing the link to specifically point to `http://python-cloudant.readthedocs.org/default.html` or renaming index.html to something different and renaming default.html to index.html.

ricellis:
See also #155 if we're updating doc links.

emlaver:
The links are still pointing to the index module `index.html` in these and probably other locations:
- Github `Cloudant Python client library docs (readthedocs.io)` link under `Related Documentation`
- Links under `Short URLs` header: https://readthedocs.org/projects/python-cloudant/

Another PR will be opened to fix this.  The `default.html` will be renamed to `index.html` so that `http://python-cloudant.readthedocs.io` will resolve to the library documentation home page.  The current `index.html` containing index module documentation will be renamed to `module_index.html` or something similar.

PR #179, opened 2016-06-02
Update library home documentation page and README readthedocs link

emlaver:
## What

The readthedocs links are pointing to the `index.html` index module page in these and probably other locations:

- GitHub `Cloudant Python client library docs (readthedocs.io)` link under `Related Documentation`
- Links under `Short URLs` header: https://readthedocs.org/projects/python-cloudant/

The `default.html` is now `index.html` so that http://python-cloudant.readthedocs.io will resolve to the library documentation home page. The `index.html` index module documentation page is now `module_index.html`.

## How

- Renamed default.html library documentation page to index.html 
- Renamed index module index.html to module_index.html
- Fixed README.rst readthedocs link to use http://python-cloudant.readthedocs.io which resolves to http://python-cloudant.readthedocs.io/index.html

## Testing

HTML links are opening correctly in a local sphinx build. 
Run a local build by running these commands in the `docs` directory:
1. `make clean`
2. `make html`
Open the home documentation page `index.html` under `docs/_build/html`.  

## Issues

#161


ricellis:
+1

alfinkel:
This all looks right to me.  I know that in the past I was able to generate documentation on readthedocs using a specific repo branch which I think would be useful in this case to validate that the behavior you are looking for is in fact there on readthedocs.  Unfortunately, I don't see this branch listed in the versions tab on python-cloudant readthedocs so I'm not sure if that feature was retired or not there.  It might be worth poking around there to see if you can make that happen just to confirm that things are the way they should be before doing another release.

If you find that you cannot generate docs from the branch then I would suggest generating the docs from the master branch as you normally do as part of the release process, only I would generate those docs before cutting the new release in GH and pipit, this way if problems do arise you won't have to bump up yet another release number.  Hope that makes sense.

+1

PR #181, opened 2016-06-20
Rewrite rules property in design_document.py

### LINKED ISSUE #175: Support `rewrites` in design document
Design document can contain rules for URL rewriting, by using an array in the `rewrites` field.  Each rule is a JSON object with 4 fields: `from`, `to`, `method`, and `query`.
Example of rule in `rewrites`:
```
"rewrites": [
  {
    "from": "/foo/:var",
    "to": "/foo",
    "method": "GET",
    "query": {"v": "var"}
  }
]
```
Documentation: https://docs.cloudant.com/design_documents.html#rewrite-rules
Subitem part of #137

emlaver:
## What

Support URL rewriting with ``rewrites`` property in design document module. Documentation and an example are included with the property accessor.

## How

- Added ``rewrite`` accessor property with docstring that includes documentation

## Testing

Added test case for using ``rewrite`` property in DesignDocumentTests.
Created local sphinx build to review documentation.

## Reviewers

@tomblench

## Issues
fixes #175

### DIFF
--- a/src/cloudant/design_document.py
+++ b/src/cloudant/design_document.py
@@ -48,6 +48,44 @@ class DesignDocument(Document):
         self.setdefault('indexes', dict())
 
     @property
+    def rewrites(self):
+        """
+        Provides an accessor property to a list of dictionaries with rewrite
+        rules in the locally cached DesignDocument.  Each rule for URL rewriting
+        is a JSON object with four fields: ``from``, ``to``, ``method``,
+        and ``query``.
+
+        Note: Requests that match the rewrite rules must have a URL path that
+        starts with ``/$DATABASE/_design/doc/_rewrite``.
+
+        Rewrite rule example:
+
+        .. code-block:: python
+
+            # Add the rule to ``rewrites`` and save the design document
+            ddoc = DesignDocument(self.db, '_design/ddoc001')
+            ddoc['rewrites'] = [
+                {"from": "/old/topic",
+                 "to": "/new/",
+                 "method": "GET",
+                 "query": {}
+                 }
+            ]
+            ddoc.save()
+
+        Once the rewrite rule is saved to the remote database, the GET
+        request URL ``/$DATABASE/_design/doc/_rewrite/old/topic?k=v`` would be
+        rewritten as ``/$DATABASE/_design/doc/_rewrite/new?k=v``.
+
+        For more details on URL rewriting, see the `rewrite rules
+        documentation <https://docs.cloudant.com/design_documents.html
+        #rewrite-rules>`_.
+
+        :returns: List of dictionaries containing rewrite rules as key/value
+        """
+        return self.get('rewrites')
+
+    @property
     def views(self):
         """
         Provides an accessor property to the View dictionary in the locally
--- a/tests/unit/design_document_tests.py
+++ b/tests/unit/design_document_tests.py
@@ -1035,5 +1035,31 @@ class DesignDocumentTests(UnitTestDbBase):
                 '{"store": true}); }\n}'}
         )
 
+    def test_rewrite_rule(self):
+        """
+        Test that design document URL is rewritten to the expected test document.
+        """
+        ddoc = DesignDocument(self.db, '_design/ddoc001')
+        ddoc['rewrites'] = [
+            {"from": "",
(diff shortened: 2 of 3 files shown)


src/cloudant/design_document.py
```diff
@@ -48,6 +48,44 @@ def __init__(self, database, document_id=None):
...
+                 "to": "/new/",
+                 "method": "GET",
+                 "query": {}
+                 }
+            ]
+            ddoc.save()
+
+        Once the rewrite rule is saved to the remote database, the GET
+        request URL ``/$DATABASE/_design/doc/_rewrite/old/topic?k=v`` would be
+        rewritten as ``/$DATABASE/_design/doc/_rewrite/new/k=v``.
```
ricellis: I appreciate that this sample is straight from the Cloudant docs, but is it actually correct? Why would the k=v parameters become part of the path?
Shouldn't it be:
`/$DATABASE/_design/doc/_rewrite/old/topic?k=v`
`/$DATABASE/_design/doc/_rewrite/new/?k=v` (which admittedly looks wrong because of the trailing / in `/new/`). In fact some of the couch examples show trailing slashes being omitted, so maybe it will just be rewritten as:
`/$DATABASE/_design/doc/_rewrite/new?k=v`

tests/unit/design_document_tests.py
```diff
@@ -1035,5 +1035,50 @@ def test_get_search_index(self):
...
+             }
+        ]
+        self.assertIsInstance(ddoc.rewrites, list)
+        self.assertIsInstance(ddoc.rewrites[0], dict)
+        ddoc.save()
+        ddoc_remote = DesignDocument(self.db, '_design/ddoc001')
+        ddoc_remote.fetch()
+        self.assertIsInstance(ddoc_remote.rewrites, list)
+        self.assertIsInstance(ddoc_remote.rewrites[0], dict)
+        self.assertEqual(ddoc_remote, {
```
tomblench: Is it possible to do an assert on the final url to check that it gets rewritten correctly?

src/cloudant/design_document.py
emlaver: You're correct.  After additional tests, the documentation needs to be updated.  Here are the steps I did and results using `animaldb` sample database:
First, set `rewrites` property: 
```
 {
       "from": "/latin_name/",
       "to": "/new/latin_name",
       "method": "GET",
       "query": {
       }
   }
```
Then, run either curl command: 
`curl http://localhost:5984/animaldb/_design/views101/_rewrite/latin_name/?limit=1`
Or
`curl http://localhost:5984/animaldb/_design/views101/_rewrite/latin_name?limit=1`
Finally, the rewrite result:
`/animaldb/_design/views101/new/latin_name?limit=1`

tests/unit/design_document_tests.py
emlaver: @tomblench Yes, I believe that test should be replaced with the code below:
```
    def test_fetch_rewrite_rule(self):
        """
        Test that original URL is rewritten to the expected test document.
        """
        ddoc = DesignDocument(self.db, '_design/ddoc001')
        ddoc['rewrites'] = [
            {"from": "",
             "to": "/../../rewrite_doc",
             "method": "GET",
             "query": {}
             }
        ]
        self.assertIsInstance(ddoc.rewrites, list)
        self.assertIsInstance(ddoc.rewrites[0], dict)
        ddoc.save()
        doc = Document(self.db, 'rewrite_doc')
        doc.save()
        resp = requests.get('/'.join([ddoc.document_url, '_rewrite']))
        self.assertEquals(
            resp.json(),
            {
                '_id': 'rewrite_doc',
                '_rev': doc['_rev']
            }
        )
```

src/cloudant/design_document.py
emlaver: Fixed in 685dbf0.

tests/unit/design_document_tests.py
emlaver: Updated test case in 685dbf0.

ricellis:
+1

tomblench:
+1

emlaver:
@ricellis FB ticket 69184 created for updating rewrite rule documentation.

PR #185, opened 2016-06-29
Move duplicated code to common_util.py

### LINKED ISSUE #170: Add encoder to POST/GET requests
The database/document encoder attribute is used in Query's `__call__` function and should be added in other functions requiring the request library i.e. the `View` class and `CouchDatabase` class.

### LINKED ISSUE #182: Move duplicated code for GET/POST request to internal common_util.py
There's an almost identical block of code in `View.__call__` and `database.all_docs`:
https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L366-L375.  
This code should be moved into the `common_util` module.  It will also be used by future sub-objects in design document #137.

emlaver:
## What

There's an almost identical block of code in `View.__call__` and `database.all_docs` that's been moved into `get_docs` method in the common_util module.
The new common util `get_docs` method also accepts `encoder` for use with a custom ``JSONEncoder`` subclass.
## How

- Moved duplicated code for GET/POST request to internal method `get_docs`
- Changed visibility of `Document` attribute `encoder` for use in  `View.__call__`

## Testing

No new test cases required. 

## Issues

fixes #182 
fixes #170

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -234,6 +234,31 @@ def codify(code_or_str):
         return _Code(code_or_str)
     return code_or_str
 
+def get_docs(r_session, url, encoder=None, **params):
+    """
+    Provides a helper for functions that require GET or POST requests
+    with a JSON, text, or raw response containing documents.
+
+    :param r_session: Authentication session from the client
+    :param str url: URL containing the endpoint
+    :param JSONEncoder encoder: Custom encoder from the client
+
+    :returns: Raw response content from the specified endpoint
+    """
+    keys_list = params.pop('keys', None)
+    keys = None
+    if keys_list:
+        keys = json.dumps({'keys': keys_list}, cls=encoder)
+    f_params = python_to_couch(params)
+
+    resp = None
+    if keys:
+        resp = r_session.post(url, params=f_params, data=keys)
+    else:
+        resp = r_session.get(url, params=f_params)
+    resp.raise_for_status()
+    return resp
+
 # Classes
 
 class _Code(str):
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -27,8 +27,7 @@ from ._common_util import (
     SEARCH_INDEX_ARGS,
     SPECIAL_INDEX_TYPE,
     TEXT_INDEX_TYPE,
-    python_to_couch
-)
+    get_docs)
 from .document import Document
 from .design_document import DesignDocument
 from .view import View
@@ -363,16 +362,10 @@ class CouchDatabase(dict):
         :returns: Raw JSON response content from ``_all_docs`` endpoint
 
         """
-        all_docs_url = posixpath.join(self.database_url, '_all_docs')
-        params = python_to_couch(kwargs)
-        keys_list = params.pop('keys', None)
-        resp = None
-        if keys_list:
-            keys = json.dumps({'keys': keys_list})
-            resp = self.r_session.post(all_docs_url, params=params, data=keys)
-        else:
-            resp = self.r_session.get(all_docs_url, params=params)
-        resp.raise_for_status()
+        resp = get_docs(self.r_session,
+                        '/'.join([self.database_url, '_all_docs']),
+                        self.client.encoder,
+                        **kwargs)
         return resp.json()
 
     @contextlib.contextmanager
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -64,7 +64,7 @@ class Document(dict):
         self._document_id = document_id
         if self._document_id is not None:
             self['_id'] = self._document_id
-        self._encoder = self._client.encoder
+        self.encoder = self._client.encoder
 
     @property
     def document_url(self):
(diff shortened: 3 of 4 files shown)


src/cloudant/_common_util.py
```diff
@@ -234,6 +234,24 @@ def codify(code_or_str):
         return _Code(code_or_str)
     return code_or_str
 
+def _get_docs(url, **kwargs):
+    """
+    Provides a helper to get a response with documents for GET/POST requests.
```
ricellis: Reading this comment I wouldn't know when it was ok to use this utility (even internally).

ricellis:
+1 with a bit more explanation around the `_get_docs`

alfinkel:
I don't think you are using the naming standard correctly with `_get_docs`.  The leading underscore implies an private function.  `_get_docs` is called outside of the  `_common_util` module.  Suggest changing the name of the function.

emlaver:
@alfinkel Fixed naming to `get_docs` in 98b9287

src/cloudant/_common_util.py
```diff
@@ -234,6 +234,24 @@ def codify(code_or_str):
...
     return code_or_str
 
+def get_docs(url, **kwargs):
+    """
+    Provides a helper to get a response with documents for GET/POST requests.
+    """
+    r_session = kwargs.pop('r_session', None)
+    encoder = kwargs.pop('encoder', None)
+
+    params = python_to_couch(kwargs)
```
alfinkel: This function will not work without a valid session object, that being the case why not specify it in the function signature as a mandatory argument that way you would get an error straight away if there is no session provided?  Also I feel like the `encoder` should explicitly be called out in the signature as well rather than including it as part of **kwargs.  `kwargs` also is not very expressive as to what they are.  My suggestion would be to change lines 237-244 above to:

```python
def get_docs(r_session, url, encoder=None, **params):
    """
    Provides a helper to get a response with documents for GET/POST requests.
    """
    params = python_to_couch(params)
```

src/cloudant/_common_util.py
```diff
@@ -234,6 +234,24 @@ def codify(code_or_str):
...
+    keys_list = params.pop('keys', None)
+    resp = None
+    if keys_list:
+        keys = json.dumps({'keys': keys_list}, cls=encoder)
```
alfinkel: Looking at this code now I see that we have a possibility for a problem down the road.  Right now you are sending the `params` to be processed by `python_to_couch` function and right now it skips `keys` and does not perform any translation or conversion to the `keys` key or its value.  So you can get away with doing this `json.dumps` statement here.  If at any point the `python_to_couch` code changes and begins to process `keys` then you may run into a situation where you are performing multiple `json.dumps` commands on `keys`.

I know that you moved the code as it was originally but I think that had this same flaw.  This is probably a good time to correct that flaw and ensure that you pop off `keys` from `params` and handle `keys` as you do here and then pass `params` to `python_to_couch`.

src/cloudant/database.py
```diff
@@ -363,16 +362,10 @@ def all_docs(self, **kwargs):
...
-        params = python_to_couch(kwargs)
-        keys_list = params.pop('keys', None)
-        resp = None
-        if keys_list:
-            keys = json.dumps({'keys': keys_list})
-            resp = self.r_session.post(all_docs_url, params=params, data=keys)
-        else:
-            resp = self.r_session.get(all_docs_url, params=params)
-        resp.raise_for_status()
+        resp = get_docs(posixpath.join(self.database_url, '_all_docs'),
```
alfinkel: Opportunity to switch from using `posixpath.join` to `'/'.join`

src/cloudant/_common_util.py
emlaver: Updated function signature in 1d2df70.

src/cloudant/_common_util.py
```diff
@@ -234,6 +234,28 @@ def codify(code_or_str):
         return _Code(code_or_str)
     return code_or_str
 
+def get_docs(r_session, url, encoder=None, **kwargs):
```
alfinkel: Any thoughts on renaming `kwargs` to `params` as described [here](https://github.com/cloudant/python-cloudant/pull/185#discussion_r69210554)?

src/cloudant/_common_util.py
```diff
@@ -234,6 +234,28 @@ def codify(code_or_str):
...
+    :returns: Raw response content from the specified endpoint
+    """
+    keys_list = kwargs.pop('keys', None)
+    params = python_to_couch(kwargs)
+    resp = None
+    if keys_list:
+        keys = json.dumps({'keys': keys_list}, cls=encoder)
+        resp = r_session.post(url, params=params, data=keys)
+    else:
+        resp = r_session.get(url, params=params)
```
alfinkel: You did not address [this](https://github.com/cloudant/python-cloudant/pull/185/files/98b9287bd549dc0e90da11cd55e03e2dfed484ee#r69211986).  Either provide a reason not to do it or make the change.  I'm fine with either.

src/cloudant/_common_util.py
emlaver: Renamed argument name for clarify in 4dcfbbe.

src/cloudant/_common_util.py
emlaver: Made the necessary changes in 4dcfbbe from our discussion on Slack.

alfinkel:
+1

PR #187, opened 2016-07-06
Made 429 replays optional and configurable

ricellis:
## What

Made 429 replays optional and configurable.

## Why

Some users would prefer to receive an error earlier than the current 429 replay behaviour.

## How

Added support for `adapter` argument on client constructors.
Moved 429 `HTTPAdapter` configuration into `Replay429Adapter` class in
`cloudant.adapters`.
Added `retries` and `initialBackoff` args to `Replay429Adapter`.
Updated docs including `CHANGES.rst` and for new `adapters` module.

## Testing
Added tests for init and adding adapter with `Replay429Adapter` and args.
Manually tested that the behaviour is as expected with a 429 response.

### DIFF
--- a/tests/unit/adapter_tests.py
+++ b/tests/unit/adapter_tests.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright © 2016 IBM Corp. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from cloudant.client import CouchDB
+from cloudant.adapters import Replay429Adapter
+from requests.packages.urllib3.util import Retry
+from .unit_t_db_base import UnitTestDbBase
+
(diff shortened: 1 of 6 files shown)


tests/unit/adapter_tests.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Test that a new Replay429Adapter is accepted as a parameter for a client.
+        """
+        try:
+            self.client = CouchDB(
+                self.user,
+                self.pwd,
+                url=self.url,
+                adapter=Replay429Adapter())
+        except Exception as err:
```
emlaver: If possible, we should use `assertRaises` for test cases: https://github.com/cloudant/python-cloudant/issues/80

ricellis: Well the tests are not asserting that an exception is raised in this case. The objective was to fail if an exception is raised. In b6284d2 I've removed the try/except entirely - the test will still not pass in the case of an exception, but will be in `Error` state instead of failing.

emlaver:
LGTM +1

tests/unit/adapter_tests.py
```diff
@@ -0,0 +1,66 @@
...
+from cloudant.adapters import Replay429Adapter
+from requests.packages.urllib3.util import Retry
+from .unit_t_db_base import UnitTestDbBase
+
+class AdapterTests(UnitTestDbBase):
+    """
+    Unit tests for transport adapters
+    """
+
+    def test_new_Replay429Adapter(self):
```
tomblench: I'm a bit confused about what these tests are here for, except to show the different options? Should there be some asserts?

ricellis: It was basically to demonstrate that the different options are usable in the constructor as a protection against regressing the init.
The values get passed into requests/urllib3 so I didn't assert anything on them because it would require knowing something about where they end up in those libs.

tomblench:
+1

PR #189, opened 2016-07-08
Added support for list functions

### LINKED ISSUE #178: Support list functions in design document
Add support for list functions under the `lists` key in the design document module. In Database class, you could have a `query_list(list_name, ddoc_id, mr_index, **kwargs)` function to query the _list endpoint.

This is a subitem part of #137

emlaver:
## What

Add support for managing and querying list functions.
Avoid repeating if statements for each property in design document’s `save` and `fetch` functions by looping through a set containing the properties.

## How

- Handled list function execution against the _list endpoint in CouchDatabase.get_list_result
- Added support for managing list functions in design document
- Added HTTP Headers argument to `get_docs` method signature
- Replaced repeating if statements for each property with for-loop over a set of all properties

## Testing

Unit test cases to assert new list function management in DesignDocumentTests.
Unit test cases to assert `get_list_result` in DatabaseTests.

## Issues
fixes #178

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -762,6 +762,49 @@ class CouchDatabase(dict):
 
         return resp.json()
 
+    def get_list_function_result(self, ddoc_id, list_name, view_name, **kwargs):
+        """
+        Retrieves a customized MapReduce view result from the specified
+        database based on the list function provided.  List functions are
+        used, for example,  when you want to access Cloudant directly
+        from a browser, and need data to be returned in a different
+        format, such as HTML.
+
+        Note: All query parameters for View requests are supported.
+        See :class:`~cloudant.database.get_view_result` for
+        all supported query parameters.
+
+        For example:
+
+        .. code-block:: python
+
+            # Assuming that 'view001' exists as part of the
+            # 'ddoc001' design document in the remote database...
+            # Retrieve documents where the list function is 'list1'
+            resp = db.get_list_result('ddoc001', 'list1', 'view001', limit=10)
+            for row in resp['rows']:
+                # Process data (in text format).
+
+        For more detail on list functions, refer to the
+        `Cloudant documentation <https://docs.cloudant.com/
+        design_documents.html#list-functions>`_.
+
+        :param str ddoc_id: Design document id used to get result.
+        :param str list_name: Name used in part to identify the
+            list function.
+        :param str view_name: Name used in part to identify the view.
+
+        :return: Formatted view result data in text format
+        """
+        ddoc = DesignDocument(self, ddoc_id)
+        headers = {'Content-Type': 'application/json'}
+        resp = get_docs(self.r_session,
+                        '/'.join([ddoc.document_url, '_list', list_name, view_name]),
+                        self.client.encoder,
+                        headers,
+                        **kwargs)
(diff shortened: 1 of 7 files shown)


src/cloudant/database.py
```diff
@@ -762,6 +762,48 @@ def view_cleanup(self):
 
         return resp.json()
 
+    def get_list_function_result(self, ddoc_id, list_name, view_name, **kwargs):
+        """
+        Retrieves a customized MapReduce view result from the specified
+        database based on the list function provided.  List functions are
+        used when you want to access Cloudant directly from a browser, and
```
ricellis: Whilst a browser is the most likely use case, I don't think we should imply it is the only use case, perhaps just add "for example" e.g. 
> ... used<em>, for example,</em> when you ...

src/cloudant/design_document.py
```diff
@@ -149,6 +160,21 @@ def add_search_index(self, index_name, search_func, analyzer=None):
 
         self.indexes.__setitem__(index_name, search)
 
+    def add_list_function(self, list_name, list_func):
+        """
+        Appends a list function to the locally cached DesignDocument
+        indexes dictionary.
+
+        :param list_name: Name used to identify the list function.
+        :param list_func: Javascript list function.
```
ricellis: do we need to say "as a string"?

src/cloudant/design_document.py
```diff
@@ -195,6 +221,21 @@ def update_search_index(self, index_name, search_func, analyzer=None):
 
         self.indexes.__setitem__(index_name, search)
 
+    def update_list_function(self, list_name, list_func):
+        """
+        Modifies/overwrites an existing list function in the
+        locally cached DesignDocument indexes dictionary.
+
+        :param list_name: Name used to identify the list function.
+        :param list_func: Javascript list function.
```
ricellis: do we need to say "as a string"?

src/cloudant/design_document.py
alfinkel: For friendlier rendering in the documentation the param list should look like:

```
param str list_name: ...
param str list_func: ...
```

You should apply that throughout your newly added code since you are leaving out the param "type" in a lot of your docstrings.

emlaver: I think it's better to add `str` (also found in add_search_index):
`:param str list_name: Name used to identify the list function.`
`:param str list_func: Javascript list function.`

src/cloudant/design_document.py
emlaver: See comment https://github.com/cloudant/python-cloudant/pull/189/files/9ffaad0442c0079985bf038d514898acd324f32d#r70287886

src/cloudant/database.py
emlaver: Added "for example" in cf4c283.

src/cloudant/design_document.py
emlaver: Added `str` param type in cf4c283.

src/cloudant/design_document.py
emlaver: Fixed in cf4c283.

ricellis:
+1

rhyshort:
+1

PR #196, opened 2016-07-21
Added support for show functions

### LINKED ISSUE #186: Support show functions in design document module
Supporting show functions is part of a larger work item for design document sub-objects (#137).
- In database module, query _show endpoint: i.e. `get_show_result(show_name, ddoc_show_id, doc_id,  **kwargs)`
- In design document module, support show functions with a `shows` property accessor and management functions

emlaver:
## What

Add support for managing and querying show functions.

## How

- Handled show function execution against the _show endpoint in `CouchDatabase.get_show_function_result`
- Added support for managing show functions in design document

## Testing

Unit test cases to assert show function management in DesignDocumentTests.
Unit test cases to assert `get_show_function_result` in DatabaseTests.

## Issues
fixes #186

### DIFF
--- a/src/cloudant/design_document.py
+++ b/src/cloudant/design_document.py
@@ -44,7 +44,7 @@ class DesignDocument(Document):
         if document_id and not document_id.startswith('_design/'):
             document_id = '_design/{0}'.format(document_id)
         super(DesignDocument, self).__init__(database, document_id)
-        self._nested_object_names = frozenset(['views', 'indexes', 'lists'])
+        self._nested_object_names = frozenset(['views', 'indexes', 'lists', 'shows'])
         for prop in self._nested_object_names:
             self.setdefault(prop, dict())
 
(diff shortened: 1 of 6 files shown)


rhyshort:
+1

src/cloudant/design_document.py
```diff
@@ -522,6 +595,17 @@ def get_list_function(self, list_name):
...
         return self.lists.get(list_name)
 
+    def get_show_function(self, show_name):
+        """
+        Retrieves a specific show function from the locally cached DesignDocument
+        shows dictionary by name.
+
+        :param str show_name: Name used to identify the show function.
+
+        :returns: Index dictionary for the specified show function name
```
ricellis: I'm a bit confused by `Index dictionary` in this returns statement. Doesn't it just return the string of the function?
I note that both `get_index`(self, index_name) and `get_list_function(self, list_name)` use the same terminology, but I think it only makes sense in the index case so probably need to fix this and the `list` case.

ricellis:
+1 besides that one doc thing I noted.

src/cloudant/design_document.py
emlaver: Updated in https://github.com/cloudant/python-cloudant/pull/196/commits/22c15e4a2de6c80755c6e1aa61fb6c44f95b9c83#diff-f2994c8b7af6529726f4b4fbdd48e119R594 and https://github.com/cloudant/python-cloudant/pull/196/commits/22c15e4a2de6c80755c6e1aa61fb6c44f95b9c83#diff-f2994c8b7af6529726f4b4fbdd48e119R594

PR #199, opened 2016-07-27
Added support for update handlers

### LINKED ISSUE #193: Support update handlers
- Update handler custom function
  - Query update handler with GET or POST request
- `updates` accessor property  

This is a subitem part of #137

emlaver:
## What

Support executing update handler functions.

## How

- Added ``updates`` property with documentation and example
- Handled update handler execution against the `_update` 
endpoint in `update_handler_result`

## Testing

Added unit tests for `update_handler_result` in DatabaseTests.

## Issues
fixes #193

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -842,6 +842,57 @@ class CouchDatabase(dict):
                         headers)
         return resp.text
 
+    def update_handler_result(self, ddoc_id, handler_name, doc_id=None, data=None, **params):
+        """
+        Creates or updates a document from the specified database based on the
+        update handler function provided.  Update handlers are used, for
+        example, to provide server-side modification timestamps, and document
+        updates to individual fields without the latest revision. You can
+        provide query parameters needed by the update handler function using
+        the ``params`` argument.
+
+        Create a document with a generated ID:
+
+        .. code-block:: python
+
+            # Assuming that 'update001' update handler exists as part of the
+            # 'ddoc001' design document in the remote database...
+            # Execute 'update001' to create a new document
+            resp = db.update_handler_result('ddoc001', 'update001', data={'name': 'John',
+                                            'message': 'hello'})
+
+        Create or update a document with the specified ID:
+
+        .. code-block:: python
+
+            # Assuming that 'update001' update handler exists as part of the
+            # 'ddoc001' design document in the remote database...
+            # Execute 'update001' to update document 'doc001' in the database
+            resp = db.update_handler_result('ddoc001', 'update001', 'doc001',
+                                            data={'month': 'July'})
+
+        For more details, see the `update handlers documentation
+        <https://docs.cloudant.com/design_documents.html#update-handlers>`_.
+
+        :param str ddoc_id: Design document id used to get result.
+        :param str handler_name: Name used in part to identify the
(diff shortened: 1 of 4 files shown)


src/cloudant/database.py
```diff
@@ -842,6 +842,55 @@ def get_show_function_result(self, ddoc_id, show_name, doc_id):
...
+        the ``params`` argument.
+
+        Create a document with a generated ID:
+
+        .. code-block:: python
+
+            # Assuming that 'update001' update handler exists as part of the
+            # 'ddoc001' design document in the remote database...
+            # Execute 'update001' to create a new document
+            resp = db.update_handler_result('ddoc001', 'update001')
```
rhyshort: This doesn't quite make sense, I don't see the content of the document that you are passing the update handler

src/cloudant/database.py
```diff
@@ -842,6 +842,55 @@ def get_show_function_result(self, ddoc_id, show_name, doc_id):
...
+
+        Create or update a document with the specified ID:
+
+        .. code-block:: python
+
+            # Assuming that 'update001' update handler exists as part of the
+            # 'ddoc001' design document in the remote database...
+            # Execute 'update001' to update document 'doc001' in the database
+            resp = db.update_handler_result('ddoc001', 'update001', 'doc001')
```
rhyshort: Again, this doesn't show the content of the doc being passed to the function.

src/cloudant/database.py
ricellis: That's true it needs a body arg as well. Even though I think it is possible to get the handler to provide all the content as demonstrated by the test, we shouldn't limit it to that special case.

emlaver: The document creation happens within the update handler function.  Maybe it's better to change the example code block to:
```
        # Update handler 'update001' to create new document with generated uuid
        ddoc = DesignDocument(self.db, '_design/ddoc001')
        ddoc['updates'] = {
            'update001': 'function(doc, req) { if (!doc) { '
                         'return [{_id : req.uuid, message : \'created by update001\'}, '
                         '\'Created new doc\']; } } '
        }
       ddoc.save()
       # Execute 'update001' to create a new document
       # If document is created, response object will contain `Created new doc`
       resp = db.update_handler_result('ddoc001', 'update001')
```

rhyshort: I think it would be better if we had an example of calling the update handler as part of creating / updating a document, so we would provide the document body, I don't think there is much value for a user to have an example which is creating a document using just the update handle alone.

ricellis: I think we're looking for an example like:
`db.update_handler_result('ddoc001', 'update001', data={"k1": "v1", "k2": "v2"})`

I think the function might need updating to pop `data` off of the `**=params` stack, or else we need to declare `data` as a separate arg which I think is the general pattern in the library?

emlaver: Updated examples, unit tests, and added `data` argument in 157a2de.

src/cloudant/database.py
emlaver: Updated example in 157a2de.

ricellis:
+1

rhyshort:
+1

PR #204, opened 2016-08-03
Corrected changes example doc

### LINKED ISSUE #194: Fix changes method documentation example
Found this https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L450

looks like we reference `db_updates` rather than `changes` in the example code.  Should be `changes`.

ricellis:
## What

Fixed incorrect code example for `src.cloudant.database.CouchDatabase#changes`

## Testing

Documentation.

## Issues

Fixes #194


emlaver:
+1

PR #206, opened 2016-08-09
205 cloudant tests

### LINKED ISSUE #205: Test failures running against Cloudant service
There are a few test failures on the current master when running against the Cloudant service:

- [x] tests.unit.database_tests.CloudantDatabaseTests.test_get_search_result_executes_search_query
- [x] tests.unit.database_tests.CloudantDatabaseTests.test_get_search_result_executes_search_query_with_group_option
- [x] tests.unit.design_document_tests.DesignDocumentTests.test_geospatial_index
- [x] tests.unit.design_document_tests.DesignDocumentTests.test_get_info
- [x] tests.unit.client_tests.CloudantClientTests.test_billing_data
- [x] tests.unit.database_tests.CloudantDatabaseTests.test_create_json_index
- [x] tests.unit.database_tests.CloudantDatabaseTests.test_create_multiple_indexes_one_ddoc
- [x] tests.unit.database_tests.CloudantDatabaseTests.test_get_query_indexes_raw
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_feed_descending
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_feed_using_since
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_feed_using_timeout
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_feed_with_heartbeat
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_last_seq
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_longpoll_feed
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_normal_feed_default
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_normal_feed_explicit
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_raw_content
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_get_raw_feed_with_heartbeat
- [ ] tests.unit.db_updates_tests.CloudantDbUpdatesTests.test_stop_iteration_of_continuous_feed_using_since_now
- [x] tests.unit.design_document_tests.DesignDocumentTests.test_fetch_map_reduce
- [x] tests.unit.index_tests.IndexTests.test_create_an_index_using_ddoc_index_name
- [x] tests.unit.index_tests.IndexTests.test_create_an_index_using_design_prefix
- [x] tests.unit.index_tests.IndexTests.test_create_an_index_with_empty_ddoc_index_name
- [x] tests.unit.index_tests.IndexTests.test_create_an_index_without_ddoc_index_name
- [x] tests.unit.index_tests.IndexTests.test_index_usage_via_query
- [ ] tests.unit.infinite_feed_tests.InfiniteFeedTests.test_infinite_db_updates_feed
- [x] tests.unit.result_tests.ResultTests.test_get_item_key_slice_start_greater_than_stop

    * 2.next master
- Python version 
    * 2.7.11
- A small code sample that demonstrates the issue.
    * Run tests against Cloudant service

ricellis:
## What

Resolved Cloudant service test issues.

## How

* Fixed allowed types for group_sort

## Testing

* Updated billing tests
    * Used the current month as a data range to prevent any issues with no data with future test account changes.
    * Added new `bill_type` field.
* Used sort for search tests because the results need to be in a guaranteed order to be able to assert they are the same as expected, that means we need a sort or group_sort. 
* Fixed design document tests for newer Cloudant version
    * test_get_info: Removed Cloudant/Couch 2 fields from comparison & removed Cloudant design doc name change that is not needed.
    * test_geospatial_index: Removed `rev` property from row result before comparison.
* Removed `w` parameter from index creation test request options (we don't know how many nodes we will be testing against).
* Fixed index tests to not expect `w` parameter in response.
* Fixed test_get_query_indexes_raw to expect `total_rows` in result.
* Fixed test_get_item_key_slice_start_greater_than_stop to always expect 400 (old behaviour was an empty array on Cloudant, new behaviour is 500, but I think should be 400 see case 71810. This test will still fail until that is resolved.
* Refactored test_index_usage_via_query: Changed to assert that a warning is returned with the result rather than expecting an exception.

## Issues

Fixes #205

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@
   the 2.0.0 behaviour use: ``adapter=Replay429Adapter(retries=10, initialBackoff=0.25)``. If ``retries`` or
   ``initialBackoff`` are not specified they will default to 3 retries and a 0.25 s initial backoff.
 - [FIX] ``415 Client Error: Unsupported Media Type`` when using keys with ``db.all_docs``.
+- [FIX] Allowed strings as well as lists for search ``group_sort`` arguments.
 
 2.0.3 (2016-06-03)
 ==================
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -120,7 +120,7 @@ SEARCH_INDEX_ARGS = {
     'drilldown': list,
     'group_field': STRTYPE,
     'group_limit': (int, NONETYPE),
-    'group_sort': list,
+    'group_sort': (STRTYPE, list),
     'include_docs': bool,
     'limit': (int, NONETYPE),
     'query': (STRTYPE, int),
--- a/tests/unit/client_tests.py
+++ b/tests/unit/client_tests.py
@@ -26,7 +26,7 @@ import json
 import base64
 import sys
 import os
-from datetime import datetime
+import datetime
 
 from cloudant import cloudant, couchdb, couchdb_admin_party
 from cloudant.client import Cloudant, CouchDB
@@ -471,17 +471,19 @@ class CloudantClientTests(UnitTestDbBase):
         """
         try:
             self.client.connect()
+            now = datetime.datetime.now()
             expected = [
                 'data_volume',
                 'total',
                 'start',
                 'end',
                 'http_heavy',
-                'http_light'
+                'http_light',
+                'bill_type'
                 ]
             # Test using year and month
-            year = 2016
-            month = 1
+            year = now.year
+            month = now.month
             data = self.client.bill(year, month)
             self.assertTrue(all(x in expected for x in data.keys()))
             #Test without year and month arguments
@@ -563,6 +565,7 @@ class CloudantClientTests(UnitTestDbBase):
         """
         try:
             self.client.connect()
+            now = datetime.datetime.now()
             expected = [
                 'data_vol',
                 'granularity',
@@ -570,8 +573,8 @@ class CloudantClientTests(UnitTestDbBase):
                 'end'
                 ]
             # Test using year and month
-            year = 2016
-            month = 12
+            year = now.year
+            month = now.month
             data = self.client.volume_usage(year, month)
             self.assertTrue(all(x in expected for x in data.keys()))
             #Test without year and month arguments
@@ -653,6 +656,7 @@ class CloudantClientTests(UnitTestDbBase):
         """
         try:
             self.client.connect()
+            now = datetime.datetime.now()
             expected = [
                 'requests',
                 'granularity',
@@ -660,8 +664,8 @@ class CloudantClientTests(UnitTestDbBase):
                 'end'
                 ]
             # Test using year and month
-            year = 2016
-            month = 1
+            year = now.year
+            month = now.month
             data = self.client.requests_usage(year, month)
             self.assertTrue(all(x in expected for x in data.keys()))
             #Test without year and month arguments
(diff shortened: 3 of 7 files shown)


emlaver:
LGTM +1

src/cloudant/design_document.py
```diff
@@ -498,7 +498,12 @@ def fetch(self):
                         view_def.pop('reduce', None),
                         **view_def
                     )
-
+        # dbcopy is transformed to a new storage location, we need to de-transform
+        if 'options' in self:
+            if 'epi' in self['options']:
+                if 'dbcopy' in self['options']['epi']:
+                    for view in self['options']['epi']['dbcopy'].keys():
+                        self['views'][view]['dbcopy'] = self['options']['epi']['dbcopy'][view]
```
rhyshort: I wonder if its worth making it some kind of option to do the translation, once DBNext is everywhere newer users using DBCopy may get confused with its location.

ricellis: I see your point - I had been thinking similar which is why I had copied it into the old location instead of moving it so it would be accessible in both. It's a bit awkward that the python dict allows direct access to set this (in either location), but we seem to do it via a `kwargs` when creating a view. I'll look into it again and see if there is a better way.

ricellis: @rhyshort so looking at this again my current change means there is no behaviour change from existing.

Adding dbcopy in a view:
`ddoc.add_view(view_name='myview', map_func='', reduce_func='', dbcopy='mycopy')`

This stores the dbcopy parameter in the view (which is the old Cloudant location) and it is accessible via: `ddoc['views']['myview']['dbcopy']`
If the ddoc is saved to older version of Cloudant it stays there. 
If it is saved to a CouchDB 2.0 version the **server** performs a transform that moves the dbcopy into the `{'options' : {'epi' : {'dbcopy' : {'myview' : 'mycopy'}}}` location.

When a design document is retrieved if a `dbcopy` parameter exists in the old location it remains there so is still accessible via `ddoc['views']['myview']['dbcopy']`
If the dbcopy parameter is in the new location the de-transform I added references it from the old location as well so it is accessible from both:
`ddoc['views']['myview']['dbcopy']`
and
`ddoc['options']['epi']['dbcopy']['myview']`

I think there is a slight hole with this approach in that when creating a view using the old style `dbcopy` kwarg it is not accessible in the new location until after a save/fetch with remote. However, it it still possible to add the `dbcopy` directly in the new location through the design document's dictionary. I feel like someone adding `dbcopy` in the old way wouldn't be looking for it in the new location anyway.

Does that make things more clear?
Are you suggesting that we should not make `ddoc['views']['myview']['dbcopy']` accessible unless specifically asked for?

rhyshort: Yes, that is what I'm suggesting, since someone who is new to the client and using DBNext, (which is most on MT now IIRC) will expect there to only be in the new location when creating a new dbcopy view.  So I would expect it to be copied if a config option was set to make it appear as if dpcopy was in the old location.

ricellis: OK moving this to #207 for a discussion that doesn't block the rest of these test fixes.

rhyshort:
+1

PR #208, opened 2016-08-19
Split dbcopy assertions into separate test

### LINKED ISSUE #207: Fix tests for new dbcopy location
Cloudant versions based on CouchDB 2.0 store view dbcopy information in the design document epi options config:
```javascript
{"options": {"epi": {"dbcopy": {"myview": "mycopy"}}}}
```
whereas historically it was stored in the view object:
```javascript
 {"views": {"myview": {"map": "...", "reduce": "...", "dbcopy": "mycopy"}}}
```

This means that it cannot always be retrieved from the same dict location as previously. See additional discussion:

https://github.com/cloudant/python-cloudant/pull/206#discussion_r74218882
https://github.com/cloudant/python-cloudant/pull/206#discussion_r74221218
https://github.com/cloudant/python-cloudant/pull/206#discussion_r75280808
https://github.com/cloudant/python-cloudant/pull/206#discussion_r75288949

    - 2.x
- Python version 
    - Any
- A small code sample that demonstrates the issue.
    - Trying to retrieve the value of `dbcopy` from the old location fails e.g. when running `tests.unit.design_document_tests.DesignDocumentTests.test_fetch_map_reduce` against a new version

ricellis:
## What

Split dbcopy assertions into separate test

## How

Leave the dbcopy information in whatever location it is stored in the design document without trying to change behaviours for different versions.

## Testing

Isolate testing of `dbcopy` from `test_fetch_map_reduce` in new `test_fetch_dbcopy`.
Update assertion to use new Cloudant location for CI testing.
Removed `dbcopy` parameter from tests where it is not used.

## Issues
Fixes #207

### DIFF
--- a/tests/unit/design_document_tests.py
+++ b/tests/unit/design_document_tests.py
@@ -298,9 +298,7 @@ class DesignDocumentTests(UnitTestDbBase):
         ddoc = DesignDocument(self.db, '_design/ddoc001')
         view_map = 'function (doc) {\n  emit(doc._id, 1);\n}'
         view_reduce = '_count'
-        db_copy = '{0}-copy'.format(self.db.database_name)
         ddoc.add_view('view001', view_map, view_reduce)
-        ddoc.add_view('view002', view_map, view_reduce, dbcopy=db_copy)
         ddoc.add_view('view003', view_map)
         ddoc.save()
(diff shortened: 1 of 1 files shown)


emlaver:
LGTM +1

Issue #195, opened 2016-07-20
Improve exception handling when key name is invalid during document update


emlaver:
Replace the 500 HTTP error seen in #71 with a more informative message.

emlaver:
I've submitted a FogBugz ticket for this item.  This error should be a 400 and not a 500 HTTP code.

Issue #65, opened 2015-12-24
Exception handling


alfinkel:
We need to firm up exception handling throughout the library.  
- Exception handling throughout needs to be revisited.
- But specifically, I want to call out the way the library (generally) behaves when a user is used to connect to an account that they do not own.

alfinkel:
To drive home the point also see #71.

emlaver:
Sub issues will be created for improving exception handling:
- When a user is used to connect to an account that they do not own
- When key name is invalid during document update (#195)

rhyshort:
@emlaver Should this now be closed?

emlaver:
@rhyshort We will keep this issue open as we still have these items:
- Continue to add response.raise_for_status()
- Non-reponse related error should be handled using either CloudantException, CloudantArgumentError, or ResultException
- Use design from ResultExeption to improve CloudantException
- Review library and ensure that CloudantException and CloudantArgumentError are being used appropriately

rhyshort:
ok cool.

emlaver:
Moving design improvements for CloudantException to issue #259.

PR #209, opened 2016-08-25
Additional error details appended to HTTPError exception 

emlaver:
## What

Appended error and reason message to HTTPError exception.

## How

- Utilized event hooks in requests to manipulate the response error message.
- New util method `append_response_error_content` that handles appending to the response message.

## Testing

Added unit test in DocumentTests.

## Issues

#65 
#195

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -263,6 +263,20 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
     resp.raise_for_status()
     return resp
 
+#pylint: disable=unused-argument
+def append_response_error_content(response, **kwargs):
+    """
+    Provides a helper to add HTTP response error and reason messages.
+    """
+    try:
+        if response.status_code >= 400 and response.json():
+            reason = response.json().pop('reason', None)
+            error = response.json().pop('error', None)
+            response.reason += ' %s %s' % (error, reason)
+            return response
+    except ValueError:
+        return
+
 # Classes
 
 class _Code(str):
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -25,7 +25,8 @@ from ._2to3 import bytes_, unicode_
 from .database import CloudantDatabase, CouchDatabase
 from .feed import Feed, InfiniteFeed
 from .error import CloudantException, CloudantArgumentError
-from ._common_util import USER_AGENT
+from ._common_util import USER_AGENT, append_response_error_content
+
 
 class CouchDB(dict):
     """
@@ -75,6 +76,9 @@ class CouchDB(dict):
             self.r_session.auth = (self._user, self._auth_token)
             self.session_login(self._user, self._auth_token)
         self._client_session = self.session()
+        # Utilize an event hook to append to the response message
+        # using :func:`~cloudant.common_util.append_response_error_content`
+        self.r_session.hooks['response'].append(append_response_error_content)
 
     def disconnect(self):
         """
--- a/tests/unit/document_tests.py
+++ b/tests/unit/document_tests.py
@@ -196,11 +196,13 @@ class DocumentTests(UnitTestDbBase):
         """
         doc = Document(self.db, 'julia006')
         doc.create()
-        try:
+        with self.assertRaises(requests.HTTPError) as cm:
             doc.create()
-            self.fail('Above statement should raise an Exception')
-        except requests.HTTPError as err:
-            self.assertEqual(err.response.status_code, 409)
+        err = cm.exception
+        self.assertEqual(
+            err.response.status_code,
+            409
+        )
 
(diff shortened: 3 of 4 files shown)


src/cloudant/_common_util.py
```diff
@@ -263,6 +263,16 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
...
     return resp
 
+#pylint: disable=unused-argument
+def append_response_error_content(response, **kwargs):
+    """
+    Provides a helper to add HTTP response error and reason messages.
+    """
+    if response.status_code >= 400:
+        resp_content = response.json()
+        response.url += ' %s %s' % (resp_content['error'], resp_content['reason'])
```
ricellis: I think we should add some safety around the extraction of the JSON and `error` and `reason` keys to prevent this method causing a failure if the response isn't JSON (e.g. what if a proxy sends a 40x with some non-JSON message)? Also in case the JSON doesn't contain the keys, IIRC python will raise `KeyError` if you try to access a key that doesn't exist in a dict - and whilst they should always be there we should protect against them being absent.

src/cloudant/client.py
```diff
@@ -75,6 +76,7 @@ def connect(self):
             self.r_session.auth = (self._user, self._auth_token)
             self.session_login(self._user, self._auth_token)
         self._client_session = self.session()
+        self.r_session.hooks['response'].append(append_response_error_content)
```
ricellis: Please add a comment about what this is doing to help future visitors to the source!

tests/unit/document_tests.py
```diff
@@ -240,6 +242,26 @@ def test_fetch_existing_document_with_docid(self):
...
+        doc = Document(self.db, 'julia006')
+        # Add an invalid key and try to save document
+        doc['_invalid_key'] = 'jules'
+        resp = None
+        with self.assertRaises(requests.HTTPError) as cm:
+            resp = doc.save()
+        err = cm.exception
+        self.assertEqual(
+            str(err),
+            '500 Server Error: Internal Server Error for url: %s doc_validation '
```
ricellis: I guess this will need updating when the server side error is fixed to be a 4xx. 
Perhaps we should assert that the string ends with the error/reason we expect.
Then maybe have a second assertion of the response code 400? (or 500 until fixed).

src/cloudant/_common_util.py
emlaver: Added check to prevent raising errors with `response.json()`.  Also added checks to verify that `error` and `reason` keys exist in 778b332.

src/cloudant/client.py
emlaver: Added comment in 778b332.

tests/unit/document_tests.py
emlaver: I've updated the original assert to check that the string ends with the expected error/reason code.  A second assertion was added to check that the response code is 400 and I'll update to 500 when it's fixed in 778b332.

src/cloudant/_common_util.py
```diff
@@ -263,6 +263,20 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
...
     return resp
 
+#pylint: disable=unused-argument
+def append_response_error_content(response, **kwargs):
+    """
+    Provides a helper to add HTTP response error and reason messages.
+    """
+    try:
+        if (response.status_code >= 400 and response.json() and
+                ('error' in response.json()) and ('reason' in response.json())):
```
ricellis: I'd prefer to see the checks broken out because I believe there are cases where we get only one of the fields. In those cases it would be good to print what we have without bypassing because the other field is not present.

src/cloudant/_common_util.py
```diff
@@ -263,6 +263,20 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
...
+            resp_content = response.json()
+            response.url += ' %s %s' % (resp_content['error'], resp_content['reason'])
```
ricellis: I understand why this is being done via the url property at the moment, but I think it could be risky. I'm concerned about modifying it in case anything else tries to use the `url` (e.g. to form a follow on request to the same URL). I don't know exactly how python-requests handles it's retries and I expect the retry logic gets called before our hook anyway, but if anything calls `response.url` expecting something valid then this type of modification could cause problems.

I think we might be better modifying the `response.reason` instead of the URL.

src/cloudant/_common_util.py
emlaver: Fixed and simplified this block of code in fafcdb8.

src/cloudant/_common_util.py
emlaver: Yea, I agree - switched to `response.reason` and updated test case in fafcdb8.

ricellis:
+1

PR #210, opened 2016-08-29
Add 'q' query parameter for Cloudant Search

emlaver:
## What

Added support for executing search queries with the `q` parameter.

## How

Allow for `q` or `query` but not both parameters in `get_search_result` method
- Added `q` parameter in `get_search_result` documentation

## Testing

Added test cases in CloudantDatabaseTests.

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -1291,8 +1291,9 @@ class CloudantDatabase(CouchDatabase):
         :param int limit: Optional number to limit the maximum count of the
             returned documents. In case of a grouped search, this parameter
             limits the number of documents per group.
-        :param query: A Lucene query in the form of ``name:value``.
+        :param query/q: A Lucene query in the form of ``name:value``.
             If name is omitted, the special value ``default`` is used.
+            The ``query`` parameter can be abbreviated as ``q``.
         :param ranges: Optional JSON facet syntax that reuses the standard
             Lucene syntax to return counts of results which fit into each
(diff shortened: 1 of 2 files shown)


src/cloudant/database.py
```diff
@@ -1291,7 +1291,7 @@ def get_search_result(self, ddoc_id, index_name, **query_params):
         :param int limit: Optional number to limit the maximum count of the
             returned documents. In case of a grouped search, this parameter
             limits the number of documents per group.
-        :param query: A Lucene query in the form of ``name:value``.
+        :param q or query: A Lucene query in the form of ``name:value``.
```
alfinkel: I don't think that this is valid Sphinx syntax so it may cause Sphinx to issue a warning.  If it does then I would separate the `q` param and the `query` param and then add a mention that only one of them is needed.

alfinkel:
What's up with the Jenkins failure?

ricellis:
@alfinkel the Jenkins failure is caused by a bug in the db version deployed on our test cluster (case 71810).

src/cloudant/database.py
emlaver: There are no Sphinx warnings for this syntax.

alfinkel:
+1

ricellis:
+1
I still think this is a bit weird to have two ways of doing this, but since it is part of the HTTP API and we mirror that pretty closely in this library it doesn't really hurt!

PR #220, opened 2016-09-23
Added raise_for_status to improve error handling

### LINKED ISSUE #216: Continue implementing response.raise_for_status() where needed
`response.raise_for_status()` can be found throughout the library.  We should continue using this code and implement it in functions that use the python Requests library.

### LINKED ISSUE #219: Bug in CouchDatabase and Document exists() False condition
This bug exists in versions 2.1.0 down to 2.0.0 alpha for any version of Python.

The problem is in the `exists()` methods for both [CouchDatabase](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L98-L105) and [Document](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/document.py#L95-L105) whereby `True` is returned if the response code is 200 and `False` if it is anything else which is not exactly valid.

In the case of a `False` exists condition, the logic should be changed to specifically check for a 404 response code and return `False` otherwise `raise_for_status()` should be called.  This way we accurately relay whether a db or document actually exists remotely and we also pass back other problems to alert the caller when necessary.

Also, it would probably be better to use the Python requests `head` method rather than `get` for these `exists()` methods since we are really not concerned with any body content in these calls.

emlaver:
## What
- `exists()` methods for both CouchDatabase and Document return `False` if response code is not 200.  Logic was added to call `raise_for_status()` for HTTP errors that are not 200 or 404.
- Added `raise_for_status()` for methods using `requests` that did not include this statement 

## How

- Called `raise_for_status` when status code is not 200 or 404
- Added missing instances of `raise_for_status` in `database.py` and `design_document.py`

## Testing

Added unit tests for asserting a bad request is raised when calling `Document.exists()` and `Database.exists()` methods.

## Issues
#65 
fixes #216 
fixes #219

### DIFF
--- a/tests/unit/database_tests.py
+++ b/tests/unit/database_tests.py
@@ -23,6 +23,7 @@ module docstring.
 """
 
 import unittest
+import mock
 import requests
 import posixpath
 import os
@@ -92,12 +93,27 @@ class DatabaseTests(UnitTestDbBase):
 
     def test_exists(self):
         """
-        Test database exists fucntionality
+        Tests that the result of True is expected when the database exists,
+        and False is expected when the database is nonexistent remotely.
         """
         self.assertTrue(self.db.exists())
         # Construct a database object that does not exist remotely
         fake_db = self.client._DATABASE_CLASS(self.client, 'no-such-db')
         self.assertFalse(fake_db.exists())
+
+    def test_exists_raises_httperror(self):
+        """
+        Test database exists raises an HTTPError.
+        """
+        # Mock HTTPError when running against CouchDB and Cloudant
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.head = mock.Mock(return_value=resp)
+        with self.assertRaises(requests.HTTPError) as cm:
+            self.db.exists()
+        err = cm.exception
+        self.assertEqual(err.response.status_code, 400)
+        self.client.r_session.head.assert_called_with(self.db.database_url)
 
     def test_create_db_delete_db(self):
         """
@@ -748,6 +764,26 @@ class DatabaseTests(UnitTestDbBase):
              '_rev': ddoc_remote['_rev'], '_id': 'julia001',
              'message': 'hello'}
         )
+
+    def test_update_handler_raises_httperror(self):
+        """
+        Test update_handler_result raises an HTTPError.
+        """
+        # Mock HTTPError when running against CouchDB or Cloudant
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.put = mock.Mock(return_value=resp)
+        with self.assertRaises(requests.HTTPError) as cm:
+            self.db.update_handler_result('ddoc001', 'update001', 'julia001',
+                                          field='new_field', value='new_value',
+                                          data={'message': 'hello'})
+        err = cm.exception
+        self.assertEqual(err.response.status_code, 400)
+        ddoc = DesignDocument(self.db, 'ddoc001')
+        self.client.r_session.put.assert_called_with(
+            '/'.join([ddoc.document_url, '_update', 'update001', 'julia001']),
+            data={'message': 'hello'},
+            params={'field': 'new_field', 'value': 'new_value'})
 
 @unittest.skipUnless(
     os.environ.get('RUN_CLOUDANT_TESTS') is not None,
--- a/tests/unit/design_document_tests.py
+++ b/tests/unit/design_document_tests.py
@@ -24,6 +24,8 @@ module docstring.
 import json
 import os
 import unittest
+import mock
+import requests
 
 from cloudant.document import Document 
 from cloudant.design_document import DesignDocument
@@ -754,6 +756,22 @@ class DesignDocumentTests(UnitTestDbBase):
              'name': name
             })
 
+    def test_get_info_raises_httperror(self):
+        """
+        Test get_info raises an HTTPError.
+        """
+        # Mock HTTPError when running against CouchDB and Cloudant
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.get = mock.Mock(return_value=resp)
+        ddoc = DesignDocument(self.db, '_design/ddoc001')
+        with self.assertRaises(requests.HTTPError) as cm:
+            ddoc.info()
+        err = cm.exception
+        self.assertEqual(err.response.status_code, 400)
+        self.client.r_session.get.assert_called_with(
+            '/'.join([ddoc.document_url, '_info']))
+
     @unittest.skipUnless(
         os.environ.get('RUN_CLOUDANT_TESTS') is not None,
         'Skipping Cloudant _search_info endpoint test'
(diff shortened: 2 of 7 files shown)


rhyshort:
One question about the changes, how does this impact the API surface? Would this be regarded as a breaking change at least with the `exists` case because I can't see it documented that it will raise an exception if there is an error status code eg 500.

alfinkel:
@rhyshort - The `raise_for_status()` call in both the database and document `exists()` methods will raise an `requests.exceptions.HTTPError` if response status code is 4XX or 5XX.  So this fix changes the `exists()` methods to now return `True` if 200, `False` if `404` and raise an error if 4XX or 5XX, whereas before it would return `False` if it was anything other than 200 and no error would be raised.  I suppose you _could_ consider this a breaking change but I would rather looking at it as a "fixing" change. ;)

tests/unit/database_tests.py
```diff
@@ -34,6 +34,7 @@
 from cloudant.design_document import DesignDocument
 from cloudant.index import Index, TextIndex, SpecialIndex
 from cloudant.feed import Feed, InfiniteFeed
+from mock import mock
```
alfinkel: I would say that organizationally this import should be up with the other Python libs rather than after the cloudant imports.  So move this import to sit after `import unittest` and before `import requests`

alfinkel:
After reviewing this new logic and checking tests, I found that:

- `test_exists` https://github.com/cloudant/python-cloudant/blob/6114b8ea78718526fb93c5821ae401091dc4391d/tests/unit/database_tests.py#L94-L101 tests for both found and not found conditions.  I don't think it is necessary to split this into separate tests but I do think that some additional content should be added to the docstring specifying that the True and the False conditions are being tested.
- Similarly `test_document_exists` https://github.com/cloudant/python-cloudant/blob/6114b8ea78718526fb93c5821ae401091dc4391d/tests/unit/document_tests.py#L130-L139 also tests for found and not found except that here there are really 3 conditions (2 not found and 1 found).  I think additional docstring content here would also help.

tests/unit/database_tests.py
```diff
@@ -92,13 +93,36 @@ def test_retrieve_creds(self):
...
+            mock_resp.raise_for_status.side_effect = requests.HTTPError(
+                mock.Mock(), 'bad request')
+            mock_resp.raise_for_status.response = mock.Mock(code=400)
+            # mock_session.head = mock.Mock(return_value=mock_resp)
+            self.client.r_session.head = mock.Mock(return_value=mock_resp)
+            with self.assertRaises(requests.HTTPError):
+                self.db.exists()
+            self.assertTrue(self.client.r_session.head.called)
+            self.assertEqual(1, mock_resp.raise_for_status.call_count)
+
```
alfinkel: My suggestion here would be to have one test that functions for both CouchDB and Cloudant that mocks the response code rather than introduce this `if` logic in a unit test.  Also I would not call this a "bad request" test because it implies that this is only testing a 400 return code which if you change this to a mocked test for both CouchDB and Cloudant as I suggested, this test would cover all raised HTTPErrors.  So... `test_exists_raises_httperror`.

tests/unit/document_tests.py
```diff
@@ -32,6 +32,7 @@
 
 from cloudant.document import Document
 from cloudant.error import CloudantException
+from mock import mock
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/220/files#r80681506

tests/unit/document_tests.py
```diff
@@ -137,6 +138,33 @@ def test_document_exists(self):
...
+            mock_resp.raise_for_status.response = mock.Mock(code=400)
+            # mock_session.head = mock.Mock(return_value=mock_resp)
+            self.client.r_session.head = mock.Mock(return_value=mock_resp)
+            with self.assertRaises(requests.HTTPError):
+                doc = Document(self.db)
+                doc['_id'] = 'julia006'
+                doc.exists()
+            self.assertTrue(self.client.r_session.head.called)
+            self.assertEqual(1, mock_resp.raise_for_status.call_count)
+
```
alfinkel: Same as https://github.com/cloudant/python-cloudant/pull/220/files#r80687498

alfinkel:
You've added a couple of `raise_for_status` calls in the design_document module but I don't see any tests that test that logic.

tests/unit/database_tests.py
emlaver: Fixed in 0503257.

tests/unit/database_tests.py
emlaver: Fixed in 0503257.

tests/unit/document_tests.py
emlaver: Fixed in 0503257.

tests/unit/document_tests.py
emlaver: Fixed in 0503257.

emlaver:
Added missing tests for `raise_for_status` calls in design_document module in 0503257.

tests/unit/database_tests.py
```diff
@@ -23,6 +23,7 @@
 """
 
 import unittest
+from mock import mock
```
alfinkel: I think your code will still work if you just use `import mock` here instead of `from mock import mock`.

tests/unit/design_document_tests.py
```diff
@@ -24,6 +24,8 @@
 import json
 import os
 import unittest
+from mock import mock
```
alfinkel: I think your code will still work if you just use `import mock` here instead of `from mock import mock`.

tests/unit/design_document_tests.py
```diff
@@ -754,6 +756,22 @@ def test_get_info(self):
...
+    def test_get_info_raises_httperror(self):
+        """
+        Test get_info raises an HTTPError.
+        """
+        # Mock HTTPError when running against CouchDB and Cloudant
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.get = mock.Mock(return_value=resp)
+        with self.assertRaises(requests.HTTPError) as cm:
+            ddoc = DesignDocument(self.db, '_design/ddoc001')
```
alfinkel: I would move this line before the `with self.assertRaises...` that way it ensures that the HTTPError is being raised on the call to `.info()`.

tests/unit/design_document_tests.py
```diff
@@ -781,6 +799,27 @@ def test_get_search_info(self):
...
+        """
+        Test get_search_info raises an HTTPError.
+        """
+        # Mock HTTPError when running against Cloudant
+        search_index = 'search001'
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.get = mock.Mock(return_value=resp)
+        with self.assertRaises(requests.HTTPError) as cm:
+            ddoc = DesignDocument(self.db, '_design/ddoc001')
```
alfinkel: I would move this line before the `with self.assertRaises...` that way it ensures that the HTTPError is being raised on the call to `.search_info(...)`.

tests/unit/document_tests.py
```diff
@@ -23,6 +23,7 @@
 """
 
 import unittest
+from mock import mock
```
alfinkel: I think your code will still work if you just use `import mock` here instead of `from mock import mock`.

tests/unit/document_tests.py
```diff
@@ -137,6 +139,22 @@ def test_document_exists(self):
...
+        """
+        Test document exists raises an HTTPError.
+        """
+        # Mock HTTPError when running against CouchDB and Cloudant
+        resp = requests.Response()
+        resp.status_code = 400
+        self.client.r_session.head = mock.Mock(return_value=resp)
+        with self.assertRaises(requests.HTTPError) as cm:
+            doc = Document(self.db)
+            doc['_id'] = 'julia006'
```
alfinkel: Again, I would move these 2 lines above the `with...` for the same reason mentioned above.

alfinkel:
+1

rhyshort:
ok, I've done some more reading of our docs for the client, and since we don't call out exceptions explicitly, if we add a note to the change log that points out that we know raise an error correctly when 4xx or 5xx status codes are encountered, I think we are good. @emlaver can you add the change log entry?

emlaver:
@rhyshort Added changes entry in b032a01.

PR #223, opened 2016-10-12
Added Jenkinsfile

ricellis:
## What

Added a `Jenkinsfile` for CI that runs tests on 2 python versions.

## How

Added a `Jenkinsfile`.
`Jenkinsfile` defines a test routine to be used by any python version.
The test routine is called for python versions `2.7.12` and `3.5.2` which run in parallel.

## Testing

Build status for this branch has Jenkins result produced by the new `Jenkinsfile`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,17 +7,17 @@ def test_python(pythonVersion)
     // Set up the environment and test
     withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'clientlibs-test', usernameVariable: 'DB_USER', passwordVariable: 'DB_PASSWORD']]) {
       try {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,43 @@
...
+. ./tmp/bin/activate
+echo \$DB_USER
+export ADMIN_PARTY=true
+export RUN_CLOUDANT_TESTS=1
+export CLOUDANT_ACCOUNT=\$DB_USER
+# Temporarily disable the _db_updates tests pending resolution of case 71610
+export SKIP_DB_UPDATES=1
+pip install -r requirements.txt
+pip install -r test-requirements.txt
+nosetests -w ./tests/unit --with-xunit"""
```
rhyshort: Just a tiny nit pick is it possible to get this to line up with `virutalenv` on line 10?

ricellis: Added in a1eedd2
let's see if it still works :)

Issue #225, opened 2016-10-14
"q" is not a valid option for _search query


alfinkel:
In python-cloudant 2.1.1 for both Python 2 and 3 "q" needs to be a valid parameter for calls to the database `get_search_result` method.  Initial work was done in PR #210 but apparently this bug slipped through and "q" is still not actually acknowledged as a valid parameter.  The problem is that when the parameter validation performed [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L1343) checks against the `SEARCH_INDEX_ARGS` defined [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/_common_util.py#L117-L136), "q" is not included in that dictionary of valid arguments.  The bug slipped through because the test [here](https://github.com/cloudant/python-cloudant/blob/master/tests/unit/database_tests.py#L1377) is incorrectly using "query" where it should be using "q".  Changing to "q" would reveal the bug.

So the course of action here is to add "q" to the SEARCH_INDEX_ARGS and fix the test to use "q" instead of "query".

PR #226, opened 2016-10-14
Fix q parameter search query bug

alfinkel:
## What

Added `q` entry into `SEARCH_INDEX_ARGS` so that `q` would be recognized as a valid search parameter by the database `get_search_results` method.

## How

- Added `'q': (STRTYPE, int)` to _common_util.py `SEARCH_INDEX_ARGS` dictionary.

## Testing

- Fixed the `test_get_search_result_executes_search_q` database test to use `q` instead of `query`.

## Issues

#225


emlaver:
Looks good to me +1

PR #229, opened 2016-10-14
Fix request session so all requests fail after client is disconnected

### LINKED ISSUE #218: Database and Document object request session still active after client disconnect
This bug appears in v 2.1.0 down to 2.0.0 alpha for any version of Python.

After disconnecting from a client any objects created based on that client like Database or Document objects are still able to make requests to Cloudant/CouchDB.  

This can be resolved by removing this [line](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/client.py#L76) in the client code.  This "fix" will not work for CouchDB Admin Party mode and unfortunately I don't see a way to easily remedy that.  However, in Admin Party mode this behavior is less of a bug and more of an odd side effect IMO.

Example:

```python
from cloudant import CouchDB, Cloudant
c = Cloudant('alfinkel3', '...', account='alfinkel3')
c.connect()
db = c['juliadb']
print 'This should display the document count: ', db.doc_count()
c.disconnect()

print 'This should raise a 401 and NOT display the document count: ', db.doc_count()
print 'This is what is allowing the db requests session to still make requests: ', db.r_session.auth
```
 Outputs:

```
This should display the document count:  207
This should raise a 401 and NOT display the document count:  207
This is what is allowing the db requests session to still make requests:  ('alfinkel3', '...')
```

emlaver:
## What

After disconnecting from a client any objects created based on that client like Database or Document objects are still able to make requests to Cloudant/CouchDB.

## How

- Removed auth attribute from r_session so objects created based on the client are not able to make requests
- Added `r_session` property in `database.py` and `document.py`

## Testing

Added test to assert a 401 status code when trying to access a databases's document count after disconnecting from the client.

## Issues
fixes #218

### DIFF
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -80,9 +80,7 @@ class CouchDB(dict):
             self.r_session.mount(self.server_url, self.adapter)
         if self._client_user_header is not None:
             self.r_session.headers.update(self._client_user_header)
-        if not self.admin_party:
-            self.r_session.auth = (self._user, self._auth_token)
-            self.session_login(self._user, self._auth_token)
+        self.session_login(self._user, self._auth_token)
         self._client_session = self.session()
         # Utilize an event hook to append to the response message
         # using :func:`~cloudant.common_util.append_response_error_content`
--- a/tests/unit/document_tests.py
+++ b/tests/unit/document_tests.py
@@ -722,5 +722,22 @@ class DocumentTests(UnitTestDbBase):
         finally:
             attachment.close()
 
+    def test_document_request_fails_after_client_disconnects(self):
+        """
+        Test that after disconnecting from a client any objects created based
+        on that client are not able to make requests.
+        """
+        self.client.connect()
+        doc = Document(self.db, 'julia001')
+        doc.save()
+        self.client.disconnect()
+
+        try:
+            with self.assertRaises(AttributeError):
+                doc.fetch()
+            self.assertIsNone(doc.r_session)
+        finally:
+            self.client.connect()
+
 if __name__ == '__main__':
     unittest.main()
--- a/tests/unit/replicator_tests.py
+++ b/tests/unit/replicator_tests.py
@@ -26,7 +26,6 @@ import unittest
 import uuid
 import time
 import requests
-import os
 
 from cloudant.replicator import Replicator
 from cloudant.document import Document
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
-2.2.0 (Unreleased)
+2.1.2 (Unreleased)
 ==================
+- [FIXED] Requests session is no longer valid after disconnect.
 
 2.1.1 (2016-10-03)
 ==================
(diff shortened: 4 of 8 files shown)


src/cloudant/client.py
```diff
@@ -73,7 +73,6 @@ def connect(self):
         if self._client_user_header is not None:
             self.r_session.headers.update(self._client_user_header)
         if not self.admin_party:
-            self.r_session.auth = (self._user, self._auth_token)
             self.session_login(self._user, self._auth_token)
```
alfinkel: I would remove this `if` block altogether and replace it with just `self.session_login(self._user, self._auth_token)`.  There is a check in the session_login method which will handle the admin_party condition so there is no need to double up the "if" logic now.

tests/unit/replicator_tests.py
```diff
@@ -59,13 +59,19 @@ def tearDown(self):
         """
         Reset test attributes
         """
-        self.target_db.delete()
+        if self.client.admin_party:
+            self.target_db.delete()
+        else:
+            self.client.r_session.delete(self.target_db.database_url)
```
alfinkel: I've got to be honest here.  I don't have a clue why you did this.  I assume this is what you wanted to chat about?

tests/unit/replicator_tests.py
```diff
@@ -59,13 +59,19 @@ def tearDown(self):
...
-        del self.target_db
         while self.replication_ids:
             self.replicator.stop_replication(self.replication_ids.pop())
         del self.replicator
-        self.db_tear_down()
+        if self.client.admin_party:
+            self.db_tear_down()
+        else:
+            self.client.r_session.delete(self.target_db.database_url)
+        del self.target_db
```
alfinkel: Ditto: https://github.com/cloudant/python-cloudant/pull/229/files#r83699241

tests/unit/document_tests.py
```diff
@@ -722,5 +722,22 @@ def test_attachment_management(self):
...
+        Test that after disconnecting from a client any objects created based
+        on that client are not able to make requests.
+        """
+        self.client.connect()
+        doc = Document(self.db, 'julia001')
+        doc.save()
+        self.client.disconnect()
+
+        try:
+            with self.assertRaises(AttributeError) as cm:
```
alfinkel: remove `as cm`

tests/unit/replicator_tests.py
```diff
@@ -86,8 +86,9 @@ def test_constructor_failure(self):
         """
         repl = None
         try:
-            self.client.disconnect()
-            repl = Replicator(self.client)
+            client = copy.deepcopy(self.client)
+            client.disconnect()
+            repl = Replicator(client)
```
alfinkel: is this `client` copy necessary anymore?  If so, I'd prefer it were just a clean contruction of a new client rather than a deepcopy of the `self.client`.

alfinkel:
I would also add the new `r_session` property work that you did to the PR description "How" section.

tests/unit/replicator_tests.py
```diff
@@ -22,11 +22,11 @@
 
 """
 
+import copy
```
alfinkel: remove this

alfinkel:
+1

PR #232, opened 2016-10-20
Honour the custom encoder parameter for json encoding.

### LINKED ISSUE #221: json.dumps not consistently setting cls kwarg to supplied encoder
All versions of python-cloudant and Python are affected.

We specifically allow for setting an `encoder` via key word argument when constructing a `client` instance.  The assumption is that this encoder would be set for all JSON encoding handled by the library, in other words all instances of `json.dumps` in the code should have a `cls` setting pointing to the encoder set on the client.  However, the use of the `encoder` is being used in some instances and not others.

We should fix this and update all occurrences of `json.dumps` to have the `cls` keyword argument set to the client encoder.

rhyshort:
## What

Honour the JSON encoder passed to the client at construction when transforming data into JSON.

## How

Specify the `cls` parameter for `json.dumps` calls which omitted it.

## Testing

Tests added which use a string for encoder to force a `TypeError` to be thrown.
## Issues

Fixes #221

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,7 @@
 2.3.0 (Unreleased)
 ==================
+- [FIXED] Issue where the custom JSON encoder was not used when transforming
+   data.
 
 2.2.0 (2016-10-20)
 ==================
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -727,7 +727,7 @@ class Cloudant(CouchDB):
         )
         resp = self.r_session.put(
             endpoint,
-            data=json.dumps(config),
+            data=json.dumps(config, cls=self.encoder),
             headers={'Content-Type': 'application/json'}
         )
         resp.raise_for_status()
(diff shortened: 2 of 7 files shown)


emlaver:
@rhyshort For testing, you could write a couple lines for a custom encoder and pass that in or use something like `self.client = CouchDB(…, encoder='utf-8')`.  Check out this link: http://www.yilmazhuseyin.com/blog/dev/advanced_json_manipulation_with_python/

Another thing is you'll need to pass the encoder to the `set_up_client` to be able to have `self.client` in your test case use the custom encoder.  This auto test case is a good example of passing in the parameter to `client`:
https://github.com/cloudant/python-cloudant/blob/cec434a76b22efcb6425c39be003c6cf5304f7f0/tests/unit/client_tests.py#L98

rhyshort:
Tests added, and description updated.

alfinkel:
+1

emlaver:
Looks good +1

Issue #227, opened 2016-10-14
ADMIN_PARTY environment setting is not needed in JenkinsFile


alfinkel:
I was poking around in the repo and stumbled upon the recently added JenkinsFile.  It looks like the `ADMIN_PARTY` environment variable is being set to `true` [here](https://github.com/cloudant/python-cloudant/blob/master/Jenkinsfile#L13) which is unnecessary for Cloudant tests and only applies to running tests against CouchDB.  This does not cause any problem right now given the way the tests are written but if things change then this could become an issue in the future.  I would suggest to remove this line of code from the JenkinsFile since it currently does nothing and may possibly do some harm later on.

PR #236, opened 2016-10-27
Add security document management support

alfinkel:
## What

Added the ability to retrieve a security document from a specific database and to update that document as necessary.

## How

- Add a SecurityDocument class that can: 
  - fetch the security document content remotely
  - save updated content to the server
  - a context manager is also provided
- Add the database convenience method get_security_document

## Testing

- Add security document tests
- Add get_security_document database test
- Fix Jenkinsfile to remove unnecessary ADMIN_PARTY export which fixes the security document tests

## Issues

#52 
#227

### DIFF
--- a/tests/unit/unit_t_db_base.py
+++ b/tests/unit/unit_t_db_base.py
@@ -61,6 +61,7 @@ import unittest
 import requests
 import os
 import uuid
+import json
 
 from cloudant.client import CouchDB, Cloudant
 from cloudant.design_document import DesignDocument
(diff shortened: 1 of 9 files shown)


tests/unit/unit_t_db_base.py
```diff
@@ -256,3 +257,44 @@ def create_search_index(self):
...
+                    'foo2': ['_reader']
+                }
+            }
+            self.mod_sdoc = {
+                'cloudant': {
+                    'bar1': ['_reader', '_writer'],
+                    'bar2': ['_reader']
+                }
+            }
+        if os.environ.get('ADMIN_PARTY') == 'true':
```
emlaver: This will fail if `ADMIN_PARTY=true` and `RUN_CLOUDANT_TESTS=1` which may be why the jenkins test is failing.

alfinkel: Yep that certainly is why.  I just fixed the Jenkinsfile to remove the ADMIN_PARTY export since it is not needed.  This fixes #227 as well.  Done in https://github.com/cloudant/python-cloudant/pull/236/commits/ad78e222587c2cb84dc78e727190fd359484b7dd

Issue #235, opened 2016-10-26
Automatically handle session renewal


rhyshort:
Currently a user needs to handle the starting of a new session when theirs expires by either creating a new client, or calling `client.session_login`. What we should do is handle this situation transparently to user and renew the session if we receive a response indicating the session has expired. And retry the user request and return the results.

alfinkel:
@ricellis, @rhyshort, @emlaver, 

What are your thoughts on tracking the session start time as a client attribute and basing the refresh of the cookie auth on the difference between the time of a request and the session start time?  

Prior to honoring the request we would perform this check and if the difference is greater than a certain amount (lets say 23 hours) then we would perform the POST to the _session endpoint to refresh the cookie auth before proceeding to handle the original request.  This could all be managed via a kwarg passed in when constructing the client so the default behavior can remain the same as it is now.  This would be a really simple solution since all of the classes that require an `r_session` currently already return it via a function with the `@property` decorator.  So the necessary code to do this comparison already has a convenient home.  I think the overhead would be relatively small to perform the time comparison prior to each request and a user of the library would have to expressly ask for this behavior via the additional keyword argument so they could/would be aware that they are asking for additional "reconnect" functionality that has some overhead.

Let me know your thoughts.

smithsz:
Hi, I just pushed up some work I did last week for this. We ran into the bug in rebal2.
I was contemplating running the method call through a wrapper. If any request returns a `401` response then we check the cookie is still valid. If it is valid then just return the response as is. If not, issue a `login()` and retry.

The couchdb logs for a db create with an expired cookie would look something like this:
```
11:09:28.207 [notice] a1aa5eaded 127.0.0.1 localhost:5984 PUT /newdb1 401 ok 1
11:09:28.219 [notice] b21dc6126f 127.0.0.1 localhost:5984 GET /_session 200 ok 0
11:09:28.222 [notice] 431c61ed95 127.0.0.1 localhost:5984 POST /_session 200 ok 1
11:09:28.227 [notice] 2957c1a12a 127.0.0.1 localhost:5984 PUT /newdb1 201 ok 67
```

rhyshort:
@alfinkel I wouldn't want to rely on the user getting the cookie timeout right, plus if a request is made and the cookie is near expiration, Cloudant should send a `Set-Cookie` header to update what we have.  So I think its easier and less error prone, to check the response like the other libs and login and retry, we can turn this on and off via kwargs if we should keep the current behaviour.

[...]

PR #239, opened 2016-11-01
Add session auto-renewal option

alfinkel:
## What

Add a session automatic renewal option to the `Session` object used by this library so that in the event of a request being made using expired session authentication credentials, the auth is renewed and the request is re-tried.

## How

- Sub-class `requests.Session` as `InfiniteSession` class and handle a 401 response as needed
- Add the auto_renew keyword argument to client constructor to allow for auto renewal of session auth

## Testing

- Current set of unit tests all pass
- Manual tests to validate functionality were successful against Cloudant
- Add client unit tests to test client auto renewal functionality
- Add additional auth renewal tests

## Issues

#235

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -21,8 +21,9 @@ import sys
 import platform
 from collections import Sequence
 import json
+from requests import Session
 
-from ._2to3 import STRTYPE, NONETYPE, UNITYPE, iteritems_
+from ._2to3 import STRTYPE, NONETYPE, UNITYPE, iteritems_, url_parse
 from .error import CloudantArgumentError
 
 # Library Constants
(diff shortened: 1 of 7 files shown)


src/cloudant/_common_util.py
```diff
@@ -288,3 +289,35 @@ class _Code(str):
...
+    def request(self, method, url, **kwargs):
+        """
+        Overrides ``requests.Session.request`` to perform a POST to the
+        _session endpoint to renew Session cookie authentication settings and
+        then retry the original request, if necessary.
+        """
+        resp = super(InfiniteSession, self).request(method, url, **kwargs)
+        path = url_parse(url).path.lower()
+        post_to_session = method.upper() == 'POST' and path == '/_session'
+        if resp.status_code == 401 and not post_to_session:
```
smithsz: I think we need to capture `403` responses here too. DBcore clusters may still return the 403 `{"error":"credentials_expired",...}` response.

I guess something along the lines of,
```python
is_expired = any((
    resp.status_code == 403 and resp.json().get('error') == 'credentials_expired',
    resp.status_code == 401
))
if not post_to_session and is_expired:
    ...
```

src/cloudant/client.py
```diff
@@ -63,6 +70,7 @@ def __init__(self, user, auth_token, admin_party=False, **kwargs):
         self.encoder = kwargs.get('encoder') or json.JSONEncoder
         self.adapter = kwargs.get('adapter')
         self.r_session = None
+        self._auto_renew = kwargs.get('auto_renew', False)
```
smithsz: Why default `False`?

alfinkel: Because that preserves current behavior of the library.

src/cloudant/_common_util.py
alfinkel: Sure, makes sense.  Done in https://github.com/cloudant/python-cloudant/pull/239/commits/e88a5eb701421fcde605895f6a43f50710cbf765.

Issue #245, opened 2016-11-14
Stack overflow iterating over InfiniteFeed


appeltel:
Noticed in a log file that iterating over an `InfiniteFeed` object when calling the database method `infinite_changes(...)` had caused a stack overflow, small snippet from the stack trace:

```
…snip...
  File "/opt/sapi/manifold/venv/lib/python2.7/site-packages/cloudant/feed.py", line 259, in next
    return super(InfiniteFeed, self).next()
  File "/opt/sapi/manifold/venv/lib/python2.7/site-packages/cloudant/feed.py", line 164, in next
    return self.next()
  File "/opt/sapi/manifold/venv/lib/python2.7/site-packages/cloudant/feed.py", line 259, in next
    return super(InfiniteFeed, self).next()
  File "/opt/sapi/manifold/venv/lib/python2.7/site-packages/cloudant/feed.py", line 164, in next
…snip...
```
Appears that due to the `next` function being recursive on a skipped line (i.e. heartbeat, etc...) having several hundred skips in a row will blow the stack. This also can create a messy stack trace if there are a number of skips and an error.

Replacing the recursive call with a loop should resolve the problem, I'll follow this up with a patch.

PR #246, opened 2016-11-14
Iterate over skipped changes lines

appeltel:
## What

Changed Feed.next() and InfiniteFeed.next() to iterate over skipped lines rather than recursing over them.

## How

Use a while loop in Feed.next() and InfiniteFeed.next(). This unfortunately results in a few lines of duplicated code as there was no clear way to make use of calling the superclass method.

## Testing

Functionality does not change, other than to tidy up stack traces and prevent overflow, so no tests are altered.

## Issues

This fixes issue #245 where recursion can cause a stack overflow.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ==================
 - [FIXED] Resolved issue where generated UUIDs for replication documents would
   not be converted to strings.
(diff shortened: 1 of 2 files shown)


alfinkel:
+1

PR #253, opened 2016-12-06
251 fix tests

### LINKED ISSUE #251: Miscellaneous test failures
- There are a few test failures that need resolving:
    - `tests.unit.document_tests.DocumentTests.test_appended_error_message_using_save_with_invalid_key`: fails because the test is checking for a `500` error, which was a server bug that has been resolved. It should be checking for `400`.
    - `tests.unit.database_tests.CloudantDatabaseTests.test_get_search_result_without_query`: fails because the test asserts on the entire returned error message as a string which includes some JSON which can be in different orders.
    - `tests.unit.design_document_tests.DesignDocumentTests.test_get_search_info`: fails because the search index hasn't always finished building so the test assertions for sequence numbers fail.

    - Master
- Python version
    - Any

ricellis:
## What

Fixed tests that failed because of ordering issues or behaviour differences between CouchDB 1.6 and newer versions.

## Testing

* Fixed `test_appended_error_message_using_save_with_invalid_key` to check for `400 Bad Request`, but allowed to continue passing with 500 for CouchDB 1.6.x.
* Make test_get_search_result_without_query assertions order independent 
* Made test_get_search_info checks not rely on indexing completion

## Issues

Fixes #251

### DIFF
--- a/tests/unit/document_tests.py
+++ b/tests/unit/document_tests.py
@@ -282,10 +282,6 @@ class DocumentTests(UnitTestDbBase):
                 str(err.response.reason),
                 'Internal Server Error doc_validation Bad special document member: _invalid_key'
             )
-            self.assertEqual(
-                err.response.status_code,
-                500
-            )
         else:
             self.assertEqual(
                 str(err.response.reason),
--- a/tests/unit/document_tests.py
+++ b/tests/unit/document_tests.py
@@ -273,14 +273,28 @@ class DocumentTests(UnitTestDbBase):
         with self.assertRaises(requests.HTTPError) as cm:
             doc.save()
         err = cm.exception
-        self.assertEqual(
-            str(err.response.reason),
-            'Internal Server Error doc_validation Bad special document member: _invalid_key'
-        )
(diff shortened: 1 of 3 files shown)


tests/unit/document_tests.py
```diff
@@ -273,14 +273,28 @@ def test_appended_error_message_using_save_with_invalid_key(self):
...
+        # Should be a 400 error code, but CouchDB 1.6 issues a 500
+        if err.response.status_code == 500:
+            #Check this is CouchDB 1.6
+            self.assertTrue(self.client.r_session.head(self.url).headers['Server'].find('CouchDB/1.6.') >= 0,
+                            '500 returned but was not CouchDB 1.6.x')
+            self.assertEqual(
+                str(err.response.reason),
+                'Internal Server Error doc_validation Bad special document member: _invalid_key'
+            )
+            self.assertEqual(
```
brynh: This assert is redundant.

ricellis: Good point. Removed in f5ac09b.

PR #255, opened 2016-12-07
Implement method to create Cloudant client from VCAP services env

### LINKED ISSUE #254: Implement method to create Cloudant client from VCAP services environment variable
Add an additional method to parse a `VCAP_SERVICES` environment variable allowing easy binding of a service instance to a Cloud Foundry application.

Example `VCAP_SERVICES` variable is JSON:
```json
{
    "cloudantNoSQLDB": [
        {
            "credentials": {
                "username": "example",
                "password": "xxxxxxx",
                "host": "example.cloudant.com",
                "port": 443,
                "url": "https://example:xxxxxxx@example.cloudant.com"
            },
            "syslog_drain_url": null,
            "label": "cloudantNoSQLDB",
            "provider": null,
            "plan": "Lite",
            "name": "Cloudant NoSQL DB"
        }
    ]
}
```

smithsz:
## What
Implement `cloudant_bluemix` method that creates a Cloudant client from a VCAP_SERVICES environment variable. This allows easy binding of a service instance to a Cloud Foundry application.

## How
Parse the VCAP_SERVICES environment variable as JSON for necessary configuration required to create a Cloudant client. 

## Testing

## Issues
Fixes #254.


src/cloudant/__init__.py
```diff
@@ -62,6 +63,56 @@ def cloudant(user, passwd, **kwargs):
     cloudant_session.disconnect()
 
 @contextlib.contextmanager
+def cloudant_bluemix(service_name, **kwargs):
```
ricellis: Thoughts around using `service_name=None` and defaulting to use the only service found or erroring if there are more than 1 for a default case?
It would avoid the need to transcribe service names into app code for probably the most common use case.

smithsz: Agreed, this would be better. Addressed in squash commit 05acaee0b270a503b264ed31f70908e805e65017.

src/cloudant/__init__.py
```diff
@@ -62,6 +63,57 @@ def cloudant(user, passwd, **kwargs):
     cloudant_session.disconnect()
 
 @contextlib.contextmanager
+def cloudant_bluemix(service_name=None, **kwargs):
+    """
+    Provides a context manager to create a Cloudant session and provide access
+    to databases, docs etc.
+
+    :param str service_name: Optional Cloud Foundry service instance name. Only
```
rhyshort: I wonder if we should call out bluemix specifically rather than Cloud Foundry

smithsz: Yep, will rename.

requirements.txt
```diff
@@ -1,2 +1,2 @@
 requests >=2.7.0, <3.0.0
-
+mock==2.0.0
```
alfinkel: We split out deplendencies between requirements.txt and test_requirements.txt.  The .travis.yml executes both (https://github.com/cloudant/python-cloudant/blob/master/.travis.yml#L14).  We already have a dependency for `mock` in https://github.com/cloudant/python-cloudant/blob/master/test-requirements.txt#L1.  I would remove this dependency.  I think even though the one in test_requirements.txt is 1.3.0, it should serve your needs.

src/cloudant/__init__.py
```diff
@@ -62,6 +63,57 @@ def cloudant(user, passwd, **kwargs):
...
+                            "url": "https://example:xxxxxxx@example.cloudant.com"
+                        },
+                        "syslog_drain_url": null,
+                        "label": "cloudantNoSQLDB",
+                        "provider": null,
+                        "plan": "Lite",
+                        "name": "Cloudant NoSQL DB"
+                    }
+                ]
+        }
```
alfinkel: This docstring code block appears to have a missing close parenthesis.  If you correct the structure of the block you can set it to `.. code-block:: json` once it is valid.  I think then it would render nicely in Sphinx.

requirements.txt
smithsz: Will do. Thanks.

src/cloudant/__init__.py
smithsz: Good spot. I'll amend.

tests/unit/cloud_foundry_tests.py
```diff
@@ -0,0 +1,172 @@
...
+    def test_raise_error_for_invalid_credentials_type(self, m_getenv):
+        m_getenv.return_value = self._test_vcap_services_multiple
+        with self.assertRaises(CloudantException):
+            CloudFoundryService('Cloudant NoSQL DB 6')
+
+    @mock.patch('os.getenv')
+    def test_raise_error_for_missing_service(self, m_getenv):
+        m_getenv.return_value = self._test_vcap_services_multiple
+        with self.assertRaises(CloudantException):
+            CloudFoundryService('Cloudant NoSQL DB 7')
```
smithsz: Agreed, I'll add additional assertions to check the exception messages.

PR #260, opened 2016-12-13
Improved logic for appending http error and message

emlaver:
## What
As part of reviewing the error handling of the library (#65), this method wasn't handling the ValueError from `response.json()` in the most efficient manner. 
It's also confusing to see the `**kwargs` that appears to be unused but is required as part of the event hook. 

## How
- Improved logic for appending additional error message
- Updated docstring to include why `response` and `kargs` are necessary

## Testing
No new tests and all existing tests are passing.

## Issues
#65

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -268,16 +268,24 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
 #pylint: disable=unused-argument
 def append_response_error_content(response, **kwargs):
     """
-    Provides a helper to add HTTP response error and reason messages.
+    Provides a helper to act as callback function for the response event hook
+    and add a HTTP response error with reason message to ``response.reason``.
+    The ``response`` and ``**kwargs`` are necessary for this function to
+    properly operate as the callback.
+
+    :param response: HTTP response object
+    :param kwargs: HTTP request parameters
     """
-    try:
-        if response.status_code >= 400 and response.json():
-            reason = response.json().pop('reason', None)
-            error = response.json().pop('error', None)
-            response.reason += ' %s %s' % (error, reason)
-            return response
-    except ValueError:
-        return
+    if response.status_code >= 400:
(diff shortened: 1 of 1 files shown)


src/cloudant/_common_util.py
```diff
@@ -268,16 +268,24 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
 #pylint: disable=unused-argument
 def append_response_error_content(response, **kwargs):
     """
-    Provides a helper to add HTTP response error and reason messages.
+    Provides a helper to act as callback function for the response event hook
+    and add a HTTP response error with reason message to `response.reason`.
+    The `response` and `**kwargs` are necessary for this function to
```
alfinkel: I think you meant to use double ticks rather than single here (response.reason, response, **kwargs).  In Sphinx using single ticks will _italisize_ while double ticks will `highlight`.  If you meant to _italisize_ then ignore this comment.

alfinkel: I guess in the long run it really doesn't matter because I think these functions aren't included with our docs.  So this is really a non-blocking comment.

src/cloudant/_common_util.py
```diff
@@ -268,16 +268,24 @@ def get_docs(r_session, url, encoder=None, headers=None, **params):
...
+    properly operate as the callback.
+
+    :param response HTTP response object
+    :param kwargs HTTP request parameters
```
alfinkel: Again, probably does not matter right now but to be thorough, the syntax here should be `:param response:` and `:param kwargs:`

src/cloudant/_common_util.py
emlaver: I made the changes in the squash commit 737cd30.

src/cloudant/_common_util.py
emlaver: I made the changes in the squash commit 737cd30.

PR #264, opened 2016-12-22
259 improve error handling design

### LINKED ISSUE #259: Improve CloudantException design
Using the ResultException class as a guide, we can improve the error handling design by 1) have specific class exceptions for many or all of the modules (i.e. `CloudantClientException` and 2) move the error messages from `error.py` and the other modules to an internal `messages.py` module to reduce clutter.

After a discussion with @alfinkel, will follow these guidelines:
(1) try and preserve Python `dict` behavior
(2) only throwing Cloudant exceptions when trying to communicate errors encountered pre-remote database interaction

emlaver:
## What
Using `ResultException` as an example, we can improve the error handling design by having class exceptions for modules that currently contain `CloudantException`, and by moving the error messages from `error.py` and other moduless to an internal `messages.py` module to reduce clutter.

## How
- New exception classes for client, database, design document, document, feed, index, replicator modules
- All CloudantException and ResultException messages moved to _messages.py module
- Added status codes to all exception messages.
   - 4xx for status codes that resembles a typical response code and 1xx for other status codes.  
- Removed duplicate error message in `Client._usage_endpoint` by using boolean
- Fixed up exception messages in modules

## Testing
Fixed up exception messages in test cases.
Added new test cases for exception classes

## Issues
fixes #259 
#65

### DIFF
--- a/src/cloudant/_messages.py
+++ b/src/cloudant/_messages.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+# Copyright (c) 2016 IBM. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Module that contains exception messages for the Cloudant Python client
+library.
+"""
+CLIENT = {
+    100: 'A general Cloudant client exception was raised.',
+    101: 'Value must be set to a Database object. Found type: {0}.',
+    102: 'You must provide a url or an account.',
+    404: 'Database {0} does not exist. Verify that the client is valid and try again.',
+    409: 'Database {0} already exists.'
+}
+
+DATABASE = {
+    100: 'A general Cloudant database exception was raised.',
+    101: 'Unexpected index type. Found: {0}.',
+    400: 'Invalid database name during creation. Found: {0}',
+    401: 'Unauthorized to create database {0}.',
+    409: 'Document with id {0} already exists.'
+}
+
+DESIGN_DOCUMENT = {
+    100: 'A general Cloudant design document exception was raised.',
+    101: 'Cannot add a MapReduce view to a design document for query indexes.',
+    102: 'Cannot update a query index view using this method.',
+    103: 'Cannot delete a query index view using this method.',
+    104: 'View {0} must be of type View.',
+    105: 'View {0} must be of type QueryIndexView.',
+    106: 'Function for search index {0} must be of type string.',
+    107: 'Definition for query text index {0} must be of type dict.'
+}
+
+DOCUMENT = {
+    100: 'A general Cloudant document exception was raised.',
+    101: 'A document id is required to fetch document contents. '
+         'Add an _id key and value to the document and re-try.',
+    102: 'The field {0} is not a list.',
+    103: 'Attempting to delete a doc with no _rev. Try running .fetch and re-try.'
+}
+
+FEED = {
+    100: 'A general Cloudant feed exception was raised.',
+    101: 'Infinite _db_updates feed not supported for CouchDB.'
+}
+
+INDEX = {
+    100: 'A general Cloudant index exception was raised.',
+    101: 'Creating the \"special\" index is not allowed.',
+    102: 'Deleting the \"special\" index is not allowed.'
+}
+
+REPLICATOR = {
+    100: 'A general Cloudant replicator exception was raised.',
+    101: 'You must specify either a source_db Database object or a manually composed'
+         ' \'source\' string/dict.',
+    102: 'You must specify either a target_db Database object or a manually composed'
+         ' \'target\' string/dict.',
+    404: 'Replication with id {0} not found.'
+}
+
+RESULT = {
+    100: 'A general result exception was raised.',
+    101: 'Failed to interpret the argument {0} as a valid key value or as a valid slice.',
+    102: 'Cannot use {0} when performing key access or key slicing. Found {1}.',
+    103: 'Cannot use {0} for iteration. Found {1}.',
+    104: 'Invalid page_size: {0}'
+}
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -24,7 +24,7 @@ import requests
 from ._2to3 import bytes_, unicode_
 from .database import CloudantDatabase, CouchDatabase
 from .feed import Feed, InfiniteFeed
-from .error import CloudantException, CloudantArgumentError
+from .error import CloudantArgumentError, CloudantClientException
 from ._common_util import (
     USER_AGENT,
     append_response_error_content,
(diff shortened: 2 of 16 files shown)


src/cloudant/_messages.py
```diff
@@ -0,0 +1,67 @@
...
+         'Add an _id key and value to the document and re-try.',
+    101: 'The field {0} is not a list.',
+    102: 'Attempting to delete a doc with no _rev. Try running .fetch and re-try.'
+}
+FEED = {
+    100: 'Infinite _db_updates feed not supported for CouchDB.'
+}
+INDEX = {
+    100: 'Creating the \"special\" index is not allowed.',
+    101: 'Deleting the \"special\" index is not allowed.'
```
emlaver: Should we consider merging exception messages? i.e. `Creating or deleting the special index is not allowed.`

alfinkel: I think that consolidating exception messages is okay if they truly relay the same thing.  But I don't think that merging the message content of multiple messages that can stand on their own (like your example) does not buy you anything and can at times provide odd and misleading information based on the exception situation.  I would steer clear of doing this.  Our messages should provide useful information and there really is no cost to having _more_ messages.

src/cloudant/error.py
```diff
@@ -61,18 +71,116 @@ class ResultException(CloudantException):
...
+
+    :param int code: A code value used to identify the client exception.
+    :param args: A list of arguments used to format the exception message.
+    """
+    def __init__(self, code=None, *args):
+        try:
+            msg = CLIENT[code].format(*args)
+        # pylint: disable=broad-except
+        except Exception:
+            msg = CLIENT[code]
```
smithsz: I'm wondering if we should handle a potential`KeyError` here? Maybe use a default error code or something.

src/cloudant/error.py
```diff
@@ -61,18 +71,116 @@ class ResultException(CloudantException):
...
+    Provides a way to issue Cloudant library client specific exceptions.
+
+    :param int code: A code value used to identify the client exception.
+    :param args: A list of arguments used to format the exception message.
+    """
+    def __init__(self, code=None, *args):
+        try:
+            msg = CLIENT[code].format(*args)
+        # pylint: disable=broad-except
+        except Exception:
```
smithsz: Can we be more specific here and just catch`IndexError` and `KeyError`?

src/cloudant/_messages.py
emlaver: @alfinkel I agree, with this example it will be confusing to know which method threw the error.  As you said, better to keep the messages separate.

src/cloudant/error.py
emlaver: We could have a default `100` error code that says something like `A general Cloudant exception was raised.` in each exception class.  I'm not sure how helpful this message is but I also wouldn't want another exception to happen on this line.

src/cloudant/error.py
emlaver: Yea, I don't see a problem with being more specific i.e. `except (KeyError, IndexError)`

emlaver: Fixed in 4cfafb4.

emlaver:
@smithsz @alfinkel I'm considering changing the exception class names to module+exception i.e. from `CloudantClientException` to `ClientException`.  Any thoughts?

src/cloudant/error.py
emlaver: I've added a default code and message in 8fd83a0.  Test cases added for the new exception classes in 1239350.

src/cloudant/_messages.py
```diff
@@ -0,0 +1,73 @@
...
+Module that contains exception messages for the Cloudant Python client
+library.
+"""
+CLIENT = {
+    100: 'A general Cloudant client exception was raised.',
+    101: 'Value must be set to a Database object. Found type: {0}.',
+    102: 'You must provide a url or an account.',
+    404: 'Database {0} does not exist. Verify that the client is valid and try again.',
+    412: 'Database {0} already exists.'
+}
```
alfinkel: Nothing major but I think it would be good to separate these dictionary constants with a blank line (should apply to the entire module).

src/cloudant/_messages.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+Module that contains exception messages for the Cloudant Python client
+library.
+"""
+CLIENT = {
+    100: 'A general Cloudant client exception was raised.',
+    101: 'Value must be set to a Database object. Found type: {0}.',
+    102: 'You must provide a url or an account.',
+    404: 'Database {0} does not exist. Verify that the client is valid and try again.',
+    412: 'Database {0} already exists.'
```
alfinkel: Wouldn't this be a `409` to match your other "already exists" messages?

src/cloudant/client.py
```diff
@@ -198,37 +198,34 @@ def create_database(self, dbname, **kwargs):
...
 
         :returns: The newly created database object
         """
         new_db = self._DATABASE_CLASS(self, dbname)
         if new_db.exists():
             if kwargs.get('throw_on_exists', True):
-                raise CloudantException(
-                    "Database {0} already exists".format(dbname)
-                )
+                raise CloudantClientException(412, dbname)
```
alfinkel: `409` as suggested above?

src/cloudant/_messages.py
emlaver: Yes, I'll change this.

src/cloudant/error.py
```diff
@@ -16,6 +16,16 @@
 Module that contains common exception classes for the Cloudant Python client
```
alfinkel: You should probably change your copyright notice to `Copyright (c) 2015 IBM. All rights reserved.` for this module.

tests/unit/client_tests.py
```diff
@@ -233,19 +271,14 @@ def test_create_existing_database(self):
...
-            self.fail('Above statement should raise a CloudantException')
-        except CloudantException as err:
-            self.assertEqual(
-                str(err),
-                'Database {0} already exists'.format(dbname)
-                )
-        finally:
-            self.client.delete_database(dbname)
-            self.client.disconnect()
+        self.assertEqual(cm.exception.status_code, 412)
```
alfinkel: `409` as suggested earlier

tests/unit/replicator_tests.py
```diff
@@ -29,11 +29,49 @@
...
 from cloudant.document import Document
-from cloudant.error import CloudantException
+from cloudant.error import CloudantReplicatorException, CloudantClientException
 
 from .unit_t_db_base import UnitTestDbBase
 from .. import unicode_
 
+class CloudantReplicatorExceptionTests(unittest.TestCase):
+    """
+    Ensure CloudantDocumentException functions as expected.
```
alfinkel: CloudantReplicatorException

alfinkel:
Have you considered doing something similar to this with the CloudantArgumentError expections?

emlaver:
@alfinkel I've considered moving the `CloudantArgumentError` error messages to the `messages` module.  I currently don't see it being worthwhile to have separate classes for argument errors in each module.

src/cloudant/_messages.py
emlaver: Fixed in 8316529.

src/cloudant/_messages.py
emlaver: Fixed in 8316529.

src/cloudant/client.py
emlaver: Fixed in 8316529.

src/cloudant/error.py
emlaver: Fixed in 8316529.

tests/unit/client_tests.py
emlaver: Fixed in 8316529.

tests/unit/replicator_tests.py
emlaver: Fixed in 8316529.

alfinkel:
I agree that it's not worthwhile splitting the argument errors exceptions into their own classes.  However, my question was whether you considered moving _all_ argument error messages into your `_messages` module and basically handling argument error messages the same way as you handle the individual class exception messages.

PR #265, opened 2016-12-22
Fix failing tests

smithsz:
## What
Couple of minor test fixes,

- Altering the database N & Q values has changed the second element in the search result `order` array. This second element is only for troubleshooting and can be ignored in the tests (see https://docs.cloudant.com/search.html#query-syntax).
- Add some retrying around `test_create_replication` using Flaky.

### DIFF
--- a/tests/unit/database_tests.py
+++ b/tests/unit/database_tests.py
@@ -1411,35 +1411,24 @@ class CloudantDatabaseTests(UnitTestDbBase):
             limit=5,
             include_docs=True
         )
-        self.assertEqual(len(resp['rows']), 5)
+        self.assertEqual(5, len(resp['rows']))
         self.assertTrue(resp['bookmark'])
-        resp.pop('bookmark')
-        for row in resp['rows']:
-            self.assertTrue(row['doc']['_rev'].startswith('1-'))
-            row['doc'].pop('_rev')
-        self.assertEqual(
-            resp,
-            {'rows': [{'fields': {'name': 'julia'}, 'doc': {'_id': 'julia000',
-                                                            'age': 0,
-                                                            'name': 'julia'},
-                       'id': 'julia000', 'order': ['julia000', 0]},
-                      {'fields': {'name': 'julia'}, 'doc': {'_id': 'julia001',
-                                                            'age': 1,
-                                                            'name': 'julia'},
-                       'id': 'julia001', 'order': ['julia001', 0]},
-                      {'fields': {'name': 'julia'},'doc': {'_id': 'julia002',
-                                                           'age': 2,
-                                                           'name': 'julia'},
-                       'id': 'julia002', 'order': ['julia002', 0]},
-                      {'fields': {'name': 'julia'}, 'doc': {'_id': 'julia003',
-                                                            'age': 3,
-                                                            'name': 'julia'},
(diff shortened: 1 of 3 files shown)


tests/unit/replicator_tests.py
```diff
@@ -116,6 +118,7 @@ def test_replication_with_generated_id(self):
         clone = Replicator(self.client)
         clone.create_replication(self.db, self.target_db)
 
+    @flaky(max_runs=3)
```
emlaver: Is there an alternative to fixing this test so we don't have to rely on `flaky`? Is it worth trying to adjust the N and Q values for the node? We've had this test pass consistently in the past.

rhyshort: N is set to 1, so I doubt tweaking the values some more will make a difference tbh. @smithsz thoughts?

emlaver: I went back into the Jenkins build history and only found one build on Oct 20th that had `test_create_replication` failing.  From where I'm standing, it seems like there's something else going on here.

tests/unit/database_tests.py
```diff
@@ -1413,32 +1413,14 @@ def test_get_search_result_executes_search_query(self):
...
-                      {'fields': {'name': 'julia'}, 'doc': {'_id': 'julia003',
-                                                            'age': 3,
-                                                            'name': 'julia'},
-                       'id': 'julia003', 'order': ['julia003', 0]},
-                      {'fields': {'name': 'julia'},
-                       'doc': {'_id': 'julia004', 'age': 4,
-                               'name': 'julia'},
-                       'id': 'julia004', 'order': ['julia004', 1]}], 'total_rows': 100}
+            ['julia000', 'julia001', 'julia002', 'julia003', 'julia004'],
+            [row['id'] for row in resp['rows']]
```
emlaver: This is only asserting `_id`. What about asserting the other fields? i.e. `age`, `name`, `fields`, `order`

tests/unit/database_tests.py
```diff
@@ -1479,25 +1460,17 @@ def test_get_search_result_executes_search_query_with_group_option(self):
...
+        results = []
+        for group in resp['groups']:
+            results.append(group['by'])
+            self.assertEqual(1, group['total_rows'])
+            self.assertEqual(1, len(group['rows']))
+            self.assertEqual('julia', group['rows'][0]['fields']['name'])
+
         self.assertEqual(
-            resp,
-            {'total_rows': 100, 'groups': [
```
emlaver: We were asserting the contents of `rows` before, why not continue that in the for loop?

tests/unit/database_tests.py
smithsz: I'll add.

tests/unit/database_tests.py
smithsz: Will do.

tests/unit/replicator_tests.py
smithsz: I'm unsure. The test looks fine to me in that it creates a valid rep document. The issue seems to be with the service not reliably starting the replication. Maybe a latent bug?

smithsz: I'll see if I can get a repro script together to demonstrate this failure to the DBcore team. In the meantime lets rely on flaky to get this passing.

PR #269, opened 2017-01-04
249 add note and example for 429 adapter in getting_started.rst

### LINKED ISSUE #249: Add Replay429Adapter to the getting started documentaton
Any application connecting to cloudant ccm will probably require using a Replay429Adapter?

If so, we should probably document this in the getting started documentation?

emlaver:
## What

- Added a note to use `Replay429Adapter` if retrying requests is needed after a 429 error
- Added `client` construction example that's configured with 429 adapter

## Testing

No tests added.

## Issues

fixes #249

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -23,6 +23,10 @@ then disconnecting from the server.
 Later in the `Context managers`_ section we will see how to 
 simplify this process through the use of the Python *with* statement.
 
+Note: If you require retrying requests after an HTTP 429 error, the
+``Replay429Adapter`` can be added when constructing a ``Cloudant``
+client and configured with an initial back off and retry count.
+
 Connecting with a client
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -23,6 +23,10 @@ then disconnecting from the server.
 Later in the `Context managers`_ section we will see how to 
 simplify this process through the use of the Python *with* statement.
 
+Note: If you require retrying requests after an HTTP 429 error, the
+``Replay429Adapter`` can be added during the ``Cloudant`` client
+construction and configured with an initial back off and retry count.
```
rhyshort: Sounds a little odd to me, maybe " can be added when constructing a Cloudant Client..."

emlaver: Fixed in 6778514.

PR #270, opened 2017-01-04
237 add Python 2 long type support to argument types

### LINKED ISSUE #237: Support large numbers with long type
A customer gets an argument error with a large number value:

```
view_result = my_db.get_view_result( "design_docs_id", "view_id_with_very_big_key", raw_result=False, startkey = 10000000000, endkey = 20000000000, include_docs = True)
      print view_result[0]
```

Error callstack:

```
Traceback (most recent call last):
  File "C:\Users\john.doe\workspace\GreatestPythonApp\some\pckg\path\some_module_name.py", line 44, in <module>
    print view_result[0]
  File "C:\Python27\lib\site-packages\cloudant\result.py", line 203, in __getitem__
    data = self._handle_result_by_index(arg)
  File "C:\Python27\lib\site-packages\cloudant\result.py", line 238, in _handle_result_by_index
    return self._ref(skip=skip+idx, limit=1, **opts)
  File "C:\Python27\lib\site-packages\cloudant\view.py", line 230, in __call__
    **kwargs)
  File "C:\Python27\lib\site-packages\cloudant\_common_util.py", line 254, in get_docs
    f_params = python_to_couch(params)
  File "C:\Python27\lib\site-packages\cloudant\_common_util.py", line 167, in python_to_couch
    py_to_couch_validate(key, val)
  File "C:\Python27\lib\site-packages\cloudant\_common_util.py", line 187, in py_to_couch_validate
cloudant.error.CloudantArgumentError: Argument startkey not instance of expected type: (<type 'int'>, <type 'basestring'>, <class '_abcoll.Sequence'>)
```

Adding `long` type for validating startkey, endkey, and other arguments will fix this issue.

emlaver:
## What

- Added `LONGTYPE` in `_2to3.py` to support `long` in `common_util` argument types

## Testing

- Added `long` testing in existing parameter translation tests
- Added `test_all_docs_get_with_long_type` test case 

## Issues

fixes #237

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -23,7 +23,7 @@ from collections import Sequence
 import json
 from requests import Session
 
-from ._2to3 import STRTYPE, NONETYPE, UNITYPE, iteritems_, url_parse
+from ._2to3 import LONGTYPE, STRTYPE, NONETYPE, UNITYPE, iteritems_, url_parse
 from .error import CloudantArgumentError, CloudantException
 
 # Library Constants
@@ -50,19 +50,19 @@ SPECIAL_INDEX_TYPE = 'special'
 
 RESULT_ARG_TYPES = {
     'descending': (bool,),
-    'endkey': (int, STRTYPE, Sequence,),
+    'endkey': (int, LONGTYPE, STRTYPE, Sequence,),
     'endkey_docid': (STRTYPE,),
     'group': (bool,),
-    'group_level': (int, NONETYPE,),
+    'group_level': (int, LONGTYPE, NONETYPE,),
     'include_docs': (bool,),
     'inclusive_end': (bool,),
-    'key': (int, STRTYPE, Sequence,),
+    'key': (int, LONGTYPE, STRTYPE, Sequence,),
     'keys': (list,),
-    'limit': (int, NONETYPE,),
+    'limit': (int, LONGTYPE, NONETYPE,),
     'reduce': (bool,),
-    'skip': (int, NONETYPE,),
+    'skip': (int, LONGTYPE, NONETYPE,),
     'stale': (STRTYPE,),
-    'startkey': (int, STRTYPE, Sequence,),
+    'startkey': (int, LONGTYPE, STRTYPE, Sequence,),
     'startkey_docid': (STRTYPE,),
 }
 
@@ -75,6 +75,7 @@ TYPE_CONVERTERS = {
     list: lambda x: json.dumps(x),
     tuple: lambda x: json.dumps(list(x)),
     int: lambda x: x,
+    LONGTYPE: lambda x: x,
     bool: lambda x: 'true' if x else 'false',
     NONETYPE: lambda x: x
 }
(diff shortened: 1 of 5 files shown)


tests/unit/database_tests.py
```diff
@@ -442,6 +442,17 @@ def test_all_docs_get(self):
         self.assertEqual(data['rows'][1]['key'], 'julia011')
         self.assertEqual(data['rows'][2]['key'], 'julia012')
 
+    def test_all_docs_get_with_long_type(self):
+        """
+        Test the all_docs GET request functionality
+        """
+        self.populate_db_with_documents()
+        data = self.db.all_docs(limit=10000000000, skip=10)
```
rhyshort: When I check the type of `10000000000`  in the interpreter it is `int` rather than long`. 

```bash
~> python 
Python 2.7.10 (default, Jul 30 2016, 19:40:32) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> o = 10000000000
>>> type(o)
<type 'int'>
>>> exit()
```

I think we need to suffix `10000000000` with `L` so we have `10000000000L`

smithsz: yeah, needs to bigger than `sys.maxint` to yield a long int, or add the `L` suffix.

emlaver: @rhyshort @smithsz Test cases won't pass with `L` suffix since all integers in Python 3 are long integers.  I've added a `LONG_NUMBER` variable that's bigger than `sys.maxint` in 79e1e34.

src/cloudant/_common_util.py
```diff
@@ -124,19 +125,22 @@
     'group_sort': (STRTYPE, list),
     'include_docs': bool,
     'limit': (int, NONETYPE),
-    'query': (STRTYPE, int),
-    'q': (STRTYPE, int),
+    'query': (STRTYPE, int, LONGTYPE),
```
smithsz: In Py3 `LONGTYPE` is `NONETYPE`. We shouldn't allow NoneTypes to be passed as the query param.

smithsz: We could have a NUMBER_TYPES set in the _2to3.py to get round this...
```
from ._2to3 import NUMBER_TYPES

'query': (STRTYPE, ) + NUMBER_TYPES
```

src/cloudant/_common_util.py
```diff
@@ -124,19 +125,22 @@
...
     'highlight_post_tag': STRTYPE,
-    'highlight_number': (int, NONETYPE),
-    'highlight_size': (int, NONETYPE),
+    'highlight_number': (int, LONGTYPE, NONETYPE),
+    'highlight_size': (int, LONGTYPE, NONETYPE),
     'include_fields': list
 }
 
+# Test long type in Python 2
+LONG_NUMBER = 92233720368547758070
```
smithsz: This should be in test code. Maybe add a test utils?
Also, not sure I like the magic number here. Maybe `LONG_NUMBER = PY2 and 1L or 1` instead.

src/cloudant/_common_util.py
emlaver: @smithsz It may be better to change LONGTYPE to  `LONGTYPE = long if PY2 else int` since all integers are long ints in Py3.

src/cloudant/_common_util.py
emlaver: I tried that before but `SyntaxError: invalid syntax` will always be thrown if `L` syntax exists when running tests in Python 3.

src/cloudant/_common_util.py
smithsz: yeah, that's better 👍

src/cloudant/_common_util.py
smithsz: oh, annoying. how about `LONG_NUMBER = PY2 and long(1) or 1`?

emlaver: That works.  I've put the `LONG_NUMBER` constant in a new test util module in b5a36b5.

src/cloudant/_common_util.py
emlaver: I've fixed this in b5a36b5.

PR #272, opened 2017-01-06
18 add connection pooling to getting started documentation

### LINKED ISSUE #18: Documentation: connection pooling
One of the failings of the currently python library is that it doesn't describe how to use connection pooling.

This ticket is to track documentation that is quite important to developers around setting up and configuring connection pooling for this library.  For example:
1. information on instantiating the library so that it can be shared, e.g. using a singleton
2. calling out which methods are threadsafe and which aren't
3. describe how to configure connection pooling
4. describe how the library authenticates to Cloudant (using cookie or basic authentication)
5. describe if session expiry gets handled automatically by the library [1].  if not, what does the library consumer need to do to handle this?
6. how to use the library in common python application server environments [2] so that connection pooling can be configured at the app server level

---

[1] this assumes the library uses cookie authentication.  The [docs](https://docs.cloudant.com/authentication.html#cookie-authentication) state that the cookie can be used until it expires.
[2] e.g. django, flask, gunicorn

emlaver:
## What
- information on instantiating the library so that it can be shared
- thread safety from `urllib3` used in `requests` library
- describe how to configure connection pooling
- describe how the library authenticates to Cloudant (using cookie or basic authentication)
- describe if session expiry gets handled automatically by the library
- how to use the library in common python application server environments (e.g. django, flask, gunicorn) so that connection pooling can be configured at the app server level

## How
- Added connection pooling, resources, and cookie auth to `getting_started.rst`

## Issues
fixes #18

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -62,6 +62,101 @@ Connecting with a client
     # Disconnect from the server
     client.disconnect()
 
+**************
+Authentication
+**************
+
+When constructing a ``Cloudant`` client, you can authenticate using the
+[cookie authentication](http://guide.couchdb.org/editions/1/en/security.html#cookies) functionality.
+The server will always attempt to automatically renew the cookie
+shortly before its expiry. However, if the client does not send a
+request to the server during this renewal window and
+``auto_renew=False`` then the cookie is not renewed.
+
+Using ``auto_renew=True`` will attempt to renew the cookie at
+any point during the lifetime of the session when either of the
+following statements hold true:
+
+- The server returns a ``credentials_expired`` error message.
+- The server returns a ``401 Unauthorized`` status code.
+- The server returns a ``403 Forbidden`` status code.
+
+.. code-block:: python
+
+    # Create client using auto_renew to automatically renew expired cookie auth
+    client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com',
+                     connect=True,
+                     auto_renew=True)
+
+****************
+Resource sharing
+****************
+
+The ``Cloudant`` or ``CouchDB`` client objects make HTTP calls using the ``requests`` library.
+``requests`` uses the [urllib3](https://pypi.python.org/pypi/urllib3) library which features
+connection pooling and thread safety.
+
+Connection pools can be managed by using the ``requests`` library's
+[HTTPAdapter](https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L78)
+when constructing a ``Cloudant`` or ``ClouchDB`` client instance.
+The default number set by the ``urllib3`` library for cached connection pools is 10.
+Use the ``HTTPAdapter`` argument ``pool_connections`` to set the number of
+urllib3 connection pools to cache, and the ``pool_maxsize`` argument to set the
+maximum number of connections to save in the pool.
+
+Although the ``client`` session is documented as thread safe and it's possible for a
+static ``client`` to be accessible by multiple threads, there are still cases that do not
+guarantee thread safe execution.  It's recommended to use one ``client`` object per thread.
+
+.. code-block:: python
+
+    # Create client with 15 cached pool connections and a max pool size of 100
+    httpAdapter = HTTPAdapter(pool_connections=15, pool_maxsize=100)
+    client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com'
+                     connect=True,
+                     adapter=httpAdapter)
+
+Note: Idle connections within the pool may be terminated by the server, so will not remain open
+indefinitely meaning that this will not completely remove the overhead of creating new connections.
+
+Using library in app server environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This library can be used in an app server, and the example
+below shows how to use ``client`` in a ``flask`` app server.
+
+.. code-block:: python
+
+   from flask import Flask
+   import atexit
+
+   app = Flask(__name__)
+
+   @app.route('/')
+   def hello_world():
+      # Cookie authentication can be renewed automatically using ``auto_renew=True``
+      # which is typically what you would require when running in an application
+      # server where the connection may stay open for a long period of time
+
+      # Note: Each time you instantiate an instance of the Cloudant client, an
+      # authentication request will be made to Cloudant to retrieve the session cookie.
+      # If the performance overhead of this call is a concern for you, consider
+      # using vanilla python requests with a custom subclass of HTTPAdapter that
+      # performs the authentication call to Cloudant when it establishes the http
+      # connection during the creation of the connection pool.
+      client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com',
+                        connect=True,
+                        auto_renew=True)
+
+      # do something with client
+      return 'Hello World!'
+
+   # When shutting down the app server, use ``client.disconnect()`` to properly
+   # logout and end the ``client`` session
+   @atexit.register
+   def shutdown():
+      client.disconnect()
+
 *********
 Databases
 *********


docs/getting_started.rst
```diff
@@ -54,6 +54,49 @@ Connecting with a client
...
+The default number set by the ``urllib3`` library for cached connection pools is 10.
+Use the ``HTTPAdapter`` argument ``pool_connections`` to set the number of
+urllib3 connection pools to cache, and the ``pool_maxsize`` argument to set the
+maximum number of connections to save in the pool.
+
+.. code-block:: python
+
+    # Create client with 15 cached pool connections and a max pool size of 100
+    httpAdapter = HTTPAdapter(pool_connections=15, pool_maxsize=100)
+    client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com'
```
snowch: Can the `client` reference be shared between multiple threads, or does each thread need it's own `Cloudant` instance?

docs/getting_started.rst
```diff
@@ -54,6 +54,49 @@ Connecting with a client
...
+If you need the cookie to automatically renew when it expires,
+construct the client using ``auto_renew=True``:
+
+.. code-block:: python
+
+    # Create client using auto_renew to automatically renew expired cookie auth
+    client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com',
+                     connect=True,
+                     auto_renew=True)
+
```
snowch: What is the default value for `auto_renew`?  cloudant-python clients running in an app server will connect() at app server startup and disconnect() at app server shutdown and will therefore most likely need this set to `True`?

docs/getting_started.rst
```diff
@@ -54,6 +54,49 @@ Connecting with a client
...
+maximum number of connections to save in the pool.
+
+.. code-block:: python
+
+    # Create client with 15 cached pool connections and a max pool size of 100
+    httpAdapter = HTTPAdapter(pool_connections=15, pool_maxsize=100)
+    client = Cloudant(USERNAME, PASSWORD, url='https://acct.cloudant.com'
+                     connect=True,
+                     adapter=httpAdapter)
+
```
emlaver: @snowch After some testing with `Thread`, you can share the `client` with multiple threads.  This is the example I created to test sharing `client` with threads:  https://gist.github.com/emlaver/0a5b8227088e6f6812f6d1061ea78d76. I used the logging code from your stackoverflow question (http://stackoverflow.com/questions/29046503/cloudant-python-https-connection-pooling) to check that only one https connection was created for all the threads.

docs/getting_started.rst
emlaver: The default value for `auto_renew` is `False`.  Would something like this suffice?
```
This library can be used in app server environments (i.e. django).
When connecting to ``Cloudant`` or ``CouchDB``, you have the option of using the
``connect=True`` argument during ``client`` construction or by calling ``client.connect()``.
If the app server requires automatic renewal of the cookie authentication, add
the ``auto_renew=True`` argument when constructing the ``client``.
When disconnecting, use ``client.disconnect()`` to properly logout and end a client session.
```
This would be further down in the documentation so the user would already have seen how to construct a `client`.

docs/getting_started.rst
emlaver: @snowch Added a comment in 10889e5.

docs/getting_started.rst
emlaver: @snowch Added a section for app server in 10889e5.

docs/getting_started.rst
```diff
@@ -86,6 +86,9 @@ Use the ``HTTPAdapter`` argument ``pool_connections`` to set the number of
 urllib3 connection pools to cache, and the ``pool_maxsize`` argument to set the
 maximum number of connections to save in the pool.
 
+As long as the state of ``client`` does not changed it will be
+thread safe and accessible as a static variable to all threads.
+
```
snowch: @emlaver sorry to be pedantic, it would be good to let users know how they can find out what changes the state of the client.

docs/getting_started.rst
```diff
@@ -97,6 +100,16 @@ maximum number of connections to save in the pool.
 Note: Idle connections within the pool may be terminated by the server, so will not remain open
 indefinitely meaning that this will not completely remove the overhead of creating new connections.
 
+Using library in app server environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This library can be used in an app server e.g. django and flask. When connecting
+to ``Cloudant`` or ``CouchDB``, you have the option of using the ``connect=True``
+argument during ``client`` construction or by calling ``client.connect()``.
+If the app server requires automatic renewal of the cookie authentication, add
```
snowch: how does the user find out if an app server does or doesn't require automatic renewal of the cookie authentication?

docs/getting_started.rst
```diff
@@ -97,6 +100,16 @@ maximum number of connections to save in the pool.
...
+the ``auto_renew=True`` argument when constructing the ``client``.
+When disconnecting, use ``client.disconnect()`` to properly logout and end a client session.
```
snowch: When disconnecting from what?

emlaver:
@snowch Thanks for the sample `Flask` code.  I've used it in 7d3adbe.

docs/getting_started.rst
```diff
@@ -54,6 +54,83 @@ Connecting with a client
...
+
+Connection pools can be managed by using the ``requests`` library's
+[HTTPAdapter](https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L78)
+when constructing a ``Cloudant`` or ``ClouchDB`` client instance.
+The default number set by the ``urllib3`` library for cached connection pools is 10.
+Use the ``HTTPAdapter`` argument ``pool_connections`` to set the number of
+urllib3 connection pools to cache, and the ``pool_maxsize`` argument to set the
+maximum number of connections to save in the pool.
+
+As long as the state of ``client`` does not changed it will be
```
smithsz: *change

smithsz: The requests Session is documented as threadsafe but there still appear to be a couple of corner cases where it isn’t perfectly threadsafe.
Perhaps we'd be safer advising folks to use a single Cloudant Client/connection per thread here?

docs/getting_started.rst
```diff
@@ -54,6 +54,83 @@ Connecting with a client
...
     client.disconnect()
 
+**************
+Authentication
+**************
+
+When constructing a ``Cloudant`` client, you can authenticate using the
+[cookie authentication](http://guide.couchdb.org/editions/1/en/security.html#cookies) functionality.
+If you need the cookie to automatically renew when it expires,
+construct the client using ``auto_renew=True``:
```
emlaver: Yes, I will go ahead and add these details.

docs/getting_started.rst
emlaver: @smithsz @snowch we could say something like:
```
Although the `requests` Session is documented as thread-safe and it's possible for 
a static `client` to be accessible by multiple threads, there are still cases that do not
guarantee safe execution.  It's recommended to use a single Cloudant `client` per thread.
```

docs/getting_started.rst
emlaver: @snowch I'm going to remove this. The `requests toolbelt` library (and the comment from Sam) says that although `requests` is documented as theadsafe there are cases where it's not perfectly threadsafe: http://toolbelt.readthedocs.io/en/latest/threading.html.  I think it's best to [add a comment that recommends one `client` per thread.](https://github.com/cloudant/python-cloudant/pull/272#discussion_r95386901)

docs/getting_started.rst
emlaver: Will remove this with a [new comment about `client` per thread.](https://github.com/cloudant/python-cloudant/pull/272#discussion_r95386901)

docs/getting_started.rst
```diff
@@ -54,6 +54,94 @@ Connecting with a client
...
+Note: Idle connections within the pool may be terminated by the server, so will not remain open
+indefinitely meaning that this will not completely remove the overhead of creating new connections.
+
+Using library in app server environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This library can be used in an app server, and the example
+below shows how to use ``client`` in a ``flask`` app server.
+
+.. code-block:: python
```
emlaver: @snowch @ricellis Updated flask example using a global `Session` to construct one `client` per thread: https://gist.github.com/emlaver/5a6909ad0ac801de3b492226265bf6db

emlaver: @snowch @ricellis I think it's best to move forward without the global variables and construct one client per thread.  The example in the above comment adds confusion and I don't have enough evidence that performance will be all that better.
```
...
   @app.route('/')
   def hello_world():
      # Cookie authentication can be renewed automatically using ``auto_renew=True``
      # which is typically what you would require when running in an application
      # server where the connection may stay open for a long period of time
      client = Cloudant(USERNAME, PASSWORD,
                                    url='https://acct.cloudant.com',
                                    auto_renew=True,
                                    connect=True)

      # do something with client
      return 'Hello World!'
...
```

ricellis: I agree with that.

snowch: Will instantiating the client in the http handler result in a call being made to the cloudant auth endpoint every time a http request is handled?

emlaver: @snowch @ricellis From my flask testing, a new auth cookie is created each time the client is constructed which you've mentioned in the `NOTE` comment.

ricellis: I'm not familiar with flask, but my reading of this is that each defined `@app.route` is a HTTP handler and they can be serviced on multiple threads. Due to the ambiguity of requests/urllib3 around thread-safe cookie handling we are saying it is unsafe to access a single client object from multiple threads and hence we can't just store a cloudant `client` for use by all the http handlers. Instantiating a `client` on each handler call obviously has a performance penalty.

The alternative solution in my mind is to create a pool of `client` resources and lease one in the http handler? Similar to the way a  JDBC connection pool operates, does flask offer that kind of resource pooling? If it was done that way then the `client`s would be instantiated in the resource pool and used by only 1 HTTP handler thread at a time?

ricellis: http://flask.pocoo.org/docs/0.12/tutorial/dbcon/#step-4-database-connections ?

snowch: fyi - http://stackoverflow.com/questions/41744745/how-to-extend-urllib3-connection-creation-to-add-additional-authentication-step

emlaver: @ricellis @snowch I've updated the flask app with Chris's note in 4c5e55b.  I suggest that we keep the note, finish reviewing and close out this PR, and open up a new issue to continue the discussion around an alternate solution for resource pooling in app servers.

emlaver: @snowch Opened issue #275.

snowch:
+1

PR #274, opened 2017-01-20
257 timeout argument

### LINKED ISSUE #257: Specify a default timeout that isn't forever
The [Requests library does not have a timeout set by default](http://docs.python-requests.org/en/master/user/advanced/#timeouts). To prevent python-cloudant waiting forever for a HTTP connection or a response we should set a reasonable default timeout on all requests and allow customization.

    - All 2.x
- Python version
    - Any

emlaver:
## What

Allow for a timeout argument in `client` construction to prevent python-cloudant waiting forever for a HTTP connection or a response on all requests.

## How

- ClientSession class to add timeout of to all requests

## Testing

New tests to assert that connect and read timeout exception occurs during client construction and replication.

## Issues

fixes #257

### DIFF
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (C) 2015, 2016 IBM Corp. All rights reserved.
+# Copyright (C) 2015, 2016, 2017 IBM Corp. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -19,7 +19,6 @@ instance.
 import base64
 import json
 import posixpath
-import requests
 
 from ._2to3 import bytes_, unicode_
 from .database import CloudantDatabase, CouchDatabase
@@ -28,8 +27,8 @@ from .error import CloudantArgumentError, CloudantClientException
 from ._common_util import (
     USER_AGENT,
     append_response_error_content,
-    InfiniteSession
-)
+    InfiniteSession,
+    ClientSession)
 
 
 class CouchDB(dict):
@@ -56,6 +55,15 @@ class CouchDB(dict):
     :param bool auto_renew: Keyword argument, if set to True performs
         automatic renewal of expired session authentication settings.
         Default is False.
+    :param float timeout: Timeout in seconds (use float for milliseconds, for
+        example 0.1 for 100 ms) for connecting to and reading bytes from the
+        server.  If a single value is provided it will be applied to both the
+        connect and read timeouts.  To specify different values for each timeout
+        use a tuple.  For example, a 10 second connect timeout and a 1 minute
+        read timeout would be (10, 60).  This follows the same behaviour as the
+        `Requests library timeout argument
+        <http://docs.python-requests.org/en/master/user/quickstart/#timeouts>`_.
+        but will apply to every request made using this client.
     """
     _DATABASE_CLASS = CouchDatabase
 
@@ -69,6 +77,7 @@ class CouchDB(dict):
         self.admin_party = admin_party
         self.encoder = kwargs.get('encoder') or json.JSONEncoder
         self.adapter = kwargs.get('adapter')
+        self._timeout = kwargs.get('timeout', None)
         self.r_session = None
         self._auto_renew = kwargs.get('auto_renew', False)
         connect_to_couch = kwargs.get('connect', False)
@@ -87,10 +96,16 @@ class CouchDB(dict):
             self.r_session = InfiniteSession(
                 self._user,
                 self._auth_token,
-                self.server_url
+                self.server_url,
+                timeout=self._timeout
             )
         else:
-            self.r_session = requests.Session()
+            self.r_session = ClientSession(
+                self._user,
+                self._auth_token,
+                self.server_url,
+                timeout=self._timeout
+            )
         # If a Transport Adapter was supplied add it to the session
         if self.adapter is not None:
             self.r_session.mount(self.server_url, self.adapter)
--- a/tests/unit/replicator_tests.py
+++ b/tests/unit/replicator_tests.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015 IBM. All rights reserved.
+# Copyright (c) 2015, 2016, 2017 IBM Corp. All rights reserved.
 #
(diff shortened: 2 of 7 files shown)


tests/unit/changes_tests.py
```diff
@@ -353,6 +355,20 @@ def test_get_feed_using_since_now(self):
         expected = set(['julia003', 'julia004', 'julia005'])
         self.assertSetEqual(set([x['id'] for x in changes]), expected)
 
+    def test_timeout_occurs_getting_feed(self):
+        """
+        Test getting content back for a feed using since set to "now"
+        """
+        self.set_up_client(auto_connect=True, timeout=1)
```
ricellis: Maybe we could make this 0.1 seconds or something to make the test faster?

tests/unit/replicator_tests.py
```diff
@@ -210,6 +211,47 @@ def test_create_replication(self):
             ])
         )
 
+    def test_timeout_in_create_replication(self):
```
ricellis: Any particular reasoning behind using the creation of a replication for testing the timeout? Does it clean up the replicator document correctly afterwards?

tests/unit/replicator_tests.py
```diff
@@ -210,6 +211,47 @@ def test_create_replication(self):
...
+        """
+        Test that a read timeout exception is thrown when creating a
+        replicator with a timeout value of 1 second.
+        """
+        # Setup client with a timeout
+        self.set_up_client(auto_connect=True, timeout=1)
```
ricellis: Again we could probably shorten the timeout.

tests/unit/unit_t_db_base.py
```diff
@@ -131,7 +131,8 @@ def setUp(self):
         """
         self.set_up_client()
 
-    def set_up_client(self, auto_connect=False, auto_renew=False, encoder=None):
+    def set_up_client(self, auto_connect=False, auto_renew=False, encoder=None,
+                      timeout=300):
```
ricellis: if the default is 300 why do we need to specify this here?

emlaver: You're right, no need to specify it here.

tests/unit/replicator_tests.py
emlaver: `test_create_replication` was one of a few tests that failed consistently when `timeout=1` and so I went ahead with creating this unit test.   The tests are even more unstable with `timeout=.1`.

emlaver: Maybe we should consider running a timeout test in each module.

ricellis: I don't think it is necessary given that we'd just be testing underlying Requests functionality. One test of connect and one test of read would be sufficient for me that the pass-through of args is working as we desire.

emlaver: Yes, I agree.  I'll add a connect test.

src/cloudant/client.py
```diff
@@ -56,6 +55,13 @@ class CouchDB(dict):
     :param bool auto_renew: Keyword argument, if set to True performs
         automatic renewal of expired session authentication settings.
         Default is False.
+    :param int timeout: the read and connect timeout.  The connect timeout
+        is the number of seconds Request will wait for the client to establish a
+        connection to a Cloudant or a CouchDB account. The read timeout is the
+        number of seconds the client will wait for Cloudant or CouchDB to send
+        a response.  This timeout value will be applied to both the connect and
+        the read timeouts.  Specify a tuple if you would like to set the values
+        separately.  Default is 5 minutes.
```
emlaver: It may be best to leave it as `None` until the next version.  I'm under the assumption that more people would look at the CHANGES log for a major version and review for any behavior changes.

src/cloudant/client.py
```diff
@@ -56,6 +55,13 @@ class CouchDB(dict):
     :param bool auto_renew: Keyword argument, if set to True performs
         automatic renewal of expired session authentication settings.
         Default is False.
+    :param int timeout: the read and connect timeout.  The connect timeout
```
emlaver: Updated in 66ded8d.

src/cloudant/client.py
emlaver: I've changed it to `None` in 66ded8d but I'm open to changing it to another (possibly longer than 5 mins) default timeout.

tests/unit/replicator_tests.py
emlaver: Added connect test in 66ded8d.

tests/unit/unit_t_db_base.py
emlaver: Set to `None`  in 66ded8d.

tests/unit/unit_t_db_base.py
```diff
@@ -132,7 +132,7 @@ def setUp(self):
         self.set_up_client()
 
     def set_up_client(self, auto_connect=False, auto_renew=False, encoder=None,
-                      timeout=300):
```
ricellis: I actually think it would be good if all our tests had a timeout. I just didn't think we needed to specify `300` if that was the default. Given that we're not applying a default until the next major we should probably apply one in all of our tests (except where it is obviously overridden to test the behaviour of the timeouts). Maybe `(30, 300)`?

emlaver: Added a timeout in ea2928c.

emlaver:
@ricellis Updated getting started with a note and example in ea2928c.

ricellis:
+1 but don't forget to update the copyright years in the modified files.

tests/unit/replicator_tests.py
```diff
@@ -210,6 +211,46 @@ def test_create_replication(self):
...
+        # Test that the replication document was created
+        expected_keys = ['_id', '_rev', 'source', 'target', 'user_ctx']
+        # If Admin Party mode then user_ctx will not be in the key list
+        if self.client.admin_party:
+            expected_keys.pop()
+        self.assertTrue(all(x in list(repl_doc.keys()) for x in expected_keys))
+        self.assertEqual(repl_doc['_id'], repl_id)
+        self.assertTrue(repl_doc['_rev'].startswith('1-'))
+        # Now that we know that the replication document was created,
+        # check that the replication occurred.
```
brynh: Aren't we checking that the replication timed-out rather than it occurred?

emlaver: Yes, fixed in 35205c5.

Issue #67, opened 2015-12-24
Optimize performance


alfinkel:
We need to consider providing a way for the developer to optimize some of the functionality provided by the library.
- One possible thing to consider is to add a "refresh" local cache attribute to the account, database, and document objects.  This will allow the developer to have more control over the library objects that they use.  Right now account, database, and document extend dict so with every operation performed to an object that updates it remotely raises two compelling questions:
  - Do you automatically refresh the local cached version for consistency/clarity purposes??
  - Do you not automatically refresh the local cached version for performance reasons??
- Right now, each object class (account, database, document) and even each method within the class seem to handle these two questions inconsistently.
- Another consideration that belongs as part of this case and that is somewhat related to the bullet above is the way in which attachments are handled via a document object.  Specifically, when adding or updating attachments we currently perform an extra fetch to refresh the document local dictionary.  It would be beneficial if there was a way to remove this last fetch and instead refresh the local document dictionary based on response content and headers, if possible.
  - See this related PR discussion:  https://github.com/cloudant/python-cloudant/pull/28/files#r44277879

robolivable:
I ran into an issue with the local caching feature of this library when utilizing it with multiple Celery workers to write millions of documents. Within an hour or so of running 5+ workers on a single box, the worker execution times would slow down, as the memory consumption grew drastically (execution times slowed down, I presume due to memory management with swap space).

I was able to work around the issue by making manual detour calls to the database using the internal requests session instance, to bypass the cache. Anyway, the point I'm getting at is that this library can probably also benefit from a "local cache limit" or a "clear local cache" function to provide some more control there.

ricellis:
We're not making enhancements here our new [cloudant-python-sdk](https://github.com/IBM/cloudant-python-sdk/)(beta) does not have the dict caching so these problems do not materialize there.

PR #277, opened 2017-01-30
Implement local cache limit

smithsz:
## What
Implement an LRU caching policy when storing documents locally.

By default the cache is not bound to a limit and can therefore grow indefinitely (thus copying the current behaviour).

_Example:_
```python

# no cache limit (default behaviour)
client1 = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME)

# limit the local cache to 100 documents
client2 = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME, cache_size=100)

# don't cache documents locally. always fetch from the remote server.
client3 = Cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME, cache_size=0)
```

## How
`CouchDatabase` types now inherit `LRUDict` which uses an `OrderedDict` under the hood to implement a simple LRU cache.

## Testing

## Issues
See #67.

### DIFF
--- a/src/cloudant/lru_dict.py
+++ b/src/cloudant/lru_dict.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+# Copyright (c) 2017 IBM. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+A simple `dict` like collection implementing a Least Recently Used (LRU)
+caching policy.
+"""
+
+from collections import MutableMapping, OrderedDict
+from threading import Lock
+
+
+class NullContext(object):
+    """
+    No-op context manager, executes block without doing any additional
+    processing.
+    """
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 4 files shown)


src/cloudant/lru_dict.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def _update_lru(self, key):
+        """
+        Move key to the top of the LRU cache.
+
+        :param str key: dictionary key
+        """
+        move_to_end = getattr(self._dict, 'move_to_end', None)
+        if move_to_end:
+            move_to_end(key, last=True)  # >=Python3.2 only
```
emlaver: If this code only works for Python 3.2 and greater than this will be a problem.  We need to continue supporting users of Py 2.7.

smithsz: If `getattr(self._dict, 'move_to_end', None)` returns `None` then we drop through to the `else` and run the ported code.

alfinkel:
A general comment:  This PR seems incomplete to me or perhaps I'm just not understanding the scope of the PR.  Issue #67 came about due to the inconsistent way we handle caching in the the library both in the client and the database.  Based on your addition to the "Getting Started" guide it appears as though you're setting the caching notion at the client level but your implementation seems specific to the database class.

alfinkel:
The notion of the `LRUDict` raises some alerts for me.  As it is currently implemented it does not resolve to a `dict` which concerns me.  TBH I think this effort may be better served with a design specification before jumping into implementation.  One that identifies the overall strategy of caching for the library and the impact of changing its current behavior on current consumers of the python-cloudant library (for example our services API).  At first glance this may cause us (SAPI) issues internally.

smithsz:
Thanks for taking a look. That's a fair comment. I'll close this PR and consider opening a design specification for redesigning the caching implementation.

alfinkel:
Thanks.  Whenever you put together a spec I'd like to have a look if that's OK.  It probably makes sense to keep SAPI in the loop when making changes that affect current behavior so we can appropriately gauge impact.

Issue #142, opened 2016-04-08
create_database checks existence twice (3 RTTs)


ricellis:
I was doing some testing and noticed that the `create_database` call makes two `GET` requests before the `PUT` to create a remote database.
This is what happens:
1. `client.create_database()` calls `database.exists()` which does a `GET`
2. If it didn't exist `client.create_database()` calls `database.create()`
3. `database.create()` calls `exists()` which does another `GET`
4. If the database does not exist then  `database.create()` finally does the `PUT`.

I think it could be beneficial to replace these three round trips with a single request to `PUT` given that if the database already exists a `412` is returned.
There is some complication in that currently `database.create()` would not throw for the `412` response case, but `client.create_database` optionally needs to throw.
- Cloudant (python-cloudant) version(s) that are affected by this issue.
  - 2.0.0b2 (probably others too)
- Python version 
  - any (tested on 2.7)
- A small code sample that demonstrates the issue.

``` python
client.create_database('abc')
```

alfinkel:
These are now `HEAD` calls rather than `GET` fwiw but imo at least one of the `HEAD` calls should be removed.  See my comments on #279.

PR #285, opened 2017-03-07
142 existence twice check

porthunt:

- [ ] You have signed the CLA as per the instructions in [CONTRIBUTING.rst] (https://github.com/cloudant/python-cloudant/blob/master/CONTRIBUTING.rst#contributor-license-agreement) **_(not applicable, IBMer)._**

## What

In this current PR, the method `create_database()` on `Client.CouchDB` doesn't double check if the database already exists (`new_db.exists()`). 
## How

The method sends a `none_on_exists` keyword argument to `create()`. 

- If `none_on_exists` is `True`, it will return None (and it will raise the `CloudantClientException`).
- If `none_on_exists` is `False`, it will return the already existing database.

*The default `none_on_exists` is set to `True` if there are no `none_on_exists` keyword argument.*

## Testing

Find below the test:

```python
from cloudant import Cloudant
from cloudant.error import CloudantClientException
import unittest


class TestTwiceCheck(unittest.TestCase):

    database = 'foobar-test'

    # creating new database. Expecting {} of the new database.
    def test_create(self):
        self.assertEqual(client.create_database(TestTwiceCheck.database), {})

    # now that the database already exists,
    # it will raise the CludantClientException
    def test_exists(self):
        self.assertRaises(CloudantClientException,
                          client.create_database, TestTwiceCheck.database)

    # If we use throw_on_exists as False, it will return
    # the already created database. That way we won't
    # double check the exists() method.
    def test_exists_bypass(self):
        self.assertEqual(client.create_database(TestTwiceCheck.database,
                         throw_on_exists=False), {})


if __name__ == '__main__':
    global client
    client = Cloudant('user', 'pwd', account='user')
    client.connect()

    unittest.main(exit=False)
    # deleting the database so we can run the test again without
    # raising the CloudantClientException (database already exists.)
    client.delete_database(TestTwiceCheck.database)
```

## Issues
Related to issue #142.

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -329,16 +329,21 @@ class CouchDatabase(dict):
         else:
             return view.result
 
-    def create(self):
+    def create(self, **kwargs):
         """
         Creates a database defined by the current database object, if it
         does not already exist and raises a CloudantException if the operation
         fails.  If the database already exists then this method call is a no-op.
 
+        :param bool none_on_exists: Boolean flag dictating whether or
+            not to return None object when the database already exists.
         :returns: The database object
         """
         if self.exists():
-            return self
+            if kwargs.get('none_on_exists', True):
+                return None
+            else:
+                return self
 
         resp = self.r_session.put(self.database_url)
         if resp.status_code == 201 or resp.status_code == 202:
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -224,10 +224,8 @@ class CouchDB(dict):
         :returns: The newly created database object
         """
         new_db = self._DATABASE_CLASS(self, dbname)
-        if new_db.exists():
-            if kwargs.get('throw_on_exists', True):
-                raise CloudantClientException(409, dbname)
-        new_db.create()
+        if new_db.create(none_on_exists=kwargs.get('throw_on_exists', True)) is None:
+            raise CloudantClientException(409, dbname)
         super(CouchDB, self).__setitem__(dbname, new_db)
         return new_db
 
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -224,8 +224,8 @@ class CouchDB(dict):
         :returns: The newly created database object
         """
         new_db = self._DATABASE_CLASS(self, dbname)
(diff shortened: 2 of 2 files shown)


porthunt:
I'll review the `database_tests.py` to reflect the new changes.

emlaver:
@porthunt I've sent you an e-mail about getting approval from your manager to contribute to this library.  Also, it would be better to have this test in `tests/unit/client_tests.py`.  Thanks for your work.

porthunt:
@emlaver Sure, she will send the email today. I will fix the `database-tests.py` and add the test to `client_tests.py`. Thanks!

src/cloudant/database.py
```diff
@@ -329,16 +329,21 @@ def get_view_result(self, ddoc_id, view_name, raw_result=False, **kwargs):
...
+        :param bool none_on_exists: Boolean flag dictating whether or
+            not to return none object when the database already exists.
         :returns: The database object
         """
         if self.exists():
-            return self
+            if kwargs.get('none_on_exists', True):
+                return None
+            else:
+                return self
```
alfinkel: I think that passing back `None` if a database exists is an odd behavior to introduce to fix this bug.  Instead I propose that you add `throw_on_exists` to the signature of the database `create` method and default it to `False` so as not to change the current default behavior of this method call.  Like `def create(self, throw_on_exists=False):`.  Then I would change the `if self.exists():` check to `if not throw_on_exists and self.exists():`.  Again preserving the current behavior of the method.  With the logic changed to this the method can now throw a CloudantDatabaseException 409 which we can then specifically catch in the client `create_database` method and re-wrap as a CloudantClientException.  Once again preserving the current behavior of that `create_database` method.  Obviously we will also need to change the client `create_database` logic to call `new_db.create(kwargs.get('throw_on_exists', False))` and remove the `new_db.exists()` if logic dance in order to complete the fix.

This allows us to:
- Preserve current client and database object functionality which means a micro release.
- Remove the extra call to `exists()` bug.

porthunt: Seems a good idea to me. I will rewrite and submit it again. Thanks for the idea.

PR #286, opened 2017-03-09
142 Fixing database existance double check

porthunt:

## What

Removed the double check on the `exists()` method related to issue #142 .

## How

The method `create_database()` already uses a `throw_on_exists` to throw a `CloudantClientException` if the flag is `True`. The solution was to send this flag to `create()` and then changing this method to `if not throw_on_exists and self.exists()` instead of `if self.exists()`. That way, if the flag is `False`, it will not raise the exception, but if its `True`, it will raise a `CloudantDatabaseException`, that will be handled by the `create_database()` method and raise a `CloudantClientException`.

## Testing

`test_create_delete_database()` on `client_test.py` is already handling this change. Also, I have made an update on `database_test.py` to raise the exception if the database is already created. Now the method `test_create_db_delete_db()` is testing not only `create()`, but also `create(throw_on_exists=True)`.

## Issues

#142


porthunt:
My manager took a training session yesterday and couldn't send the email. I will talk to her tomorrow. I will appreciate your feedback.

src/cloudant/client.py
```diff
@@ -224,10 +227,11 @@ def create_database(self, dbname, **kwargs):
...
         """
         new_db = self._DATABASE_CLASS(self, dbname)
-        if new_db.exists():
-            if kwargs.get('throw_on_exists', True):
-                raise CloudantClientException(409, dbname)
-        new_db.create()
+        try:
+            new_db.create(kwargs.get('throw_on_exists', False))
+        except CloudantDatabaseException:
+            raise CloudantClientException(409, dbname)
```
alfinkel: You need to specifically catch the 409 `CloudantDatabaseException` here and re-raise as a `CloudantClientException` 409.  Anything other than a 409 should just re-raise as the original `CloudantDatabaseException`.  An if/else block here checking for the 409 status code here would do the trick.

tests/unit/database_tests.py
```diff
@@ -202,6 +202,11 @@ def test_create_db_delete_db(self):
             # No issue should arise if attempting to create existing database
             db_2 = db.create()
             self.assertEqual(db, db_2)
+            # If we use throw_on_exists=True, it will arise a
```
alfinkel: raise not arise

tests/unit/database_tests.py
```diff
@@ -202,6 +202,11 @@ def test_create_db_delete_db(self):
...
+            # CloudantDatabaseException if the database already exists.
+            with self.assertRaises(CloudantDatabaseException) as cm:
+                db.create(throw_on_exists=True)
+            self.assertEqual(cm.exception.status_code, 100)
```
alfinkel: This is troubling, I would think that this should return a 409.  Why do you think it should return a 100?  The troubling part is the fact that the test appears to pass.

porthunt:
Fixed the `arise` typo and the identation issue that Travis got. Will work on the exception handling tonight.

tests/unit/database_tests.py
porthunt: It is a `CloudantDatabaseException`. Isn't the `409` code for `CloudantClientException`?

alfinkel: The `CloudantDatabaseException` should basically take the original response status code and use it as its own code.  The goal here was not to mask the actual error code.  So my assumption here was that a 409 would be returned by the `raise_for_status` here: https://github.com/porthunt/python-cloudant/blob/f823d1f68e3431ee63f9f6f4cd9d93a0bf3e39e0/src/cloudant/database.py#L351-L353.  That 409 would then be re-raised as a `CloudantDatabaseException` preserving the original HTTP response code and then re-raised again later as a `CloudantClientException` while still preserving the original HTTP response code.  It appears that the 409 assumption was incorrect?  If so then we have a different problem to deal with.

porthunt: I see. The point is: `self.r_session.put(self.database_url)` is not returning a 409 error, but 412: 
a

Since there are no 412 code on `_messages.py`, it is creating a `CloudantDatabaseException` with code 100. We can handle this creating a new 412 error on `DATABASES = { ... }` (_messages.py). 

I don't think a 409 would resolve the issue since this code is `Document with id _ already exists.`, and that's not the issue we should have, correct?

emlaver: I've opened #287 since the exception is 412 when a database already exists.  This exception happens [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L343) and we should be preserving this status code when raised by `CloudantDatabaseException` and `CloudantClientException`.

alfinkel: So there we go @porthunt, looks like your new test inadvertently caught us a bug.

src/cloudant/client.py
porthunt: Now that the issue #287 was created, should we continue with this pull request, catching the 100 code, or tru to work first on the new issue?

alfinkel: I think we should wait for that issue to be resolved.  We also still need your manager's email before we can do anything with this PR anyway.

emlaver: #287 PR merged and issued resolved.

porthunt: Okay, I will redo the tests tonight and fix the catch to 412. Also, my manager sent the email.

porthunt:
Done. Do you want me to squash 85dd18f too? I would be a committer to some other commits:

image

PR #288, opened 2017-03-09
287 throw exception with 412 code when db exists 

### LINKED ISSUE #287: Cloudant exception should throw 412 when database already exists
Currently we are throwing a `CloudantClientException` with 409 status code when it should be 412: https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/client.py#L229. 
This can be confirmed by trying to create an existing database and checking the status code for `resp`: https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L343.

emlaver:
## What

Replaced the existing exception error code of 409 with 412 when the database already exists.

## How

- Updated error code to 412 in `_messages` and `client.create_database`

## Testing

No new tests.
Updated existing `test_create_existing_database` test to assert the correct 412 exception.

## Issues

fixes #287

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 2.5.0 (Unreleased)
 ==================
 - [FIXED] Fixed ``TypeError`` when setting revision limits on Python>=3.6.
+- [FIXED] Fixed Cloudant exception code 409 with 412 when creating a database that already exists.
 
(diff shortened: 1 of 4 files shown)


alfinkel:
Probably want to update the change log because this is the type of change we want to be sure that we remember to capture.

Issue #284, opened 2017-02-27
Add support for _local document management


alfinkel:
Add support for the manipulation of `_local` documents.

alfinkel:
I began work on this but soon came to the realization that `_local` document support needs to be handled differently between CouchDB and Cloudant.  I closed the PR #290 but it can be reopened later when we have some additional time to work out all of the differences and come up with a clean homogeneous solution for the handling of `_local` documents.

ricellis:
Won't fix here.

Our new [cloudant-python-sdk](https://github.com/IBM/cloudant-python-sdk/)(beta) can do [local docs](https://cloud.ibm.com/apidocs/cloudant?code=python#putlocaldocument).

PR #290, opened 2017-03-31
Add support for Cloudant _local documents

alfinkel:
## What

Added the LocalDocument class to provide support for Cloudant `_local` documents

## How

- Added a LocalDocument class that extends a `dict` and provides `exists`, `create`, `fetch`, `save`, `delete`, and context manager functionality.
- Added docs as well

## Testing

- Added tests that validate new local document functionality.  The majority of the tests target Cloudant only.
- Added tests to valdate the new local document exception class functionality.

## Issues

#284

### DIFF
--- a/tests/unit/local_document_tests.py
+++ b/tests/unit/local_document_tests.py
@@ -1,0 +1,388 @@
+#!/usr/bin/env python
+# Copyright (c) 2017 IBM. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+local document module - Unit tests for the LocalDocument class
+
+See configuration options for environment variables in unit_t_db_base
+module docstring.
+"""
+
+import unittest
+import mock
+import json
+import requests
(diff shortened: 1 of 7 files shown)


tests/unit/local_document_tests.py
```diff
@@ -0,0 +1,388 @@
...
+        self.assertEqual(
+            doc.document_url, '/'.join(
+                [self.db.database_url, '_local', 'http%3A%2F%2Fexample.com']
+            )
+        )
+    
+    @unittest.skipUnless(
+        os.environ.get('RUN_CLOUDANT_TESTS') is not None, 'Skipping Cloudant test'
+    )
+    def test_exists(self):
```
emlaver: What about putting the Cloudant tests into their own class like we have in `database_tests.py`?

tests/unit/local_document_tests.py
```diff
@@ -0,0 +1,388 @@
...
+        doc.fetch()
+        self.assertEqual(
+            doc,
+            {'_id': '_local/julia006', '_rev': '0-1', 'name': 'julia', 'age': 6}
+        )
+
+    @unittest.skipUnless(
+        os.environ.get('RUN_CLOUDANT_TESTS') is not None, 'Skipping Cloudant test'
+    )
+    def test_fetch_not_found(self):
```
rhyshort: Doesn't this also apply to Couch?

alfinkel:
It looks like CouchDB 1.6 (not sure about CouchDB 2.0) does not match the [API specification](http://docs.couchdb.org/en/1.6.1/api/local.html#) whereas Cloudant appears to adhere to it.  This PR was written based on the spec and `_local` document functionality as it exists in Cloudant.  Since this library is meant to support both Cloudant and CouchDB I am closing this PR because I currently do not have the bandwidth to work out a viable `_local` document solution for both Cloudant and CouchDB.  I'll keep the branch around in case there is time later on to revisit this.

Issue #291, opened 2017-04-07
Database.create_document throw_on_exists bug


alfinkel:
[Line 163](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L163) will still throw an HTTPError if the `throw_on_exists` flag is set to `False` in [Database.create_dcoument](
https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L154-L165).

smithsz:
Merged fix.

PR #292, opened 2017-04-11
291 doc throw on exists

smithsz:
## What
Catch error if `throw_on_exists` flag is set to `False` for document create.

## How
Add new `throw_on_exists` boolean param to the `Document.create` method. Note that default behaviour is unchanged.

## Testing
Includes additional unit test `test_create_document_that_already_exists`.

## Issues
- #291

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -157,10 +157,15 @@ class CouchDatabase(dict):
             doc = DesignDocument(self, docid)
         else:
             doc = Document(self, docid)
-        if throw_on_exists and doc.exists():
-            raise CloudantDatabaseException(409, docid)
         doc.update(data)
-        doc.create()
+        try:
+            doc.create()
+        except HTTPError as error:
+            if error.response.status_code == 409:
+                if throw_on_exists:
+                    raise CloudantDatabaseException(409, docid)
+            else:
+                raise
         super(CouchDatabase, self).__setitem__(doc['_id'], doc)
         return doc
 
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -331,8 +331,8 @@ class CouchDatabase(dict):
             return view(**kwargs)
         elif kwargs:
             return Result(view, **kwargs)
-        else:
-            return view.result
+
+        return view.result
 
     def create(self, throw_on_exists=False):
         """
@@ -1237,8 +1237,8 @@ class CloudantDatabase(CouchDatabase):
             return query(**kwargs)
         if kwargs:
             return QueryResult(query, **kwargs)
-        else:
-            return query.result
+
+        return query.result
 
     def get_search_result(self, ddoc_id, index_name, **query_params):
         """
--- a/tests/unit/database_tests.py
+++ b/tests/unit/database_tests.py
@@ -267,6 +267,17 @@ class DatabaseTests(UnitTestDbBase):
                 'Document with id julia06 already exists.'
                 )
 
+    def test_create_document_that_already_exists(self):
+        """
+        Test creating a document that already exists
+        """
+        data = {'_id': 'julia'}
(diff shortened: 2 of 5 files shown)


src/cloudant/database.py
```diff
@@ -152,15 +152,12 @@ def create_document(self, data, throw_on_exists=False):
             corresponding to the new document in the database.
         """
         docid = data.get('_id', None)
-        doc = None
         if docid and docid.startswith('_design/'):
             doc = DesignDocument(self, docid)
         else:
             doc = Document(self, docid)
-        if throw_on_exists and doc.exists():
```
smithsz: I've removed the `.exists()` call (which adds an additional `HEAD` request) in favour of just catching the `409` on create. Does this sound reasonable?

src/cloudant/database.py
```diff
@@ -152,15 +152,12 @@ def create_document(self, data, throw_on_exists=False):
...
-            raise CloudantDatabaseException(409, docid)
         doc.update(data)
-        doc.create()
+        doc.create(throw_on_exists=throw_on_exists)
         super(CouchDatabase, self).__setitem__(doc['_id'], doc)
```
smithsz: Noting that `throw_on_exists=False` could result in your local database dict not being kept in sync with the server.

alfinkel: I don't know that pushing the throw_on_exists functionality down to the Document.create is the right way to go.  For one thing throwing a `CloudantDatabaseException` from there does change current functionality somewhat in that if a user of the library has code currently written using `create` and is expecting an HTTPError then they would have to refactor their code due to this change.  Wouldn't it just be easier to do this here:

```
try:
    doc.create()
except HTTPError as error:
    if error.response.status_code == 409:
        if throw_on_exist:
            raise
    else:
        raise
```

That way the `throw_on_exist` remains where it was intended to be, in the database object.

tests/unit/database_tests.py
```diff
@@ -267,6 +267,21 @@ def test_create_document_with_id(self):
                 'Document with id julia06 already exists.'
                 )
 
+    def test_create_document_that_already_exists(self):
+        """
+        Test creating a document that already exists
+        """
+        data = {'_id': 'jane001'}
```
alfinkel: Who is Jane? :)

src/cloudant/database.py
smithsz: okay - I'll implement this. thanks.
we were throwing a `CloudantDatabaseException` before back on L161. perhaps I should leave that initial HEAD check in too as to not change current functionality.

alfinkel: Good point about the `CloudantDatabaseException`, I guess in my suggestion above rather than re-raising if throw_on_exist is True instead raise the `CloudantDatabaseException` 409 while everything else remains the same.  I'm not sure that I follow your poiont on the HEAD check though.

src/cloudant/database.py
```diff
@@ -160,7 +160,14 @@ def create_document(self, data, throw_on_exists=False):
         if throw_on_exists and doc.exists():
             raise CloudantDatabaseException(409, docid)
         doc.update(data)
-        doc.create()
+        try:
+            doc.create()
+        except HTTPError as error:
+            if error.response.status_code == 409:
+                if throw_on_exists:
+                    raise
```
alfinkel: I thought that you would remove lines 160 and 161 and replace this line 168 with a `raise CloudantDatabaseException(409, docid)`.  Shouldn't that really do the trick to handle both conditions here?

alfinkel:
It appears as though pylint is angry over code not touched in this PR.  Could the pylint version have been updated and is now causing these warnings resulting in the CI failure?  Anyway, once pylint is happy again and CI is able to pass, everything else looks good.

Issue #289, opened 2017-03-24
all_docs call with keys=[] returns all results


Sampanna01S:
- Cloudant (python-cloudant) version(s) that are affected by this issue.
   2.3.0
- Python version
N/A
- It looks like using the `keys` parameter in view requests does not work as expected when `keys` is the empty list.  The expectation would be that no rows of data is returned but the current behavior is that all rows of data are returned.

alfinkel:
I believe that all that needs to be done here is to fix this line of code: https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/_common_util.py#L245 with `if keys is not None:` and add a test to validate the fix.

smithsz:
Merged fix.

PR #293, opened 2017-04-11
Fix _all_docs call where keys is an empty list

smithsz:
## What
Fix `/_all_docs` call where `keys=[]`.

## How
The `get_docs` common function was incorrectly dropping the `keys` parameter if the value of `keys` was an empty list.

## Testing

## Issues
- #289

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 - [FIXED] Fixed ``TypeError`` when setting revision limits on Python>=3.6.
 - [FIXED] Fixed the ``exists()`` double check on ``client.py`` and ``database.py``.
 - [FIXED] Fixed Cloudant exception code 409 with 412 when creating a database that already exists.
+- [FIXED] Catch error if ``throw_on_exists`` flag is ``False`` for document create.
+- [FIXED] Fixed /_all_docs call where ``keys`` is an empty list.
 
 2.4.0 (2017-02-14)
 ==================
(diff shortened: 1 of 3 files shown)


tests/unit/common_util_tests.py
```diff
@@ -0,0 +1,86 @@
...
+        resp = util.get_docs(m_sess, url, keys=keys)
+
+        m_post.assert_called_once_with(
+            url,
+            data='{"keys": []}',
+            headers={'Content-Type': 'application/json'},
+            params={}
+        )
+
+        self.assertEquals(200, resp.status_code)
```
alfinkel: Why would you choose to write mocked tests when we've been writing tests using couchdb tests all along?  Just thinking that it makes the library appear cleaner if our tests are structured and written consistently.  I feel like if you're going to start a CommonUtilsTests module you'll need to provide tests for all of the functions in the `_common_utils` module which obviously would be out of scope for this PR.  Perhaps remove these tests and just add test(s) targetting couchdb as part of the Database tests that ensures this `get_doc` method is working?  I open to other suggestions, I just think that right now these tests are a bit misaligned with the rest of our test suite in this library.

smithsz: righto, I'll add another all_docs test in database_tests using an empty key list and remove these.

alfinkel:
Looks like the CI should right itself once #292 is merged into master and this PR is rebased based on that.  Obviously my +1 is contingent on the CI passing.

PR #305, opened 2017-06-28
304 fix bad startkey in all doc iter

### LINKED ISSUE #304: Fix bad startkey in all docs iterator
Fix `next_startkey = '0'` at https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/database.py#L633.

This should either be `u'\u0000'` or the first `/_all_docs` request should not include a `startkey` parameter.

smithsz:
## What
Fix bad `startkey` parameter in `/_all_docs` iterator.

## How
Initially use `u'\u0000'` as a `startkey` parameter. With each batch, update the `startkey` to `docs[-1]['id'] + u'\u0000'`. 

This means that fetching an additional document per batch to act as the next `startkey` is now unnecessary.

## Testing
Includes additional `/_all_docs` test.

## Issues
Fixes #304.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@
 - [FIXED] Fixed Cloudant exception code 409 with 412 when creating a database that already exists.
 - [FIXED] Catch error if ``throw_on_exists`` flag is ``False`` for document create.
 - [FIXED] Fixed /_all_docs call where ``keys`` is an empty list.
+- [FIXED] Issue where docs with IDs that sorted lower than 0 were not returned when iterating through _all_docs.
 
 2.4.0 (2017-02-14)
 ==================
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -630,18 +630,20 @@ class CouchDatabase(dict):
         if not remote:
             super(CouchDatabase, self).__iter__()
         else:
-            next_startkey = '0'
+            # Use unicode Null U+0000 as the initial lower bound to ensure any
+            # document id could exist in the results set.
+            next_startkey = u'\u0000'
             while next_startkey is not None:
                 docs = self.all_docs(
-                    limit=self._fetch_limit + 1,  # Get one extra doc
-                                                  # to use as
-                                                  # next_startkey
+                    limit=self._fetch_limit,
                     include_docs=True,
                     startkey=next_startkey
                 ).get('rows', [])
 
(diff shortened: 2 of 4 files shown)


src/cloudant/database.py
```diff
@@ -630,18 +630,16 @@ def __iter__(self, remote=True):
         if not remote:
             super(CouchDatabase, self).__iter__()
         else:
-            next_startkey = '0'
+            next_startkey = u'\u0000'
```
ricellis: I think this is worthy of a comment, it is not obvious what the purpose is to someone not intimately familiar with doc sorting.

src/cloudant/database.py
```diff
@@ -630,18 +630,16 @@ def __iter__(self, remote=True):
...
-                                                  # next_startkey
+                    limit=self._fetch_limit,
                     include_docs=True,
                     startkey=next_startkey
                 ).get('rows', [])
 
-                if len(docs) > self._fetch_limit:
-                    next_startkey = docs.pop()['id']
+                if len(docs) >= self._fetch_limit:
+                    next_startkey = docs[-1]['id'] + u'\u0000'
```
ricellis: Again I think a comment here would be helpful.

src/cloudant/database.py
smithsz: Fixed in 84aabf8.

src/cloudant/database.py
smithsz: Fixed in 84aabf8.

CHANGES.rst
```diff
@@ -6,6 +6,7 @@
 - [FIXED] Fixed Cloudant exception code 409 with 412 when creating a database that already exists.
 - [FIXED] Catch error if ``throw_on_exists`` flag is ``False`` for document create.
 - [FIXED] Fixed /_all_docs call where ``keys`` is an empty list.
+- [FIXED] Use unicode Null U+0000 as /_all_docs iterator ``startkey`` parameter.
```
ricellis: This is how you fixed it, but it would be better for users to know what problem was fixed. Something like:
>[FIXED] Issue where docs with IDs that sorted lower than 0 were not returned when iterating through _all_docs

ricellis:
@smithsz Approving, but please reword the `CHANGES` before merging as per new comment.

smithsz:
FYI This fix is included in the 2.5.0 release.

PR #306, opened 2017-06-28
303 support iam auth

### LINKED ISSUE #303: Support IAM integration
Cloudant client libraries need to be able to exchange an IAM API key for an IAM access token. The user of the client library provides the API key. The access token is obtained by calling an IAM endpoint. The access token is time limited, like a cookie, so refresh logic also needs writing.

smithsz:
## What
Add IAM authentication support.

## How
Added a class method and context manager to simplify IAM client construction:
```python
client = Cloudant.iam('account_name', 'api_key_here', connect=True)
print client.all_dbs()

# or using context manager...
from cloudant import cloudant_iam
with cloudant_iam('account_name', 'api_key_here') as client:
    print client.all_dbs()
```

Under the hood there's a new `use_iam=True` toggle. Once enabled, the `password` parameter is used as the IAM API key.

__IAM Authentication Workflow:__

1. Exchange IAM API key for access token via `IAM_TOKEN_URL`.
2. Generate a cookie from the access token using the `POST /_iam_session` Cloudant endpoint.
3. Use the _IAMSession_ cookie in all library requests.
4. Attempt to renew if the _IAMSession_ cookie has expired or on any `401` response.

## Testing
Includes additional mock tests.

## Issues
Fixes #303.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -100,6 +100,10 @@ Cloud Platform.
 See `IBM Cloud Identity and Access Management <https://console.bluemix.net/docs/services/Cloudant/guides/iam.html#ibm-cloud-identity-and-access-management>`_
 for more information.
 
+The production IAM token service at *https://iam.bluemix.net/oidc/token* is used
+by default. You can set an ``IAM_TOKEN_URL`` environment variable to override
+this.
+
 You can easily connect to your Cloudant account using an IAM API key:
 
 .. code-block:: python
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -88,6 +88,26 @@ following statements hold true:
                      connect=True,
                      auto_renew=True)
 
+
+************************************
+Identity and Access Management (IAM)
+************************************
+
+IBM Cloud Identity & Access Management enables you to securely authenticate
+users and control access to all cloud resources consistently in the IBM Bluemix
+Cloud Platform.
+
+See `IBM Cloud Identity and Access Management <https://console.bluemix.net/docs/services/Cloudant/guides/iam.html#ibm-cloud-identity-and-access-management>`_
+for more information.
+
+You can easily connect to your Cloudant account using an IAM API key:
+
+.. code-block:: python
+
+    # Authenticate using an IAM API key
+    client = Cloudant.iam(ACCOUNT_NAME, API_KEY, connect=True)
+
+
 ****************
 Resource sharing
 ****************
--- a/src/cloudant/__init__.py
+++ b/src/cloudant/__init__.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015 IBM. All rights reserved.
+# Copyright (c) 2015, 2017 IBM. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/cloudant/__init__.py
+++ b/src/cloudant/__init__.py
@@ -63,6 +63,21 @@ def cloudant(user, passwd, **kwargs):
     cloudant_session.disconnect()
 
 @contextlib.contextmanager
+def cloudant_iam(api_key, account_name, **kwargs):
+    """
+    Provides a context manager to create a Cloudant session and provide access
+    to databases, docs etc.
+
+    :param api_key: IAM authentication API key.
+    :param account_name: Cloudant account name.
+    """
+    cloudant_session = Cloudant(account_name, api_key, use_iam=True, **kwargs)
+
+    cloudant_session.connect()
+    yield cloudant_session
+    cloudant_session.disconnect()
+
+@contextlib.contextmanager
 def cloudant_bluemix(vcap_services, instance_name=None, **kwargs):
     """
     Provides a context manager to create a Cloudant session and provide access
--- a/src/cloudant/__init__.py
+++ b/src/cloudant/__init__.py
@@ -63,15 +63,29 @@ def cloudant(user, passwd, **kwargs):
     cloudant_session.disconnect()
 
 @contextlib.contextmanager
-def cloudant_iam(api_key, account_name, **kwargs):
+def cloudant_iam(account_name, api_key, **kwargs):
     """
-    Provides a context manager to create a Cloudant session and provide access
-    to databases, docs etc.
+    Provides a context manager to create a Cloudant session using IAM
+    authentication and provide access to databases, docs etc.
 
-    :param api_key: IAM authentication API key.
     :param account_name: Cloudant account name.
+    :param api_key: IAM authentication API key.
+
+    For example:
+
+    .. code-block:: python
+
+        # cloudant context manager
+        from cloudant import cloudant_iam
+
+        with cloudant_iam(ACCOUNT_NAME, API_KEY) as client:
+            # Context handles connect() and disconnect() for you.
+            # Perform library operations within this context.  Such as:
+            print client.all_dbs()
+            # ...
+
     """
-    cloudant_session = Cloudant(account_name, api_key, use_iam=True, **kwargs)
+    cloudant_session = Cloudant.iam(account_name, api_key, **kwargs)
 
     cloudant_session.connect()
     yield cloudant_session
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -24,8 +24,7 @@ from collections import Sequence
 import json
 from requests import RequestException, Session
 
-from ._2to3 import LONGTYPE, STRTYPE, NONETYPE, UNITYPE, iteritems_, url_parse, \
-    url_join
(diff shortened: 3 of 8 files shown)


src/cloudant/__init__.py
```diff
@@ -63,6 +63,21 @@ def cloudant(user, passwd, **kwargs):
     cloudant_session.disconnect()
 
 @contextlib.contextmanager
+def cloudant_iam(api_key, account_name, **kwargs):
+    """
+    Provides a context manager to create a Cloudant session and provide access
+    to databases, docs etc.
+
+    :param api_key: IAM authentication API key.
+    :param account_name: Cloudant account name.
```
emlaver: Can we add an example similar to what we have below in `cloudant_bluemix`?

smithsz: Yep, will do. I'll improve the getting started stuff too once I find the proper IAM blumix docs link.

src/cloudant/_common_util.py
```diff
@@ -287,66 +290,187 @@ def __new__(cls, code):
             return str.__new__(cls, code.encode('utf8'))
         return str.__new__(cls, code)
 
-class InfiniteSession(Session):
+
+class ClientSession(Session):
```
ricellis: The majority of the work that `CookieSession` and `IAMSession` have to do is the same, other than the extra piece to get an IAM token (and the content-type of the `POST` to `*_session` which should be `application/json` for IAM).
It would be nice (but by no means essential if it is harder than it initially looks) if we could use the same login/logout/expired checks in this base class and just customize them slightly (e.g. with the url/data/response conditions).

docs/getting_started.rst
```diff
@@ -88,6 +88,28 @@ following statements hold true:
...
+
+You can easily connect to your Cloudant account using an IAM API key:
+
+.. code-block:: python
+
+    # Authenticate using an IAM API key
+    client = Cloudant(USERNAME, API_KEY, account=ACCOUNT_NAME,
+                     auto_renew=True,
+                     connect=True,
+                     use_iam=True)
```
ricellis: I'm not a fan of exposing yet another parameter on here, I think it is particularly confusing that the `USERNAME` is effectively ignored in this scenario. I wonder whether we can have another way of initializing that effectively hides this complexity.

tests/unit/iam_auth_tests.py
```diff
@@ -0,0 +1,157 @@
...
+from cloudant._common_util import IAMSession
+
+
+class IAMAuthTests(unittest.TestCase):
+    """ Unit tests for IAM authentication. """
+
+    @mock.patch('cloudant._common_util.ClientSession.request')
+    def test_iam_get_access_token(self, m_req):
+        m_response = mock.MagicMock()
+        m_response.json.return_value = {'access_token': 'bar'}
```
ricellis: This mock response should probably have more of the expected token content in it.

docs/getting_started.rst
smithsz: Added new class method in 9f181e78210416d3e3fdc3753edda7c31388f0d0.

tests/unit/iam_auth_tests.py
```diff
@@ -0,0 +1,262 @@
...
+        m_get.return_value = m_response
+
+        iam = IAMSession(MOCK_API_KEY, 'http://127.0.0.1:5984')
+        info = iam.info()
+
+        m_get.assert_called_once_with(iam._session_url)
+
+        self.assertEqual(info, m_info)
+        self.assertTrue(m_response.raise_for_status.called)
+
```
ricellis: I feel like there is a golden path test missing here, checking that a request is as expected without any expiry/renewal conditions.

I also think it would be valuable to have a test at the `client` level verifying that using the class method creates the IAM session and appropriate golden path is followed making a normal request (token, iam_session, normal request).

smithsz: Added in 85932a984931f258660d740b6b63a5ec16bdb5e3 & fc8d64a3fc93dbf79565b715aa97949f03d352b4.

docs/getting_started.rst
smithsz: Ah, forgot to update docs. Fixed in cd1c664aa4fa91f0991ec1c8b23579af04bc76fa.

src/cloudant/client.py
```diff
@@ -67,6 +68,8 @@ class CouchDB(dict):
         `Requests library timeout argument
         <http://docs.python-requests.org/en/master/user/quickstart/#timeouts>`_.
         but will apply to every request made using this client.
+    :param bool use_iam: Keyword argument, if set to True performs
```
smithsz: Added in 9d0a5f3616b2702adedb5061102fdbec341d2649.

src/cloudant/_common_util.py
smithsz: It's a little tricky as `ClientSession` is used directly when admin party is enabled.
I still think it's worth doing. I've raise issue #310 to track the work.

src/cloudant/_common_util.py
```diff
@@ -287,66 +289,182 @@ def __new__(cls, code):
...
-        self._password = password
-        self._server_url = server_url
-        self._timeout = kwargs.get('timeout', None)
+    def __init__(self, api_key, server_url, **kwargs):
+        super(IAMSession, self).__init__(**kwargs)
+        self._api_key = api_key
+        self._auto_renew = kwargs.get('auto_renew', False)
+        self._session_url = url_join(server_url, '_iam_session')
+        self._token_url = os.environ.get(
+            'IAM_TOKEN_URL', 'https://iam.bluemix.net/oidc/token')
```
mikerhodes: Can this be made a kwarg rather than an env var? Why is it an env var?

docs/getting_started.rst
```diff
@@ -88,6 +88,26 @@ following statements hold true:
...
+
+See `IBM Cloud Identity and Access Management <https://console.bluemix.net/docs/services/Cloudant/guides/iam.html#ibm-cloud-identity-and-access-management>`_
+for more information.
+
+You can easily connect to your Cloudant account using an IAM API key:
+
+.. code-block:: python
+
+    # Authenticate using an IAM API key
+    client = Cloudant.iam(ACCOUNT_NAME, API_KEY, connect=True)
```
mikerhodes: Do you want to mention here the environment variable that allows for different IAM servers? (Note I query below whether it should be an env var rather than a kwarg).

src/cloudant/_common_util.py
```diff
@@ -287,66 +289,182 @@ def __new__(cls, code):
...
         """
-        Overrides ``requests.Session.request`` to set the timeout.
+        Overrides ``requests.Session.request`` to renew the IAM cookie
+        and then retry the original request (if required).
         """
-        resp = super(ClientSession, self).request(
-            method, url, timeout=self._timeout, **kwargs)
+        self.cookies.clear_expired_cookies()
+        if self._auto_renew and 'IAMSession' not in self.cookies.keys():
+            self.login()
```
mikerhodes: I'd rather this be more explicitly about the expiry of the IAMSession cookie, it took me a bit to understand that's what was actually happening. Like:

```python
if self._auto_renew and 'IAMSession' not in self.cookies.keys() and not self.cookies['IAMSession'].is_expired():
    self.login()
```

src/cloudant/_common_util.py
ricellis: Our view was that it should basically never be necessary to change the token URL, but we thought there was value in providing an "override" primarily for the event that for some reason the expected URL was unavailable (although it also has value in testing).

I think an env var is more suitable for this purpose in that it doesn't require a code change if, for example, one needed to switch token server to a backup to temporarily workaround some kind of outage. That was the thinking anyway.

docs/getting_started.rst
smithsz: Added in 3b4edd9fdc5f743f57bfb6eff5bcd515c860282b.

src/cloudant/_common_util.py
smithsz: You can't get a cookie object from the jar without iterating over the entire jar. `self.cookies['IAMSession']` just gives your the cookie value as a string.
We'd need something horrible like:
```python
self.cookies._cookies['mikerhodes-iam-testing.cloudant.com']['/']['IAMSession'].is_expired()
```
I think sticking to ` self.cookies.clear_expired_cookies()` is the cleanest solution.

src/cloudant/_common_util.py
mikerhodes: Okay

src/cloudant/_common_util.py
mikerhodes: I guess add a comment to the effect of this being the simplest way to check whether we have an expired cookie so the next person doesn't have to figure it out by guesswork. A decent "why I'm doing it" comment rather than a "describe what the code does" comment.

docs/getting_started.rst
mikerhodes: ok

PR #314, opened 2017-08-04
Implement Cloudant.bluemix client class method

### LINKED ISSUE #313: Implement bluemix client constructor method
Other than the `cloudant_bluemix` CM, the only way to construct a bluemix client is as follows:
```python
service = CloudFoundryService(vcap_services, instance_name)
cloudant_session = Cloudant(
    service.username,
    service.password,
    url=service.url,
)
```
We should abstract this in a class method, something like this:
```python
cloudant_session = Cloudant.bluemix(os.getenv('VCAP_SERVICES'), 'instance-name')
```

smithsz:

## What
Implement `Cloudant.bluemix()` client class method. 

## Example Usage
```python
import os

c = Cloudant.bluemix(os.getenv('VCAP_SERVICES'), 'Cloudant NoSQL DB')
```

## Testing

## Issues
Fixes #313.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
 2.6.0 (Unreleased)
 ==================
+- [NEW] Added ``Cloudant.bluemix()`` class method to the Cloudant client allowing service credentials to be passed using the CloudFoundry VCAP_SERVICES environment variable.
 - [FIXED] Fixed client construction in ``cloudant_bluemix`` context manager.
 
 2.5.0 (2017-07-06)
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -31,7 +31,8 @@ from ._common_util import (
     USER_AGENT,
     append_response_error_content,
     InfiniteSession,
-    ClientSession)
+    ClientSession,
+    CloudFoundryService)
 
 
 class CouchDB(dict):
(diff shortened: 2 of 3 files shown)


tests/unit/client_tests.py
```diff
@@ -539,6 +539,38 @@ def test_constructor_with_account(self):
             'https://{0}.cloudant.com'.format(self.account)
             )
 
+    def test_bluemix_constructor(self):
+        """
+        Test instantiating a client object using a VCAP_SERVICES environment
+        variable.
+        """
+        instance_name = 'Cloudant NoSQL DB-lv'
+        vcap_services = {'cloudantNoSQLDB': [{
```
emlaver: I think it would be good to test with a JSON object that has multiple Cloudant instances to verify that `instance_name` is working correctly.

smithsz: Good thinking. I'll add an additional test. I'll also add a test for 1 service in VCAP without specifying `instance_name`.

`tests/unit/cloud_foundry_tests.py` is testing majority of this. But since I've already gaffed this once it'd be good to test again at the client instantiation level too.

CHANGES.rst
```diff
@@ -1,5 +1,6 @@
 2.6.0 (Unreleased)
 ==================
+- [NEW] Added ``Cloudant.bluemix()`` client class method for instantiating a Cloudant Bluemix client.
```
ricellis: I'm not sure a "Cloudant Bluemix client" is really a thing.
Instantiating a client from VCAP_SERVICES metadata?
Last time we went with:
> - [NEW] Added ``cloudant_bluemix`` method to the Cloudant client allowing service credentials to be passed using the CloudFoundry VCAP_SERVICES environment variable.

src/cloudant/client.py
```diff
@@ -31,7 +31,8 @@
     USER_AGENT,
     append_response_error_content,
     InfiniteSession,
-    ClientSession)
+    ClientSession,
```
ricellis: Copyright date on this file is a bit weird, but I think it's fixed in #306 anyway.

Issue #116, opened 2016-03-16
Add default remote property to client and database


alfinkel:
Add a property/attribute to the client account and database classes that dictates the remote refresh behavior for the represented object.  Currently the behavior is dictated by a `remote` argument on each  method that interacts with the remote database.

Propose:
- Add the private attribute `_remote` to the constructors of CouchDB (client/account) and CouchDatabase (database).
- Default the client/account `_remote` to `False`.
- Default the database `_remote' to take the value of the same property from the client/account.
- Add property getter/setter methods to both CouchDB and Couchdatabase.  For example:

``` python
@property
def remote(self):
    return self._remote

@remote.setter(self):
def  setter(self, val)
    self._remote = val
```

Add logic to the following methods to handle the new property

account methods affected:
- `keys`
- `__getitem__` - Currently there is no option here.  We should add one.
- `__delitem__`
- `__setitem__`
- `get`

database methods affected:
- `keys`
- `__iter__`
- `__getitem__` - Currently there is no option here.  We should add one.
- Currently no`get` method available.  It would be good to add one to keep things consistent with the account logic.

emlaver:
Moving to backlog and low priority for now.  Changing this behavior will be a significant change to the API and should be handled in a major release with other breaking API changes.

smithsz:
I think we should revisit this again. It's unclear which methods use locally cached data and which don't.
Setting a `remote` parameter on instantiation would be far simpler than dealing with it on every call imo.
Also, we should check which `dict` methods we aren't overriding to be sure that `remote=True`  is respected (i.e. the database `__contains__` only examines the local `dict`).

ricellis:
We're no longer doing enhancements here.

PR #320, opened 2017-08-14
Update Document CM usage in docs/getting_started.rst

### LINKED ISSUE #317: deleting within a context manager does not persist..
With a couchdb database db, when I do:
```
    with Document(db, doc_id) as db_doc:
      db_doc.delete()
      logging.debug(db[doc_id].exists())
    logging.debug(db[doc_id].exists())
```
the first `logging.debug(db[doc_id].exists())` prints False, while the second prints True. This is rather strange, and points to a bug either in the context manager or the delete documentation.

smithsz:

## What
Update `Document` CM usage in 'Getting Started' docs.

Add warning about using `.delete()` method inside CM.

## Issues
Fixes #317.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -342,27 +342,27 @@ Context managers
 Now that we've gone through the basics, let's take a look at how to simplify 
 the process of connection, database acquisition, and document management 
 through the use of Python *with* blocks and this library's context managers.  
-Handling your business using *with* blocks saves you from having to connect and 
+
+Handling your business using *with* blocks saves you from having to connect and
 disconnect your client as well as saves you from having to perform a lot of 
 fetch and save operations as the context managers handle these operations for 
-you.  This example uses the ``cloudant`` context helper to illustrate the 
+you.
+
+This example uses the ``cloudant`` context helper to illustrate the
 process but identical functionality exists for CouchDB through the ``couchdb`` 
 and ``couchdb_admin_party`` context helpers.
 
 .. code-block:: python
 
-    # cloudant context helper
     from cloudant import cloudant
 
-    # couchdb context helper
+    # ...or use CouchDB variant
     # from cloudant import couchdb
-
-    from cloudant.document import Document
 
     # Perform a connect upon entry and a disconnect upon exit of the block
     with cloudant(USERNAME, PASSWORD, account=ACCOUNT_NAME) as client:
 
-    # CouchDB variant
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -394,6 +407,10 @@ and ``couchdb_admin_party`` context helpers.
 
         print('Databases: {0}'.format(client.all_dbs()))
 
+.. warning:: Don't use the ``doc.delete()`` method inside your ``Document``
+             context manager. This method immediately deletes the document on
+             the server.
```
ricellis: The documentation for deleting a doc is pretty light http://python-cloudant.readthedocs.io/en/stable/getting_started.html#delete-a-document
I feel like if we're saying don't delete it this way then we ought to document an alternative deletion method. e.g. 
i) Does setting `_deleted: true` within the context work such that the save on leaving the context cause the update style delete? In which case that would be a workaround for the document context.
ii) Presumably calling `doc.delete` from a DB context works, so perhaps that is the context manager approach to deletion to document. Are there others at that level too  e.g. `db.pop('docid')`?

We should probably also extend this to say something like "and the document save when leaving the context will re-create an empty document with the document ID". Otherwise someone might be tempted to call `doc.delete()` in the document context anyway because they don't care if it happens immediately or not if you catch my meaning.

smithsz: Okay, let's advise deleting inside the CM is done via `doc['_deleted'] = True`. That seems to work well.

`doc.delete()` works inside the CM (assuming the doc already exists). Although on exiting the CM we always write a doc too. So using `.delete()` in the CM doesn't really make any sense.

`db.pop('docid')` only deals with the local database object afaict. Noting also that using the `Document` CM doesn't update the local cache.

ricellis: >Okay, let's advise deleting inside the CM is done via doc['_deleted'] = True. That seems to work well.

Cool, perhaps we should advise wiping the non `_id` `_rev` `_deleted` keys too or at least mention that for cases where people might not want to send a large body back again?

>doc.delete() works inside the CM (assuming the doc already exists). Although on exiting the CM we always write a doc too. So using .delete() in the CM doesn't really make any sense.

I thought there was a DB CM as well as a doc CM? In any case I'm happy to go with `_deleted` approach.

>Noting also that using the Document CM doesn't update the local cache.

We probably need a clear document about what exactly happens local vs remote. I guess that's just another part of #116.

PR #321, opened 2017-08-14
Add ``Result.all()`` convenience method

### LINKED ISSUE #233: Add all() convenience method
Retrieving the entire set of data from a `Result` object is awkward when using result slices.  To retrieve the full data set for a `Result` using a slice, you must do something like:

``` python
result = Result(...)
data = result[:]
```

...which is a bit awkward.  We should add an `all()` method call that performs the same functionality as `result[:]` so that we could do something like:

``` python
result = Result(...)
data = result.all()
```

smithsz:

## What
Add ``Result.all()`` convenience method.

## How
Performs same functionality as `result[:]`.

## Testing
Includes additional unit test.

## Issues
Fixes #233.

### DIFF
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -359,6 +359,18 @@ class Result(object):
         """
         return data.get('rows', [])
 
+    def all(self):
+        """
+        Retrieve all results.
+
+        Specifying a ``limit`` parameter in the ``Result`` constructor will
+        limit the number of documents returned. Be aware that the ``page_size``
+        parameter is not honoured.
+
+        :return: results data as list in JSON format.
+        """
+        return self[:]
+
 class QueryResult(Result):
     """
     Provides a index key accessible, sliceable and iterable interface to query
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,6 @@
 Unreleased
 ==========
+- [NEW] Added ``Result.all()`` convenience method.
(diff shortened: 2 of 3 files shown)


src/cloudant/result.py
```diff
@@ -359,6 +359,14 @@ def _parse_data(self, data):
         """
         return data.get('rows', [])
 
+    def all(self):
+        """
+        Retrieve the full results data set.
+
+        :return: results data set.
```
ricellis: Is it really a `set` as in the [built-in](https://docs.python.org/2/library/stdtypes.html#set)? We seem to refer to it as a collection elsewhere.

smithsz: good spot. no, it's a `list`. I'll update the docstring.

src/cloudant/result.py
```diff
@@ -359,6 +359,17 @@ def _parse_data(self, data):
         """
         return data.get('rows', [])
 
+    def all(self):
+        """
+        Retrieve all results.
+
+        Specifying a ``limit`` parameter in the ``Result`` constructor will
```
smithsz: The `page_size` parameter is not respected when accessing results via slice. This is already noted in the docs here:
>:param int page_size: Sets the page size for result _iteration_.

emlaver: @smithsz @ricellis Is it worth pointing to this in the docs or re-stating it?

smithsz: Done.

PR #322, opened 2017-08-14
Use `'/'.join()` to concatenate URL parts

### LINKED ISSUE #127: Convert all use of posixpath.join to string join
The library uses posixpath.join throughout to construct URLs, we should instead use string `join`.

As in:

``` python
'/'.join(['http://foo.bar.com', 'my_database'])
```

smithsz:

## What
Use `'/'.join()` to concatenate URL parts.

## How
Updated `posixpath.join` references to use `'/'.join` when concatenating URL parts.

## Issues
Fixes #127.

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -17,7 +17,6 @@ API module that maps to a Cloudant or CouchDB database instance.
 """
 import json
 import contextlib
-import posixpath
 
 from requests.exceptions import HTTPError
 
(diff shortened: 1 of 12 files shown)


src/cloudant/database.py
```diff
@@ -958,8 +955,8 @@ def security_url(self):
 
         :returns: Security document URL
         """
-        parts = ['_api', 'v2', 'db', self.database_name, '_security']
-        url = posixpath.join(self._database_host, *parts)
+        url = '/'.join((
+            self._database_host, '_api/v2/db', self.database_name, '_security'))
```
ricellis: This one example is different to all the others, why aren't `_api` `v2` and `db` separate parts here?

emlaver: As Rich mentioned, we should separate all parts of the path.

PR #324, opened 2017-08-30
Add a copy operation to the Document class

mikerhodes:

## What

Added support for (shallow) copying Document objects.

## How

Added a public `copy` method and private `__copy__` method. This means that a document can be copied via a call to the object itself and also via the `copy` module, that is `copy.copy(doc)`. The latter allows for calling code to be more generic.

## Testing

I added a test that both `copy` and `copy.copy` produce accurate copies and that they are distinct objects.

## Issues

N/A, minor change.

### DIFF
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -65,6 +65,21 @@ class Document(dict):
             self['_id'] = self._document_id
         self.encoder = self._client.encoder
 
+    def copy(self):
+        """
+        Create a copy of the Document, including any locally cached content.
+        """
+        return self.__copy__()
+
+    def __copy__(self):
+        """
+        Copy the document, for use in copy.copy() calls. See `copy()`.
+        """
+        cpy = Document(self._database, document_id=self._document_id)
+        for k, v in self.items():
+            cpy[k] = v
+        return cpy
+
     @property
     def r_session(self):
         """
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -15,6 +15,7 @@
 """
 API module/class for interacting with a document in a database.
 """
+import copy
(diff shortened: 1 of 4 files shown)


src/cloudant/document.py
```diff
@@ -65,6 +65,21 @@ def __init__(self, database, document_id=None):
...
+        """
+        return self.__copy__()
+
+    def __copy__(self):
+        """
+        Copy the document, for use in copy.copy() calls. See `copy()`.
+        """
+        cpy = Document(self._database, document_id=self._document_id)
+        for k, v in self.items():
+            cpy[k] = v
```
smithsz: Could `v` be a mutable type here? If so it'll get passed by reference and you won't get a deep copy.

mikerhodes: I don't intend for it to be a deep copy, only a shallow copy. See "What", "Added support for (shallow) copying Document objects." :) 

Whether this is the right intent we can discuss, but I took the normal dict copy behaviour.

smithsz:
Unfortunately this doesn't work for subclasses of `Document`:
```python
(Pdb) from cloudant.design_document import DesignDocument
(Pdb) ddoc1 = DesignDocument(self.db, "myddoc")
(Pdb) ddoc2 = copy.deepcopy(ddoc1)
(Pdb) ddoc2.__class__
<class 'cloudant.document.Document'>  # should be class DesignDocument
```
What's the issue with calling `copy.copy(mydoc)` or `copy.deepcopy(mydoc)` without specifying these custom implementations?

mikerhodes:
Seems like, as you say, `copy.copy` and `copy.deepcopy` work fine. Oh well.

PR #326, opened 2017-09-11
Test with an IAM authenticated client

smithsz:

## What
Test with an IAM authenticated client.

### DIFF
--- a/tests/unit/client_tests.py
+++ b/tests/unit/client_tests.py
@@ -38,7 +38,7 @@ from cloudant.client_session import BasicSession, CookieSession
 from cloudant.error import CloudantArgumentError, CloudantClientException
 from cloudant.feed import Feed, InfiniteFeed
 
-from .unit_t_db_base import UnitTestDbBase
+from .unit_t_db_base import skip_for_iam, UnitTestDbBase
 from .. import bytes_, str_
 
 class CloudantClientExceptionTests(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


tests/unit/client_tests.py
```diff
@@ -163,6 +163,7 @@ def test_multiple_connect(self):
             self.client.disconnect()
             self.assertIsNone(self.client.r_session)
 
+    @skip_for_iam
     def test_auto_renew_enabled(self):
```
emlaver: It looks like we are skipping all methods in this class for `iam` testing.  What if we skip the whole class instead?

tests/unit/replicator_tests.py
```diff
@@ -157,6 +157,7 @@ def test_replication_with_generated_id(self):
         clone = Replicator(self.client)
         clone.create_replication(self.db, self.target_db)
 
+    @skip_for_iam
     @flaky(max_runs=3)
     def test_create_replication(self):
```
emlaver: Why are we skipping IAM for replication?

smithsz: You need the user/pass for the account in order to place an auth header in the replication doc for the source db (since replication doesn't yet support IAM).

emlaver: Ok I understand, let's skip tests using _replicate endpoint.

tests/unit/client_tests.py
emlaver: Disregard the comment above, we are not skipping all tests in the class.

PR #328, opened 2017-09-18
Test with a basic access authenticated client

smithsz:

## What
Test with a basic access authenticated client.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,33 @@ def test_python(pythonVersion)
   }
 }
 
+def test_python_iam(pythonVersion)
+{
+  node {
+    // Unstash the source on this node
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -1,5 +1,35 @@
 // Define the test routine for different python versions
-def test_python(pythonVersion)
+
+def test_python_basic(pythonVersion)
```
tomblench: Can you share the shell stanzas between your `test_python_*` methods? There seems to be quite a bit of copy and paste.

smithsz: Yeah, sure. I'll follow the same approach as in [CouchBackup](https://github.com/cloudant/couchbackup/blob/master/Jenkinsfile#L16).

PR #335, opened 2017-10-20
Fix index tests following Mango maximum key value patch

smithsz:

## What
Following a patch to Mango the returned index JSON now includes additional
keys/values. To ensure the tests continue to pass on both old & new versions we
opt to inspect the returned JSON rather than assert on the object in its
entirety.

### DIFF
--- a/tests/unit/index_tests.py
+++ b/tests/unit/index_tests.py
@@ -159,25 +159,23 @@ class IndexTests(UnitTestDbBase):
         self.assertEqual(index.design_document_id, '_design/ddoc001')
         self.assertEqual(index.name, 'index001')
         with DesignDocument(self.db, index.design_document_id) as ddoc:
-            self.assertEqual(ddoc['language'], 'query')
-            self.assertListEqual(list(ddoc['views'].keys()), ['index001'])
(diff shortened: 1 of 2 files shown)


tests/unit/index_tests.py
```diff
@@ -457,27 +449,23 @@ def test_create_a_search_index_no_kwargs(self):
         self.assertEqual(index.design_document_id, '_design/ddoc001')
         self.assertEqual(index.name, 'index001')
         with DesignDocument(self.db, index.design_document_id) as ddoc:
-            self.assertEqual(ddoc['language'], 'query')
-            self.assertListEqual(list(ddoc['indexes'].keys()), ['index001'])
+            self.assertEquals(ddoc['_id'], '_design/ddoc001')
```
emlaver: You're asserting `ddoc['id']` with `_design/ddoc001` here but asserting `ddoc['id']` with `index.design_document_id` in other tests i.e. [line 248](https://github.com/cloudant/python-cloudant/pull/335/files#diff-315b6b51ac87a94f6e421411b64a3459R248).  We should use one or the other to keep this consistent throughout the tests.

smithsz: good spot. will update.

PR #340, opened 2017-11-30
[WIP] 337 fix creating replication with source and target strings

### LINKED ISSUE #337: Unable to create replications when specifying src/tgt databases as string
_Repro:_
```
client = Cloudant(
    cloudant_user='user123',
    auth_token='xxxxxxxxx',
    url='https://user123.cloudant.com',
    connect=True
)
replication_doc = Replicator(client).create_replication(
    source='src_db',
    target='tgt_db'
)
```
_Error:_
```
Traceback (most recent call last):
  File "bugtest.py", line 18, in <module>
    continuous=False
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/cloudant/replicator.py", line 96, in create_replication
    if not target_db.admin_party:
AttributeError: 'NoneType' object has no attribute 'admin_party'
```

I think a null check on `target_db` is required [here](https://github.com/cloudant/python-cloudant/blob/2.7.0/src/cloudant/replicator.py#L96) before calling `.admin_party`.

emlaver:
## What

- Fix the `null` error that occurs when creating a replicator with source and target strings.  
- Properly handle `repl_id` when only `source` and `target` parameters exist
- Decide on how `source` and `target` dictionaries should be passed in (see below)

## How

- Added [null check](https://github.com/cloudant/python-cloudant/compare/337-create-replication-with-src-trg-strings?expand=1#diff-941fe21666606963577a17ff664cdfe5R100) on `target_db` before calling `.admin_party`.  Also set 'user_ctx' if the database object's 'creds' exists.
- WIP: improve the logic to set `repl_id`.  When `source` and `target` parameters exist, the passed in `repl_id` value is set in `source_db`.  I've added logic to set `repl_id` = `source_db` if `source` and `target` exist and `source_db` is a string.

## Testing

- Added unit tests to verify that source and target string parameters work as expected
- If source or target url do not contain credentials, replication_state should equal `error` since the user is unauthorized

## Issues

fixes #337

~~TODO: Remove replicator doc that currently exists after tests finish.~~

TODO: How should `source` and `target` dictionaries be passed in?
If this below is the expected way, then this will also require fixing as it's not properly handled:
```
replication_doc = Replicator(client).create_replication(
    {'source': 'src_db', 'target': 'tgt_db'}
)
```

### DIFF
--- a/src/cloudant/replicator.py
+++ b/src/cloudant/replicator.py
@@ -69,6 +69,10 @@ class Replicator(object):
         :returns: Replication document as a Document instance
         """
 
+        # WIP: Handle 'repl_id' when 'source' and 'target' parameters exist
+        if not repl_id and source_db and isinstance(source_db, str):
+            repl_id = source_db
+
         data = dict(
             _id=repl_id if repl_id else str(uuid.uuid4()),
             **kwargs
@@ -93,7 +97,8 @@ class Replicator(object):
                 )
 
         if not data.get('user_ctx'):
-            if not target_db.admin_party:
+            if (target_db and not target_db.admin_party or
+                    self.database.creds and self.database.creds.get('user_ctx')):
                 data['user_ctx'] = self.database.creds['user_ctx']
 
         return self.database.create_document(data, throw_on_exists=True)
--- a/tests/unit/replicator_tests.py
+++ b/tests/unit/replicator_tests.py
@@ -155,7 +155,11 @@ class ReplicatorTests(UnitTestDbBase):
 
     def test_replication_with_generated_id(self):
         clone = Replicator(self.client)
-        clone.create_replication(self.db, self.target_db)
+        repl_id = clone.create_replication(
+            self.db,
+            self.target_db
+        )
+        repl_id.delete()
 
     @skip_if_not_cookie_auth
     @flaky(max_runs=3)
--- a/tests/unit/replicator_tests.py
+++ b/tests/unit/replicator_tests.py
@@ -21,7 +21,7 @@ See configuration options for environment variables in unit_t_db_base
 module docstring.
 
 """
-
+import os
 import unittest
 import uuid
 import time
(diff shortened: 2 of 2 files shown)


src/cloudant/replicator.py
```diff
@@ -69,6 +69,10 @@ def create_replication(self, source_db=None, target_db=None,
         :returns: Replication document as a Document instance
         """
 
+        # WIP: Handle 'repl_id' when 'source' and 'target' parameters exist
+        if not repl_id and source_db and isinstance(source_db, str):
```
smithsz: do you mean `source` here (not `source_db`)?  `source_db` shouldn't be a string according to the docs.

> source_db: Database object to replicate from.  Can be either a ``CouchDatabase`` or ``CloudantDatabase`` instance.

> source: Optional ``str`` or ``dict``

smithsz: also, why are we setting `repl_id = source_db`? is it just a new default?

src/cloudant/replicator.py
```diff
@@ -93,7 +97,8 @@ def create_replication(self, source_db=None, target_db=None,
                 )
 
         if not data.get('user_ctx'):
-            if not target_db.admin_party:
+            if (target_db and not target_db.admin_party or
+                    self.database.creds and self.database.creds.get('user_ctx')):
```
smithsz: nice, good to check existence of `user_ctx` too.

src/cloudant/replicator.py
emlaver: @smithsz `source_db` should be a Database object, but because of duck typing it will accept a `str`.
Without this code, if you pass in your own replication id it will set to the `source_db` parameter:
```
self.replicator.create_replication(
            repl_id,
            source=self.db.database_url,
            target=self.target_db.database_url)
```

src/cloudant/replicator.py
smithsz: perhaps this is slightly clearer?
```
        if not data.get('user_ctx'):
            if target_db and not target_db.admin_party:
                if self.database.creds and self.database.creds.get('user_ctx'):
                    data['user_ctx'] = self.database.creds['user_ctx']
```

emlaver:
@smithsz @ricellis from our slack discussion, we will be removing the `source` and `target` optional parameters.

PR #342, opened 2017-12-01
Removed source and target optional parameters from replicator API

### LINKED ISSUE #341: Remove src/trg optional parameters in replicator API
Building a replicator with `source` and `target` parameters does not work and has a bug (referenced in #337) that's existed since 2.0.0.  Since having both `source/target` & `source_db/target_db` params can be confusing and `source/target` currently don't work, we will remove these optional params.

emlaver:
## What

Removed source/target optional parameters from `create_replication`.

## How

- Remove constructor parameters and docstring

## Testing

No new tests.
Removed reference of `source/target` parameters in `integration/replicator_test.py`.
Deleted replication document in `test_replication_with_generated_id` after test completed.

## Issues

fixes #341

### DIFF
--- a/tests/integration/replicator_test.py
+++ b/tests/integration/replicator_test.py
@@ -190,57 +190,6 @@ class ReplicatorTest(unittest.TestCase):
             self.assertTrue(len(updates) > 0)
             self.assertEqual(updates[-1]['_replication_state'], 'completed')
 
-    @unittest.skip("Doesn't reliably get into error state on couch side.")
-    def test_follow_replication_with_errors(self):
-        """
-        _test_follow_replication_with_errors_
-
-        Test to make sure that we exit the follow loop when we submit
-        a bad replication.
-
-        """
-        dbsource = unicode_("test_follow_replication_source_error_{}".format(
-            unicode_(uuid.uuid4())))
-        dbtarget = unicode_("test_follow_replication_target_error_{}".format(
-            unicode_(uuid.uuid4())))
-
-        self.dbs = [dbsource, dbtarget]
-
-        with cloudant(self.user, self.passwd, account=self.user) as c:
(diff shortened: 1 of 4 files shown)


tests/integration/replicator_test.py
```diff
@@ -228,9 +228,6 @@ def test_follow_replication_with_errors(self):
             ret = replicator.create_replication(
                 source_db=dbs,
                 target_db=dbt,
-                # Deliberately override these good params with bad params
```
ricellis: This removal doesn't seem correct. The test title is  since the test title suggests this is `test_follow_replication_with_errors` and this section says "override these good params with bad params" so I think these are supposed to be "broken" source and targets. Although the test is currently skipped anyway which probably means it doesn't matter much at all.

emlaver: @ricellis Right, although the test is currently skipped it would fail with this removal.  We have issue #114 for cleaning up the integration tests.  Maybe it's better to remove the whole test since it was using these optional parameters.

ricellis: Is there a change we can make that would mean if this test was re-enabled it would still fail? Can we specify a target that doesn't exist in object form?

ricellis: If there isn't I suggest we at the least add a "fail" here with a comment that the test needs a configuration change to do what it is supposed to in the event that anyone re-enables it.

emlaver: @ricellis Since the test is verifying that it will exit the follow loop on [these lines](https://github.com/cloudant/python-cloudant/blob/3535a6e35bc32f4cb2622a0295885e34ec4766e0/tests/integration/replicator_test.py#L234-L236), we can use an invalid target URL on line 204 i.e. `dbtarget = '_bad_db_name_'`

PR #365, opened 2018-03-14
Support IAM authentication in replication documents

### LINKED ISSUE #354: Support IAM authentication in replication documents.
_Parent Issue: https://github.com/cloudant/cloudant-sync/issues/31_

Currently `db.creds['basic_auth']` returns `None` when using an IAM client . As a result, all replication documents created using IAM clients _don't_ contain source/target database credentials.

_Work Required:_
* Expose the IAM key via the client session ([here](https://github.com/cloudant/python-cloudant/blob/2.7.0/src/cloudant/_client_session.py#L194)).
* Rework the [Replicator](https://github.com/cloudant/python-cloudant/blob/2.7.0/src/cloudant/replicator.py#L77-L93) class to correctly inject the IAM key into the replication document (when databases are backed by an IAM client).
* Enable replicator tests for IAM clients (currently disabled).
* Add more unit tests as appropriate.

smithsz:

## What
Support IAM authenticated databases when replicating.

## How
Expose the IAM API key via the client session and inject into the replication document where applicable. 

## Testing
Includes additional mock unit tests.

## Issues
Fixes #354.

### DIFF
--- a/src/cloudant/_client_session.py
+++ b/src/cloudant/_client_session.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
+# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -194,6 +194,15 @@ class IAMSession(ClientSession):
         self._api_key = api_key
         self._token_url = os.environ.get(
             'IAM_TOKEN_URL', 'https://iam.bluemix.net/identity/token')
+
+    @property
+    def get_api_key(self):
+        """
+        Get IAM API key.
+
+        :return: IAM API key.
+        """
+        return self._api_key
 
     def login(self):
         """
--- a/src/cloudant/replicator.py
+++ b/src/cloudant/replicator.py
@@ -60,31 +60,51 @@ class Replicator(object):
 
         :returns: Replication document as a Document instance
         """
+        if source_db is None:
+            raise CloudantReplicatorException(101)
+
+        if target_db is None:
+            raise CloudantReplicatorException(102)
 
         data = dict(
             _id=repl_id if repl_id else str(uuid.uuid4()),
             **kwargs
         )
 
-        if source_db is None:
-            raise CloudantReplicatorException(101)
-        data['source'] = {'url': source_db.database_url}
-        if not source_db.admin_party:
-            data['source'].update(
-                {'headers': {'Authorization': source_db.creds['basic_auth']}}
-            )
+        # replication source
 
-        if target_db is None:
-            raise CloudantReplicatorException(102)
+        data['source'] = {'url': source_db.database_url}
+        if source_db.admin_party:
+            pass  # no credentials required
+        elif source_db.client.is_iam_authenticated:
+            data['source'].update({'auth': {
+                'iam': {'api_key': source_db.client.r_session.get_api_key}
+            }})
+        else:
+            data['source'].update({'headers': {
+                'Authorization': source_db.creds['basic_auth']
+            }})
+
+        # replication target
+
         data['target'] = {'url': target_db.database_url}
-        if not target_db.admin_party:
-            data['target'].update(
-                {'headers': {'Authorization': target_db.creds['basic_auth']}}
-            )
+        if target_db.admin_party:
+            pass  # no credentials required
+        elif target_db.client.is_iam_authenticated:
+            data['target'].update({'auth': {
+                'iam': {'api_key': target_db.client.r_session.get_api_key}
+            }})
+        else:
+            data['target'].update({'headers': {
+                'Authorization': target_db.creds['basic_auth']
+            }})
+
+        # add user context delegation
 
         if not data.get('user_ctx'):
-            if (target_db and not target_db.admin_party or
-                    self.database.creds):
+            if target_db and target_db.admin_party:
+                pass  # noop - not required for admin party mode
+            elif self.database.creds and self.database.creds.get('user_ctx'):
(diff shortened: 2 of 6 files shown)


emlaver:
Update copyright for `_client_session.py`

src/cloudant/_client_session.py
```diff
@@ -195,6 +195,15 @@ def __init__(self, api_key, server_url, **kwargs):
         self._token_url = os.environ.get(
             'IAM_TOKEN_URL', 'https://iam.bluemix.net/identity/token')
 
+    @property
+    def api_key(self):
```
emlaver: Should this be named `get_api_key` following the design of the functions in this class e.g. `_get_access_token`?

tests/unit/replicator_mock_tests.py
```diff
@@ -0,0 +1,175 @@
...
+        self.source_db = 'source_db'
+        self.target_db = 'target_db'
+
+    def test_using_admin_party_source_and_target(self):
+        # setup mocks
+        m_admin_party_client = mock.MagicMock()
+        type(m_admin_party_client).admin_party = mock.PropertyMock(
+            return_value=True)
+        type(m_admin_party_client).server_url = mock.PropertyMock(
+            return_value=self.server_url)
```
emlaver: I'm wondering if we should put this code into it's own method so that it can be called by all tests in this class.  It would require 4 if blocks for options `admin_party`, `server_url`, `r_session`, and `is_iam_authenticated`.

src/cloudant/replicator.py
```diff
@@ -60,27 +60,46 @@ def create_replication(self, source_db=None, target_db=None,
```
ricellis: The existing `user_ctx` applies to a target only. I think we need to make `source_user(_ctx?)` and `target_user(_ctx?)` args that can be used to specify. We should alias the `user_ctx` to `target_user` to maintain backwards compatibility.

src/cloudant/replicator.py
```diff
@@ -60,27 +60,46 @@ def create_replication(self, source_db=None, target_db=None,
...
+        # replication source
+
         data['source'] = {'url': source_db.database_url}
-        if not source_db.admin_party:
-            data['source'].update(
-                {'headers': {'Authorization': source_db.creds['basic_auth']}}
-            )
+        if source_db.admin_party:
+            pass  # no credentials required
+        elif source_db.client.is_iam_authenticated:
```
ricellis: The auth used in a replicator document should be independent of the method with which the client is authenticated. For example there is no reason why I can't login python-cloudant with my account credentials and create a replicator document using any other credentials including IAM API keys.

This code should be a fallback only if the `source_user` is empty and target_user are empty. I think however, that a new version of the `create_replication` should require some credential information and the fallback should only be used in the backwards compatibility case.

src/cloudant/replicator.py
```diff
@@ -60,27 +60,46 @@ def create_replication(self, source_db=None, target_db=None,
...
+        # replication target
 
-        if target_db is None:
-            raise CloudantReplicatorException(102)
         data['target'] = {'url': target_db.database_url}
-        if not target_db.admin_party:
-            data['target'].update(
-                {'headers': {'Authorization': target_db.creds['basic_auth']}}
-            )
+        if source_db.admin_party:
```
ricellis: Shouldn't this be the target? Also shouldn't we only do these checks if a user_ctx wasn't provided?

src/cloudant/replicator.py
smithsz: Not sure I understand what you mean here. Can we please go through it when you get a mo? Thanks.

src/cloudant/replicator.py
smithsz: `source_db` and `target_db` are supplied to the `create_replication` method. They are type database objects and therefore don't share credentials used to POST the replication doc.

src/cloudant/replicator.py
ricellis: It was a long time ago and I'm struggling to reverse engineer my meaning too! I think what I was getting at was that the `user_ctx` default setting is dependent on some checks on the `target`, but not the source. My suggestion above doesn't make sense either though.

What I think makes sense is that the default `user_ctx` is set if one isn't supplied, but should not depend on any checks for the source or target database.

src/cloudant/replicator.py
ricellis: Yes, I see what you mean, everything is an object, so the specific credentials will have been supplied up front in construction of the necessary source or target database objects (or more specifically the client instances needed for those if they are distinct from the replicator.

I think it would be nice if a replication could be created with a source & target URLs and the associated API keys i.e. 4 strings instead of needing to potentially instantiate multiple clients and DBs, but that is out of scope here.

src/cloudant/replicator.py
smithsz: Done. Removed admin party check in eaaa7f1.

src/cloudant/replicator.py
ricellis: Yep, that's why I think it is out of scope to try and re-introduce that in a working fashion here.

src/cloudant/replicator.py
```diff
@@ -101,11 +100,9 @@ def create_replication(self, source_db=None, target_db=None,
 
         # add user context delegation
 
-        if not data.get('user_ctx'):
-            if target_db and target_db.admin_party:
-                pass  # noop - not required for admin party mode
-            elif self.database.creds and self.database.creds.get('user_ctx'):
-                data['user_ctx'] = self.database.creds['user_ctx']
+        if not data.get('user_ctx') and self.database.creds and \
```
emlaver: Is there a case where `user_ctx` doesn't exist in `data` and `database.creds`?

smithsz: yeah, admin party mode.

PR #368, opened 2018-04-05
Add special handling for auto_renew with _users DB

tiangolo:

## What

As the [ `_users` database could return a 404 error while the problem is actually about session expiration](http://docs.couchdb.org/en/2.1.1/intro/security.html#users-public-information), it requires special handling when using `auto_renew=True`. 

The current implementation only checks for `403` and `401`. And errors `404` should be taken into account only when dealing with the `_users` database.

## How

* Add `is_users_db_url` method to `ClientSession` class to check if a requested URL is for the `_users` DB.
* Modify the method `request` of the subclass `CookieSession` to check if the error is a `404` and if the request is for the `_users` DB.

## Testing

Currently, when I run the tests from the `master` branch, there are several tests that are not passing. That's before any of my modifications.

With respect to this PR, as this is a very specific behavior of the `_users` database in CouchDB, I'm not sure if you want me to add all the specific code with all the boilerplate for that.

## Issues

I found the issue while developing my application, investigated and solved it right away. I didn't create an issue for it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 2.8.2 (2018-04-05)
+
+- [FIXED] Added special handling for requesting the `_users` database, as it can send `404` errors when a session expires.
+
 # 2.8.1 (2018-02-16)
 
 - [FIXED] Installation failures of 2.8.0 caused by missing VERSION file in distribution.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,5 @@
 # Unreleased
 
-# 2.8.2 (2018-04-05)
-
 - [FIXED] Added special handling for requesting the `_users` database, as it can send `404` errors when a session expires.
 
 # 2.8.1 (2018-02-16)
--- a/src/cloudant/_client_session.py
+++ b/src/cloudant/_client_session.py
@@ -104,8 +104,9 @@ class ClientSession(Session):
 
     def is_users_db_url(self, url):
         """
-        Check if a requested url is for the users database, to be used with auto_renew=True
-        as this database could return a 404 error while the problem could actually be an expired session.
+        Check if a requested url is for the users database, to be used
+        with auto_renew=True as this database could return a 404 error
+        while the problem could actually be an expired session.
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 2.8.2 (2018-04-05)
+
+- [FIXED] Added special handling for requesting the `_users` database, as it can send `404` errors when a session expires.
```
emlaver: Please move this CHANGES entry under `Unreleased` and remove the `# 2.8.2 (2018-04-05)` line.  Our team will handle releasing the next version.

src/cloudant/_client_session.py
```diff
@@ -174,7 +187,12 @@ def request(self, method, url, **kwargs):
             resp.status_code == 401
         ))
 
-        if is_expired:
+        can_be_user_db_expired = (
```
emlaver: Why not add this code to the `any` block above for `is_expired`?

CHANGES.md
tiangolo: Sure!

src/cloudant/_client_session.py
tiangolo: I was thinking on doing that at first, but then my intention was to make it more explicit.

`is_expired` would be `True` for situations in which it is certain that the session is expired. While `can_be_user_db_expired` would not be certain that the session is expired, just that it "can be expired". Because CouchDB would throw a 404 error if the user doesn't exists or if the user that does the request is not the same user himself (like an anonymous user, or an expired session) or a server admin.

If you prefer, I can update and simplify the code to have everything in the `any` block.

emlaver:
@tiangolo Thanks for your work but we've decided as a team to investigate and fix the error codes without using your special casing.  I'll update this comment with any related new GH issues/PR.

tiangolo:
Thanks for reporting back. 

And thanks in advance for posting here updates on how you decide to handle it. 

Also, please let me know if you need help coding it, as it is a feature I'm needing and I'm currently just "hacking" to use it in my projects.

PR #372, opened 2018-04-17
Updated Travis CI and unit tests to run against CouchDB 2.1.1

emlaver:

- [ ] You have added tests for any code changes -- PR is for updating existing tests

## What

Updated Travis CI to use the CouchDB Docker image and updated the existing tests to pass against the latest CouchDB version.

## How

- Build CouchDB docker image in Travis
- Updated CouchDB setup to retrieve node name from `_membership` when admin user is created (ADMIN_PARTY=false)
- Added functionality for setting up and tearing down` _global_changes` database as it's required by the `_db_updates` unit tests
- Updated existing `DbUpdatesTestsBase.create_dbs` function to verify that all created test databases are listed in the `_db_updates` feed
- Additional fixes and updates to db_updates, view, and changes unit tests
## Testing

Fixed existing tests to pass against CouchDB 2.1.1.
No new tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+
 language: python
 
 python:
@@ -9,9 +11,18 @@ env:
   - ADMIN_PARTY=false
 
 services:
-  - couchdb
+  - docker
+
+before_install:
+  - docker pull apache/couchdb:2.1.1
+  - docker run -d -it -p 5984:5984 apache/couchdb:2.1.1 --with-admin-party-please --with-haproxy
 
 install: "pip install -r requirements.txt && pip install -r test-requirements.txt"
+
+before_script:
+    - curl -X PUT http://localhost:5984/_users
+    - curl -X PUT http://localhost:5984/_replicator
+
 # command to run tests
 script:
     - pylint ./src/cloudant
--- a/tests/unit/db_updates_tests.py
+++ b/tests/unit/db_updates_tests.py
@@ -34,25 +34,36 @@ class DbUpdatesTestsBase(UnitTestDbBase):
     """
     Common _db_updates tests methods
     """
-
+    
     def setUp(self):
         """
         Set up test attributes
         """
         super(DbUpdatesTestsBase, self).setUp()
         self.client.connect()
-        self.new_dbs = list()
+        self.create_db_updates()
+        self.create_dbs()
 
     def tearDown(self):
         """
         Reset test attributes
         """
-        [db.delete() for db in self.new_dbs]
+        self.delete_db_updates()
         self.client.disconnect()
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -9,9 +11,18 @@ env:
...
+  - docker
+
+before_install:
+  - docker pull apache/couchdb:2.1.1
+  - docker run -d -it -p 5984:5984 apache/couchdb:2.1.1 --with-admin-party-please --with-haproxy
 
 install: "pip install -r requirements.txt && pip install -r test-requirements.txt"
+
+before_script:
+    - curl -X PUT http://localhost:5984/_users
```
ricellis: You'll need a line in here to make sure that the endpoint is up before trying to create the DBs, otherwise we may well run into timing problems later.
Something like https://github.com/cloudant/sync-cordova-plugin/blob/master/.travis.yml#L66

tests/unit/unit_t_db_base.py
```diff
@@ -100,14 +100,18 @@ def setUpClass(cls):
                 return
 
             if os.environ.get('DB_USER') is None:
+                # Get couchdb docker node name
```
ricellis: I think the standard is `nonode@nohost` but I guess this future proofs it a bit :)

.travis.yml
emlaver: Added in 62ad10e.

tests/unit/db_updates_tests.py
```diff
@@ -34,25 +34,36 @@ class DbUpdatesTestsBase(UnitTestDbBase):
...
+    def create_dbs(self, count=2):
         self.new_dbs += [(self.client.create_database(self.dbname())) for x in range(count)]
+        # Verify that all created databases are listed in _db_updates
+        all_dbs_exist = False
+        while not all_dbs_exist:
+            changes = list()
+            feed = Feed(self.client, timeout=1000)
+            for change in feed:
+                changes.append(change)
+                if len(changes) == 3:
```
tomblench: Does the count of 3 include the _users db?

tomblench: Actually, it looks like this assert doesn't account for the case where `count` is set to the non-default value.

emlaver: @tomblench The _users and _replicator dbs are created before _global_changes.  During the creation of _global_changes in the setup of `db_updates_tests.py`, the only databases that should be seen in the feed are `_dbs` (included during `_global_changes` creation) and the two databases for testing in `create_dbs`.

You're right that the assertion only works for case `count=2`.  Since the tests only need two databases, I will remove the `count` parameter.

ricellis:
still +1

PR #378, opened 2018-05-02
Added support for IAM API key in cloudant_bluemix method

emlaver:

## What

Updated VCAP to support IAM API key and username/password credentials.

## How

- Updated CloudFoundryService to create a client session using IAM API key in VCAP
- Added iam_api_key property
- Handle cases when IAM or legacy credentials do not exist
- Added test cases to verify new IAM/creds case logic

## Testing

- Updated sample VCAP dictionaries in `CloudFoundryServiceTests`.
- Verify VCAP with IAM and legacy creds
- Assert new logic by creating VCAP client with either IAM or legacy creds
- If IAM key is missing and either/both username and password are missing, throw exception

## Issues

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
+# Copyright (C) 2015, 2018 IBM Corp. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -23,7 +23,7 @@ from collections import Sequence
 import json
 
 from ._2to3 import LONGTYPE, STRTYPE, NONETYPE, UNITYPE, iteritems_
-from .error import CloudantArgumentError, CloudantException
+from .error import CloudantArgumentError, CloudantException, CloudantClientException
 
 # Library Constants
 
@@ -306,9 +306,14 @@ class CloudFoundryService(object):
                     credentials = service['credentials']
                     self._host = credentials['host']
                     self._name = service.get('name')
-                    self._password = credentials['password']
                     self._port = credentials.get('port', 443)
                     self._username = credentials['username']
+                    if 'apikey' in credentials:
+                        self._iam_api_key = credentials['apikey']
+                    elif 'username' in credentials and 'password' in credentials:
+                        self._password = credentials['password']
+                    else:
+                        raise CloudantClientException(103)
                     break
             else:
                 raise CloudantException('Missing service in VCAP_SERVICES')
@@ -355,3 +360,8 @@ class CloudFoundryService(object):
     def username(self):
         """ Return service username. """
         return self._username
+
+    @property
+    def iam_api_key(self):
+        """ Return service IAM API key. """
+        return self._iam_api_key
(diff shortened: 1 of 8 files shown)


src/cloudant/_common_util.py
```diff
@@ -306,9 +306,17 @@ def __init__(self, vcap_services, instance_name=None, service_name=None):
...
-                    self._password = credentials['password']
                     self._port = credentials.get('port', 443)
                     self._username = credentials['username']
+                    if 'apikey' in credentials:
+                        self._iam_api_key = credentials['apikey']
+                    elif 'username' in credentials and 'password' in credentials:
+                        self._password = credentials['password']
+                    else:
+                        raise CloudantException(
+                            "Invalid service: "
```
smithsz: Why not `CloudantClientException(103)` here?

src/cloudant/client.py
```diff
@@ -788,9 +788,17 @@ def bluemix(cls, vcap_services, instance_name=None, service_name=None, **kwargs)
...
         """
         service_name = service_name or 'cloudantNoSQLDB'  # default service
-        service = CloudFoundryService(vcap_services,
-                                      instance_name=instance_name,
-                                      service_name=service_name)
+        try:
+            service = CloudFoundryService(vcap_services,
+                                          instance_name=instance_name,
+                                          service_name=service_name)
+        except CloudantException:
```
smithsz: Do we need this try/except? Why not just let it throw?

src/cloudant/client.py
```diff
@@ -788,9 +788,17 @@ def bluemix(cls, vcap_services, instance_name=None, service_name=None, **kwargs)
...
+            raise CloudantClientException(103)
+
+        if hasattr(service, 'iam_api_key'):
+            return Cloudant.iam(service.username,
+                                service.iam_api_key,
```
smithsz: How about `service.iam_api_key or service.password`?

src/cloudant/_common_util.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015, 2017 IBM Corp. All rights reserved.
+# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
```
ricellis: (C) or ©

tests/unit/cloud_foundry_tests.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2016 IBM. All rights reserved.
+# Copyright (c) 2016, 2018 IBM Corp. All rights reserved.
```
ricellis: (C) ©

src/cloudant/client.py
emlaver: I wanted to keep the exceptions consistent with each class exception.  Since `CloudFoundryService` is used during client creation, we could consider updating all `CloudantException` in the function to `CloudantClientException`.

src/cloudant/client.py
emlaver: I originally used `service.iam_api_key`, but it will throw `AttributeError` if the attribute is not part of `CloudFoundryService`.  I have some tests where `iam_api_key` is not set to test the legacy creds logic.  
Also, it only seems that the updated VCAP version has IAM key or IAM key with legacy creds.  This would provide backward compatibility with VCAPs that have no IAM API key.

tests/unit/cloud_foundry_tests.py
emlaver: Fixed in df6d3e5.

src/cloudant/_common_util.py
emlaver: Fixed in df6d3e5.

PR #382, opened 2018-05-07
[WIP] CORS utils in Client

tiangolo:
- [ x ] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](https://github.com/cloudant/python-cloudant/blob/master/DCO1.1.txt)
- [ ] You have added tests for any code changes
- [ ] You have updated the [CHANGES.md](https://github.com/cloudant/python-cloudant/blob/master/CHANGES.md)
- [ ] You have completed the PR template below:

## What

I have created several utils for working with CouchDB from Python that I'm using in some projects. Some of them I think could belong to this package.

Namely to:

* Set up CORS.
* Configure Cookie, timeout, etc.
* Enable CouchPerUser integrated plug-in (integrated in 2.x.x).
* Generate the CouchDB user ID from a username (a string concatenation)
* Set up users (a new database comes without the `_users` DB)
* Get DB ID for a user as generated by CouchPerUser
* others

I think I would create a PR for each one, but I want to first discuss with you what would be desirable / acceptable in the project, where each feature should be in and if it anything should be changed, how.

## How

Mainly using the same `client.r_session` to send requests to the specific paths.

## Testing

Before implementing tests and integrating those utils, I would like to discuss it with you.

## Issues

No issue yet.

### DIFF
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -298,6 +298,9 @@ class CouchDB(dict):
         """
         return Feed(self, raw_data, **kwargs)
 
+    def enable_cors(self):
+        raise NotImplementedError()
+
     def keys(self, remote=False):
         """
         Returns the database names for this client. Default is


ricellis:
@tiangolo thanks for volunteering with these suggestions. We have the following feedback based on our needs and plans for the library.

> * Set up CORS.

Yes, we could accept a PR for calling the CORS endpoints. We have similar functionality in the [nodejs-cloudant library](https://github.com/cloudant/nodejs-cloudant/#cors).
It would need to call the [Cloudant CORS endpoint](https://console.bluemix.net/docs/services/Cloudant/api/cors.html#configuration-endpoints) (i.e. `/_api/v2/user/config/cors`) from the `Cloudant` class.
We have had issues elsewhere with testing CORS preventing parallel execution of tests because of changing global CORS state, so we would prefer if testing of the correct requests was done with mocks for this functionality.

We could discuss also adding something for the `CouchDB` class, but I think it is complicated by [CouchDB CORS being enabled via the httpd configuration](http://docs.couchdb.org/en/stable/config/http.html?highlight=cors#cross-origin-resource-sharing). In a CouchDB 2.x setup this typically would only be available on the node-local port, which would mean having a different URL to the usual client init (and only usable when running python-cloudant on the localhost) which would add significant complexity or limit the usability of the function.

> * Configure Cookie, timeout, etc.

Cookie handling, session renewal and connect/read timeouts are already provided in python-cloudant, can you elaborate on what you would propose for this?

> * Enable CouchPerUser integrated plug-in (integrated in 2.x.x).
> * Get DB ID for a user as generated by CouchPerUser

Database per user is considered an anti-pattern and is likely to lead to problems, see [Rule 5](https://medium.com/ibm-watson-data-lab/cloudant-best-and-worst-practices-7ee2040da1b). As a result we won't accept any couch_peruser modifications, sorry.

> * Generate the CouchDB user ID from a username (a string concatenation)
> * Set up users (a new database comes without the _users DB)

For Cloudant the usernames and `_users` DB generation are handled elsewhere. It is straightforward to issue the `PUT` requests to create the `_users` DB (and the other metadata DBs if needed). I don't think we need a method to specifically issue this `PUT` and the list of metadata DBs is likely to go stale on the client side as the server evolves.

tiangolo:
Thanks for all the feedback @ricellis .

> - Set up CORS.

I don't use Cloudant, I'm only using CouchDB in my own servers, so I don't really know about Cloudant, sorry :disappointed: . I was expecting it to inherit the configurations for CouchDB, but I see that's not always the case.

In my code, I've been able to modify the CORS settings using the normal port. But that might be because I'm using Docker.

The code looks like:

```python
# A Docker container
# config.COUCHDB_URL = 'http://couchdb:5984'
# A string of hosts separated by commas
# config.COUCHDB_CORS_ORIGINS = 'http://example.com, http://stag.example.com, https://example.com, https://stag.example.com, '

def enable_cors():
    couch_cors_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/httpd/enable_cors'
    couch_cors_credentials_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/cors/credentials'
    couch_cors_origins_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/cors/origins'
    # Instead of True, a string 'true'
    client.r_session.put(couch_cors_url, json='true')
    client.r_session.put(couch_cors_credentials_url, json='true')
    client.r_session.put(couch_cors_origins_url, json=config.COUCHDB_CORS_ORIGINS)
```

> - Configure Cookie, timeout, etc.

Sorry for the lack of explanation. I was not referring to configuring the Python client but the CouchDB server itself. 

By default, CouchDB cookies last 15 min, if a front end application connects directly to the CouchDB server, a session of only 15 min could be too short in some applications. I was referring to setting that configuration in the CouchDB server itself.

The code looks like:

```python
# A Docker container
# config.COUCHDB_URL = 'http://couchdb:5984'
# config.COUCHDB_AUTH_TIMEOUT = 600 # seconds

def setup_cookie():
    couch_persistent_cookies_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/couch_httpd_auth/allow_persistent_cookies'
    couch_timeout_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/couch_httpd_auth/timeout'
    client.r_session.put(couch_persistent_cookies_url, json='true')
    res = client.r_session.put(couch_timeout_url, json=str(config.COUCHDB_AUTH_TIMEOUT))
```

> - CouchPerUser

I see that it's an anti-pattern for Cloudant given the way it is designed. Again, I was thinking on pure CouchDB. And given that permissions are per DB (in CouchDB and Cloudant), for cases where it is needed to let the user interact directly with his DB (for example, in an offline application with PouchDB that synchronizes data with CouchDB), there's no way around. But I understand that it's not something you would like having integrated.

Here's how the code looks like anyway:

```python
# A Docker container
# config.COUCHDB_URL = 'http://couchdb:5984'

def enable_couch_peruser():
    couch_peruser_url = config.COUCHDB_URL + '/_node/nonode@nohost/_config/couch_peruser/enable'
    client.r_session.put(couch_peruser_url, json='true')
```

> - Set up users

Again, I was thinking specifically in CouchDB.

---

I see that the main focus of this package is Cloudant, so I guess that my utilities for CouchDB are probably a better fit directly in app code or an external package / fork than integrated here, as they don't apply to Cloudant that well. Thanks again for the thorough feedback.

ricellis:
We do try to keep the library working for both CouchDB and Cloudant because the API surface is _mostly_ the same. It seems though that a lot of the functions you would like are for configuration options, which is one of the places that Cloudant mostly hides away from users of the service. You are right that Cloudant is our main focus, so those parts don't fit so well with our plans. As such I'll close this for now. Thanks again for the ideas and discussion.

PR #384, opened 2018-05-21
Updated Travis to run against CouchDB 1.7.1 and Python 3.6

emlaver:
- Updated COUCHDB_VERSION variable to just contain version number
- Get couchdb node name from _membership only if CouchDB version is 2.1.1
- Updated tests to support CouchDB 1.7.x
- Added helper method to check what version
- Removed javascript validation test that doesn’t apply to CouchDB 1.7.1

## What

Added CouchDB 1.7.1 to travis test matrix and updated Python 3.5 env to 3.6.

## How

- Updated logic to check if COUCHDB_VERSION starts with 1 for skipping tests when running against CouchDB 1.6.x and 1.7.x.
- Only get the CouchDB node name from _membership if CouchDB version is 2.1.1
- Added helper method to check if CouchDB version is 1.x

## Testing

No new tests.
Removed `test_view_callable_with_invalid_javascript` in `view_tests.py` as it's not valid in CouchDB 1.7.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,20 @@ language: python
 
 python:
   - "2.7"
-  - "3.5"
+  - "3.6"
 
 env:
-  - ADMIN_PARTY=true
-  - ADMIN_PARTY=false
+  - ADMIN_PARTY=true COUCHDB_VERSION=2.1.1
+  - ADMIN_PARTY=false COUCHDB_VERSION=2.1.1
+  - ADMIN_PARTY=true COUCHDB_VERSION=1.7.1
+  - ADMIN_PARTY=false COUCHDB_VERSION=1.7.1
 
 services:
   - docker
 
 before_install:
-  - docker pull apache/couchdb:2.1.1
-  - docker run -d -it -p 5984:5984 apache/couchdb:2.1.1 --with-admin-party-please --with-haproxy
+  - docker pull apache/couchdb:$COUCHDB_VERSION
+  - docker run -d -p 5984:5984 apache/couchdb:$COUCHDB_VERSION
 
 install: "pip install -r requirements.txt && pip install -r test-requirements.txt"
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@
 - [NEW] Moved `create_query_index` and other query related methods to `CouchDatabase` as the `_index`/`_find` API is available in CouchDB 2.x.
 - [NEW] Support IAM authentication in replication documents.
 - [IMPROVED] Added support for IAM API key in `cloudant_bluemix` method.
-- [IMPROVED] Updated Travis CI and unit tests to run against CouchDB 2.1.1.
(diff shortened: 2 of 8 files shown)


CHANGES.md
```diff
@@ -4,7 +4,8 @@
 - [NEW] Moved `create_query_index` and other query related methods to `CouchDatabase` as the `_index`/`_find` API is available in CouchDB 2.x.
 - [NEW] Support IAM authentication in replication documents.
 - [IMPROVED] Added support for IAM API key in `cloudant_bluemix` method.
-- [IMPROVED] Updated Travis CI and unit tests to run against CouchDB 2.1.1.
+- [IMPROVED] Updated Travis CI and unit tests to run against CouchDB 1.7.1 and 2.1.1.
+- [IMPROVED] Updated Python environment for testing in Travis from 3.5 to 3.6.
```
ricellis: I actually don't think we need these entries as they are related to the testing not the library release.
That said if we're saying that Python 3.6 is supported it may be worth saying something like:
`- [IMPROVED] Verified library operation on Python 3.6.`(we should say which 3.6.x version)

.travis.yml
```diff
@@ -4,18 +4,20 @@ language: python
...
+  - ADMIN_PARTY=false COUCHDB_VERSION=1.7.1
 
 services:
   - docker
 
 before_install:
-  - docker pull apache/couchdb:2.1.1
-  - docker run -d -it -p 5984:5984 apache/couchdb:2.1.1 --with-admin-party-please --with-haproxy
+  - docker pull apache/couchdb:$COUCHDB_VERSION
+  - docker run -d -p 5984:5984 apache/couchdb:$COUCHDB_VERSION
```
ricellis: are the admin party and haproxy flags unnecessary?

emlaver: I don't think they were needed in the first place.  They were added in the last Travis PR but the tests seem to be working fine without those flags.

emlaver:
@smithsz @ricellis The `skipIf` tags needed to be fixed since they were skipping tests against CouchDB 1.7.x, and running CouchDB `_db_updates` on Cloudant which failed as expected.  I’ve updated the tags and test logic in c4e880e.  Cloudant and CouchDB 1.7.x are passing but 2 tests are failing that are related to IAM.

tests/unit/db_updates_tests.py
```diff
@@ -57,44 +54,61 @@ def tearDown(self):
...
-
+        if not self.is_couchdb_1x_version():
+            self.dbs = ['_dbs', self.new_dbs[0].database_name, self.new_dbs[1].database_name]
+            types = ['created', 'updated']
+            for doc in changes:
+                self.assertIsNotNone(doc['seq'])
+                self.assertTrue(doc['db_name'] in self.dbs)
+                self.assertTrue(doc['type'] in types)
+        else:
+            # for doc in changes:
```
smithsz: forget to remove this? the 3 `assertDictEqual` asserts are fine.

PR #388, opened 2018-06-05
Support new view parameters

tomblench:
## What

Support new view parameters

## How

Add support and documentation for extra `kwargs` to `get_view_result`: `stable` and `update`.

## Testing

Added tests in `param_translation_tests.py`, `view_execution_tests.py`.

## Issues

#252

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
+# Copyright (C) 2015, 2018 IBM Corp. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (C) 2015, 2018 IBM Corp. All rights reserved.
+# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -60,9 +60,11 @@ RESULT_ARG_TYPES = {
     'limit': (int, LONGTYPE, NONETYPE,),
     'reduce': (bool,),
     'skip': (int, LONGTYPE, NONETYPE,),
+    'stable': (bool,),
     'stale': (STRTYPE,),
     'startkey': (int, LONGTYPE, STRTYPE, Sequence,),
     'startkey_docid': (STRTYPE,),
+    'update': (STRTYPE,),
 }
 
 # pylint: disable=unnecessary-lambda
@@ -196,7 +198,7 @@ def _py_to_couch_translate(key, val):
     equivalent.
     """
     try:
-        if key in ['keys', 'endkey_docid', 'startkey_docid', 'stale']:
+        if key in ['keys', 'endkey_docid', 'startkey_docid', 'stale', 'update']:
             return {key: val}
         elif val is None:
             return {key: None}
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -312,6 +312,8 @@ class CouchDatabase(dict):
         :param bool reduce: True to use the reduce function, false otherwise.
         :param int skip: Skip this number of rows from the start.
             Not valid when used with :class:`~cloudant.result.Result` iteration.
+        :param bool stable: Whether or not the view results should be returned
+            from a "stable" set of shards.
         :param str stale: Allow the results from a stale view to be used. This
             makes the request return immediately, even if the view has not been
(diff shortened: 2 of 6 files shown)


src/cloudant/_common_util.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (C) 2015, 2018 IBM Corp. All rights reserved.
+# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
```
ricellis: should be `C` so this was fine before

src/cloudant/database.py
```diff
@@ -312,6 +312,8 @@ def get_view_result(self, ddoc_id, view_name, raw_result=False, **kwargs):
         :param bool reduce: True to use the reduce function, false otherwise.
         :param int skip: Skip this number of rows from the start.
             Not valid when used with :class:`~cloudant.result.Result` iteration.
+        :param bool stable: Whether or not the view results should be returned
+            from a "stable" set of shards.
         :param str stale: Allow the results from a stale view to be used. This
```
ricellis: Given that there is no specific way to deprecate this `stale` parameter we should consider either:
i) adding a comment here that `stable` and `update` are preferred
ii) delete this `:param` on the basis that it won't break anything while it is still listed in `RESULT_ARG_TYPES`

src/cloudant/result.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015 IBM. All rights reserved.
+# Copyright (c) 2015, 2018 IBM Corp. All rights reserved.
```
ricellis: This does need to be uppercase

src/cloudant/result.py
```diff
@@ -150,6 +150,8 @@ class Result(object):
     :param bool reduce: True to use the reduce function, false otherwise.
     :param int skip: Skip this number of rows from the start.
         Not valid when used with key iteration.
+    :param bool stable: Whether or not the view results should be returned from
+        a "stable" set of shards.
     :param str stale: Allow the results from a stale view to be used. This
```
ricellis: My previous comment around "deprecating" this also applies here I guess

src/cloudant/database.py
```diff
@@ -321,6 +323,11 @@ def get_view_result(self, ddoc_id, view_name, raw_result=False, **kwargs):
             access and key slicing.
         :param str startkey_docid: Return records starting with the specified
             document ID.
+        :param str update: Determine whether the view in question should be
+            updated prior to or after responding to the user. Valid values are:
+            false: return results before updating the view; true: Return results
+            after updating the view; lazy: Return the view results without
+            waiting for an update, but update them immediately after the request
```
ricellis: `.` after this sentence, to match the others

src/cloudant/result.py
```diff
@@ -158,6 +160,12 @@ class Result(object):
         Not valid when used with key access and key slicing.
     :param str startkey_docid: Return records starting with the specified
         document ID.
+    :param str update: Determine whether the view in question should be
+        updated prior to or after responding to the user. Valid values are:
+        false: return results before updating the view; true: Return results
+        after updating the view; lazy: Return the view results without
+        waiting for an update, but update them immediately after the request
```
ricellis: As above `.` needed

tests/unit/param_translation_tests.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2016 IBM. All rights reserved.
+# Copyright (c) 2016, 2018 IBM Corp. All rights reserved.
```
ricellis: you guessed it

tests/unit/view_execution_tests.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2016 IBM. All rights reserved.
+# Copyright (c) 2016, 2018 IBM Corp. All rights reserved.
```
ricellis: ditto

tests/unit/view_execution_tests.py
```diff
@@ -480,8 +480,48 @@ def test_stale_update_after(self):
         try:
             self.view001(stale='update_after')
         except Exception as err:
-            self.assertFail(str(err), 'An unexpected error was encountered.')
+            self.fail('An unexpected error was encountered:' +str(err))
+            
+    def test_stable_true(self):
```
ricellis: Do these tests need to be tagged for Couch2x/Cloudant because of the new parameters? I guess Couch 1x probably ignores them?

tomblench: They seem to be ignored, at least by couch 1.7

emlaver: Since Travis is passing, I'm assuming CouchDB 1.7.1 is ignoring this new param test.

src/cloudant/database.py
tomblench: How about i)

"Note that this parameter is deprecated and the appropriate combination of `stable` and `update` should be used instead."

PR #391, opened 2018-06-19
Template updates

ricellis:

- [x] You have added tests for any code changes - N/A
- [x] You have updated the [CHANGES.md](https://github.com/cloudant/python-cloudant/blob/master/CHANGES.md) - N/A

## What

Updated issue and PR templates and contributing file.


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,5 +1,33 @@
...
+
+## Bug Description
+
+### 1. Steps to reproduce and a small code sample that demonstrates the issue
+<!--
+EXAMPLE:
+1. Initialize client
+2. List DBs
+
+cc = new CloudantClient(...)
```
emlaver: This need to be updated to something like:
```
client = Cloudant(...)
dbs = client.all_dbs()
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,5 +1,33 @@
...
+<!--
+Got a 400 bad request
+-->
+
+## Environment details
+<!--
+- Version(s) that are affected by this issue.
+    > 2.0.2
+- Python version
+    >3.6.2
```
emlaver: Add space between `>` and `3.6.2`.

ricellis: Fixed in eba45bb591cad33951571d3d7067e0021817dbca

.github/ISSUE_TEMPLATE.md
ricellis: Well the idea was that this example was the same pseudo-code in every repo so we didn't have to update it all the time. If you think that is too confusing we can do a similar thing to the snippets for versions and update it to something in keeping with the code in each repo.

I think the idea though is to show people what kind of code sample to provide, not how to use the specific lib.

emlaver: @ricellis Yea makes sense, let's keep it as pseudo-code.  Disregard the similar comments in the PRs of the other libraries.

PR #394, opened 2018-07-03
Stopped raising exception on conflict retry success

### LINKED ISSUE #393: document.update_field() raises 409 after successful retry

The `Document.update_field()` method raises an HTTPError(409) exception if it succeeds with its retries without exhausting the max retries.

Here is the code in `Document._update_field()` (see [here](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/document.py#L262)):

```
        # Attempt to save, retrying conflicts up to max_tries.
        try:
            self.save()
        except requests.HTTPError as ex:
            if tries < max_tries and ex.response.status_code == 409:
                self._update_field(
                    action, field, value, max_tries, tries=tries+1)
            raise
```

The `raise` statement should be in an `else:` clause if the previous `if` statement.

If you need a script that reproduces this, let me know, but it is too long to paste here (I think), and from looking at the code, the issue should be obvious.

The `Document.update_field()` method should not raise an HTTPError(409) exception if it ended up with a successful retry without exhausting its max retries. Only if the max retries are exhausted, should it raise the exception.

The description of the `Document.update_field()` method should document that.

There should be test cases that verify that case. :-)

### 3. What actually happened

See above.

## Environment details

    > 2.9.0
- Python version
    > 2.7.14

ricellis:
## Checklist

## Description

Stopped incorrect raising of exception after successful retry in `doc.update_field`.

Fixes #393 

## Approach

Moved the exception `raise` into an `else` block to prevent it being raised after successful execution of the recursive call into `_update_field`.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Added new tests for conflict retry cases:
- `test_update_field_maxretries` - checks that the exception is raised when `max_tries` is exhausted
- `test_update_field_success_on_retry` - checks that an exception is not raised when a retry succeeds

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [NEW] Add new view parameters, `stable` and `update`, as keyword arguments to `get_view_result`.
+- [FIXED] Case where an exception was raised after successful retry when using `doc.update_field`.
 
 # 2.9.0 (2018-06-13)
 
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -259,7 +259,8 @@ class Document(dict):
             if tries < max_tries and ex.response.status_code == 409:
                 self._update_field(
                     action, field, value, max_tries, tries=tries+1)
-            raise
+            else:
+                raise
 
     def update_field(self, action, field, value, max_tries=10):
         """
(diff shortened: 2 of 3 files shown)


andy-maier:
Great!! Thanks much for solving this.

PR #397, opened 2018-07-09
Add custom JSON encoder/decoder option to `Document` constructor

smithsz:
## Checklist

## Description

Allow a user to specify a custom JSON encoder and decoder for a `Document` type.

## Approach

Pass additional kwargs `encode` & `decode` in via the `Document` constructor.

See test for example usage.

## Schema & API Changes

Add additional kwargs to method call. No breaking API changes.

## Security and Privacy

No change.

## Testing

Adds `tests/unit/client_tests.py:ClientTests.test_constructor_with_url`.

## Monitoring and Logging

No change.

### DIFF
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -53,8 +53,10 @@ class Document(dict):
     :param database: A database instance used by the Document.  Can be
         either a ``CouchDatabase`` or ``CloudantDatabase`` instance.
     :param str document_id: Optional document id used to identify the document.
+    :param str encoder: Optional JSON encoder object.
+    :param str decoder: Optional JSON decoder object.
     """
-    def __init__(self, database, document_id=None):
+    def __init__(self, database, document_id=None, **kwargs):
         super(Document, self).__init__()
         self._client = database.client
         self._database = database
@@ -63,7 +65,8 @@ class Document(dict):
         self._document_id = document_id
         if self._document_id is not None:
             self['_id'] = self._document_id
-        self.encoder = self._client.encoder
+        self.encoder = kwargs.get('encoder') or self._client.encoder
+        self.decoder = kwargs.get('decoder') or json.JSONDecoder
 
     @property
     def r_session(self):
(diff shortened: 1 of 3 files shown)


src/cloudant/document.py
```diff
@@ -63,7 +63,8 @@ def __init__(self, database, document_id=None):
         self._document_id = document_id
         if self._document_id is not None:
             self['_id'] = self._document_id
-        self.encoder = self._client.encoder
+        self.decoder = kwargs.get('decoder') or self._client.decoder
```
emlaver: If the decoder is not configurable, can we remove `kwargs.get('decoder')`?

smithsz: It's only not configurable from the Cloudant client. We are allowing users to specify a custom decoder in the Document constructor.

A Client level decoder will be added in a future PR. It requires a little more work to ensure the decoder is passed around correctly.

tests/unit/document_tests.py
```diff
@@ -859,5 +861,49 @@ def test_document_request_fails_after_client_disconnects(self):
...
+
+        self.assertEquals(raw_doc['name'], 'julia')
+        self.assertEquals(raw_doc['dt']['_type'], 'datetime')
+        self.assertEquals(raw_doc['dt']['value'], '2018-07-09T15:11:10')
+
+        doc2 = Document(self.db, doc['_id'], decoder=DTDecoder)
+        doc2.fetch()
+
+        self.assertEquals(doc2['name'], 'julia')
+        self.assertTrue(isinstance(doc2['dt'], datetime))
```
ricellis: Can't this and the following assertion be replaced by:
```python
self.assertEquals(doc2['dt'], doc['dt'])
```
?

Since `doc` and `doc2` should both be using the object form and the `datetime` equality should work here.

src/cloudant/client.py
```diff
@@ -87,6 +87,7 @@ def __init__(self, user, auth_token, admin_party=False, **kwargs):
         self.server_url = kwargs.get('url')
         self._client_user_header = None
         self.admin_party = admin_party
+        self.decoder = json.JSONDecoder  # not currently configurable
```
ricellis: Since this already caused some confusion, perhaps append `at client scope` or similar.

src/cloudant/document.py
ricellis: Would be nice to document these parameters in the doc block.

PR #404, opened 2018-09-04
403 remove unnecessary request in Result iteration

### LINKED ISSUE #403: Unnecessary requests via cloudant.result.Result iteration

Use the couchdb.result.Result class to iterate a view or _all_docs that has more than zero items. Check the logs to notice that there are (at least) two requests.

A single request is issued when the number of rows is less than the page size.

### 3. What actually happened

Two requests are made.

## Environment details

All.

Logic is here:

https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/result.py#L359-L360

There needs to be something like:

```python
if len(result) < self._page_size:
    break
```

Alternatively you can just change the ```if result:``` conditional to something like such:

```python
for row in result:
    yield row
if len(result) < self._page_size:
    break
```

emlaver:
## Checklist

## Description
Two request are executed when only one request is required for getting a result collection that is less than the `page_size` value.

Fixes #403

## Approach

Break out of `while` loop in `Result.__iter__` when the length of `result` is less than the page size.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests.
- Verified that only one request is executed in local CouchDB logging

## Monitoring and Logging
- "No change"

### DIFF
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -341,7 +341,8 @@ class Result(object):
             raise ResultException(103, invalid_options, self.options)
 
         try:
-            if int(self._page_size) <= 0:
+            self._page_size = int(self._page_size)
+            if self._page_size <= 0:
                 raise ResultException(104, self._page_size)
         except ValueError:
             raise ResultException(104, self._page_size)
@@ -349,15 +350,17 @@ class Result(object):
         skip = 0
         while True:
             response = self._ref(
-                limit=int(self._page_size),
+                limit=self._page_size,
                 skip=skip,
                 **self.options
             )
             result = self._parse_data(response)
-            skip += int(self._page_size)
+            skip += self._page_size
             if result:
                 for row in result:
                     yield row
+                if len(result) < self._page_size:
+                    break
                 del result
             else:
                 break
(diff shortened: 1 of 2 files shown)


src/cloudant/result.py
```diff
@@ -341,11 +341,13 @@ def __iter__(self):
...
 
         try:
-            if int(self._page_size) <= 0:
+            self._page_size = int(self._page_size)
+            if self._page_size <= 0:
                 raise ResultException(104, self._page_size)
         except ValueError:
             raise ResultException(104, self._page_size)
 
+        self._page_size = int(self._page_size)
```
smithsz: Didn't you already cast this to an int on L344?

smithsz: Also, you don't need to call `int(self._page_size)` everywhere else right as `self._page_size` is an int now?

src/cloudant/result.py
```diff
@@ -358,6 +360,8 @@ def __iter__(self):
             if result:
                 for row in result:
                     yield row
+                if len(result) < self._page_size:
```
smithsz: We still do an unnecessary request if the total number of view results == `self._page_size` I think. I'm wondering if we can check total_rows against self._page_size too?

emlaver: I think it's safer for now to keep the double request when `results == page_size` than trying to use `total_rows`.

PR #405, opened 2018-09-11
Allow query parameters to be passed to custom changes filters

### LINKED ISSUE #402: Cannot set filter parameters in changes() call
The Feed._validate method makes it impossible to pass arbitrary query parameters to a _changes filter design document when calling CouchDatabase.changes().
https://github.com/cloudant/python-cloudant/blob/4435d84f24dd3563e4abd2929e2953eeba0d5b89/src/cloudant/feed.py#L114
If the key is not in arg_types, it should assume it is a custom parameter directed at the filter and hope. Maybe just check if it is an int, float, or str, as these are the only supported types for the query.

I suppose the same holds for Cloudant._db_updates().

smithsz:
## Checklist

## Description

Allow arbitrary query parameters to be passed to custom changes filters.

Fixes #402.

## Approach

Add new `ANY_ARG` and `ANY_TYPE` to list of valid changes arguments:

```py
_CHANGES_ARG_TYPES = {
    'conflicts': (bool,),
    'doc_ids': (list,),
    'filter': (STRTYPE,),
    'include_docs': (bool,),
    'style': (STRTYPE,),
    ANY_ARG: ANY_TYPE  # pass arbitrary query parameters to a custom filter
}
```

## Schema & API Changes

No change.

## Security and Privacy

No change.

## Testing

Includes additional unit test.

## Monitoring and Logging

No change.

### DIFF
--- a/src/cloudant/feed.py
+++ b/src/cloudant/feed.py
@@ -100,7 +100,7 @@ class Feed(object):
                 if isinstance(val, STRTYPE):
                     translation[key] = val
                 elif not isinstance(val, NONETYPE):
-                    arg_converter = TYPE_CONVERTERS.get(type(val))
+                    arg_converter = TYPE_CONVERTERS.get(type(val), json.dumps)
                     translation[key] = arg_converter(val)
             except Exception as ex:
                 raise CloudantArgumentError(115, key, ex)
--- a/src/cloudant/feed.py
+++ b/src/cloudant/feed.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (c) 2015, 2016 IBM. All rights reserved.
+# Copyright (c) 2015, 2018 IBM. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ import json
 
 from ._2to3 import iteritems_, next_, unicode_, STRTYPE, NONETYPE
 from .error import CloudantArgumentError, CloudantFeedException
-from ._common_util import feed_arg_types, TYPE_CONVERTERS
+from ._common_util import ANY_ARG, ANY_TYPE, feed_arg_types, TYPE_CONVERTERS
 
 class Feed(object):
     """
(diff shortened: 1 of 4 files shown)


tests/unit/changes_tests.py
```diff
@@ -465,14 +465,29 @@ def test_get_feed_using_doc_ids(self):
...
+            filter='mailbox/new_mail',
+            foo='bar',  # query parameters to a custom filter
+            include_docs=False
+        )
+        params = feed._translate(feed._options)
+        self.assertEquals(params['filter'], 'mailbox/new_mail')
+        self.assertEquals(params['foo'], 'bar')
+        self.assertEquals(params['include_docs'], 'false')
+
+    # def test_invalid_argument(self):
```
smithsz: This test is no longer valid as passing `foo='bar'` is now accepted under `ANY_ARG: ANY_TYPE`.

tests/unit/changes_tests.py
```diff
@@ -465,14 +465,20 @@ def test_get_feed_using_doc_ids(self):
         self.assertSetEqual(set([x['id'] for x in changes]), expected)
         self.assertTrue(str(feed.last_seq).startswith('100'))
 
-    def test_invalid_argument(self):
```
smithsz: This test is no longer valid as passing foo='bar' is now accepted under ANY_ARG: ANY_TYPE.

src/cloudant/feed.py
```diff
@@ -21,7 +21,7 @@
```
ricellis: This needs a copyright update

emlaver:
@smithsz There's a couple cases that cause a CloudantArgumentError `___ type is not callable`:
```
feed = Feed(self.db, foo={}) # type is dict
feed = Feed(self.db, foo=object_var) # type is object
```
We should probably only allow `int`, `float`, `string`, and `bool` types.

smithsz:
I think we should serialise `foo={}`. Obviously serialising `foo=object()` will never work and should throw.

How about making `json.dumps` the default type converter? (as per fb490ca712c017789358232d1118f8e517011120)

PR #408, opened 2018-10-11
WIP Partitioned Databases

smithsz:
## Checklist

- [ ] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes

## Description
This PR introduces a new feature, user-defined partitioned databases.

A new kind of database can be created with the `?partitioned=true` option. All documents within the database must have document ids of the following format;

`partition_name:doc_id`

both `partition_name` and `doc_id` must follow the couchdb id format (can't begin with _, etc).

All documents with the same partition_name are guaranteed to be mapped to the same shard range. When querying an index, the new `/db/_partition/$partition/_view` endpoint can query the view more efficiently, by only consulting the single shard range holding the partition. This is much more efficient and scales the same way that primary key lookup (`GET /dbname/docid`) does (approximately linearly).

## Approach
TODO

## Schema & API Changes
TODO

## Security and Privacy
TODO

## Testing
TODO

## Monitoring and Logging
TODO

## Examples

- Create partitioned database:
```py
db = self.client.create_database('partitioned_db_1', partitioned=True)
```

- Run partitioned query:
```py
result = db.partition('partition_key_a').query(selector={'name': {'$eq': 'foo'}})
```

- Run partitioned search:
```py
result = db.partition('partition_key_a').search('ddoc001', 'searchindex001', query='name:julia*')
```

- Run partitioned view:
```py
result = db.partition('partition_key_a').view('ddoc', 'my_view')
```

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -51,6 +51,8 @@ class CouchDatabase(dict):
     :param str database_name: Database name used to reference the database.
     :param int fetch_limit: Optional fetch limit used to set the max number of
         documents to fetch per query during iteration cycles.  Defaults to 100.
+    :param bool partitioned: Create as a partitioned database. Defaults to
+        ``False``.
     """
     def __init__(self, client, database_name, fetch_limit=100,
                  partitioned=False):
@@ -110,10 +112,11 @@ class CouchDatabase(dict):
 
     def get_partition(self, partition_key):
         """
-        Retrieve database partition object.
+        Retrieve a database partition object for the specified partition key.
 
-        :param partition_key: partition key as string.
-        :return: DatabasePartition object if database is partitioned, else None.
+        :param str partition_key: partition key.
+        :return: DatabasePartition object for specified partition key.
+        :rtype: `~cloudant.database_partition.DatabasePartition`
         """
         return DatabasePartition(self, partition_key)
 
@@ -1189,6 +1192,8 @@ class CloudantDatabase(CouchDatabase):
     :param str database_name: Database name used to reference the database.
     :param int fetch_limit: Optional fetch limit used to set the max number of
         documents to fetch per query during iteration cycles.  Defaults to 100.
+    :param bool partitioned: Create as a partitioned database. Defaults to
+        ``False``.
     """
     def __init__(self, client, database_name, fetch_limit=100,
                  partitioned=False):
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -119,7 +119,7 @@ class CouchDatabase(dict):
         """
         return self._partitioned
 
-    def partition(self, partition_key):
+    def get_partition(self, partition_key):
         """
         Retrieve database partition object.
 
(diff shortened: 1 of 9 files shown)


src/cloudant/database_partition.py
```diff
@@ -0,0 +1,96 @@
...
+        See :func:`~database.CouchDatabase.get_view_result`.
+
+        :return: The result.
+        """
+        return self._db.get_view_result(*args,
+                                        partition_key=self._partition_key,
+                                        **kwargs)
+
+
+class CloudantDatabasePartition(CouchDatabasePartition):
```
ricellis: Do we actually need this class? Or more specifically do we need the `CouchDatabasePartition` - can't we just have all the methods in one DatabasePartition class? I think the only time we need the `CouchDatabasePartition` is if someone is trying to use a server that has partition, but not search. Given that partition is headed for Couch 2.3+ they'd probably want support for other Couch 2.x APIs (even if they hadn't built search) so they'd almost certainly be using `CloudantDatabase` anyway because `CouchDatabase` is really "Couch1xDatabase".

Does that make sense?

src/cloudant/design_document.py
```diff
@@ -15,7 +15,7 @@
 """
```
ricellis: This file needs a copyright update

src/cloudant/index.py
```diff
@@ -44,13 +44,14 @@ class Index(object):
         :func:`~cloudant.database.CloudantDatabase.create_query_index`.
```
ricellis: This file needs a copyright update

src/cloudant/query.py
```diff
@@ -19,7 +19,7 @@
 import json
```
ricellis: This file needs a copyright update

src/cloudant/view.py
```diff
@@ -94,6 +94,7 @@ def __init__(
             view_name,
```
ricellis: This file needs a copyright update

src/cloudant/database.py
```diff
@@ -103,6 +109,27 @@ def creds(self):
...
+    @property
+    def partitioned(self):
+        """
+        Establish if this database is partitioned.
+
+        :return: True if database is partitioned, else False.
+        """
+        return self._partitioned
+
+    def partition(self, partition_key):
```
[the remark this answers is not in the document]
smithsz: How about just `CouchDatabase.get_partition` which returns a `cloudant.database_partition.DatabasePartition` object?

(as per ebb71bf)

smithsz:
To build sphinx docs locally:
```sh
sphinx-build -b html docs/ /tmp/python-cloudant-docs/ && open /tmp/python-cloudant-docs/index.html
```

src/cloudant/database.py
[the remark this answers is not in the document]
ricellis: I still think we should have the context manager if we are going down this route.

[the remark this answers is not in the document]
smithsz: Sure, I can add it in if you like. To my mind a CM is for managing a resource; ensuring things are closed or a lock is released when you're done with it etc. I can't really see the value here.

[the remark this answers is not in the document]
ricellis: You're right, we don't really need it.

PR #412, opened 2018-11-13
Replaced usage of requests `response.json()` method

### LINKED ISSUE #409: Document fetch fails when uses its default json decoder and module simplejson is present
## Bug Description
`Document.fetch()` fails when it uses its default json decoder and `simplejson` is in the environment. 

### Steps to reproduce and code sample to demonstrate the issue
To reproduce, just run a simple document fetch like this having `simplejson` package in the environment:
```
from cloudant import cloudant
from cloudant.database import CloudantDatabase

def fetch_by_id(usr, pwd,url,db_name,id):
  with cloudant(user=usr, passwd=pwd, url=url) as client:
    db = CloudantDatabase(client, db_name)
    doc = Document(db, id)
    if doc.exists():
        doc.fetch()
```

### Expected behavior
The full document should be retrieved by `fetch` function, instead when `simplejson` is reachable in the environment, it will raise the following error:
```
File \"/usr/local/lib/python3.6/site-packages/simplejson/__init__.py\", line 535, in loads
return cls(encoding=encoding, **kw).decode(s)
TypeError: __init__() got an unexpected keyword argument 'encoding'
```

### What actually happened
Starting in [this line](https://github.com/cloudant/python-cloudant/blob/2.10.0/src/cloudant/document.py#L170): `self.update(resp.json(cls=self.decoder))`:
1. `json()` function from `requests.models` is called with argument `cls` and value `json.JSONDecoder`, as `self.decoder` [was initialized](https://github.com/cloudant/python-cloudant/blob/2.10.0/src/cloudant/document.py#L69) by default when no decoder is passed to `Document`
2. a JSON decoder [is imported](https://github.com/requests/requests/blob/v2.20.0/requests/models.py#L42) selectively in `requests.compat` module: [when `simplejson` is present](https://github.com/requests/requests/blob/v2.20.0/requests/compat.py#L29), then the json decoder implementation comes from that package
3. the JSON decoder from `simplejson` package is used [to load the contents of the response](https://github.com/requests/requests/blob/v2.20.0/requests/models.py#L888)
4. now, argument `cls` is a `json.JSONDecoder` (see point 1), and it does not expect [an argument called `encoding`](https://github.com/simplejson/simplejson/blob/v3.16.0/simplejson/__init__.py#L535), thus a `TypeError` is raised. 

## Environment details
Version affected: `2.10.0`
Note that in version `2.9.0`, `cls=self.decoder` [was not passed to](https://github.com/cloudant/python-cloudant/commit/82017411beb12717780d036f7308086d8fe37ef5#diff-396eae97315a21001968a65d1ed17e61R171) `resp.json()`.

ricellis:
## Checklist

## Description

Replace usage of requests built-in `response.json()` method.

Fixes #409 

## Approach

Requests preferentially imports `simplejson` if it is present in the environment, overriding use of the system `json` module. This causes issues with our use of the default `json.JSONDecoder`. Since it is unclear whether requests will make a fix before version 3.x we need to workaround the problem.

Instead of calling `response.json()` we can call `json.loads(response.text)`, however, since later we may wish to return to `response.json()` (e.g. when requests 3.x is available) then it makes sense to do this in `_common_utils`.

* Replaced `response.json()` calls with `_common_util.response_to_json_dict()`.
* Added `CHANGES` entry.

## Schema & API Changes

- Clarified types of JSON encoder and decoder in doc strings.

## Security and Privacy

- "No change"

## Testing

The existing tests continue to pass and exercise the new code path.
I have not added any test specifically for the new method `response_to_json_dict()` as it is a one-line call to `json.loads()` and is called by many existing test cases.

- Modified existing tests that mocked `response.json()` to mock `response.text` instead.
- Added a new test axis that runs the existing tests with the `simplejson` module also installed in the environment.

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+- [FIXED] Unexpected keyword argument errors when using the library with the
+ `simplejson` module present in the environment caused by `requests` preferentially
+  loading it over the system `json` module.
+
 # 2.10.0 (2018-09-19)
 
 - [NEW] Add custom JSON encoder/decoder option to `Document` constructor.
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,11 +10,12 @@ def getEnvForSuite(suiteName) {
     case 'basic':
       envVars.add("RUN_BASIC_AUTH_TESTS=1")
       break
-    case 'cookie':
-      break
     case 'iam':
       // Setting IAM_API_KEY forces tests to run using an IAM enabled client.
       envVars.add("IAM_API_KEY=$DB_IAM_API_KEY")
+      break
+    case 'cookie':
+    case 'simplejson':
       break
     default:
       error("Unknown test suite environment ${suiteName}")
@@ -36,6 +37,7 @@ def setupPythonAndTest(pythonVersion, testSuite) {
             . ./tmp/bin/activate
             pip install -r requirements.txt
             pip install -r test-requirements.txt
+            ${'simplejson'.equals(testSuite) ? 'pip install simplejson' : ''}
             pylint ./src/cloudant
             nosetests -A 'not db or (db is "cloudant" or "cloudant" in db)' -w ./tests/unit --with-xunit
           """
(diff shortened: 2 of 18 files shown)


ricellis:
@smithsz @emlaver see bcaf416 for adding an axis that runs tests with `simplejson` present.

PR #415, opened 2018-12-12
414 decode content

### LINKED ISSUE #414: cloudant 2.10.1+ requests of large docs slower than requests.get(...).json() or curl
## Bug Description

Fetching some large documents from couch takes 20+s when the equivalent with `curl` or python `requests` takes 1s. It's a bug that was introduced in this PR: https://github.com/cloudant/python-cloudant/pull/412.

```
pip install cloudant
```

In python shell create a large document with non-ascii characters:
```
from cloudant.client import CouchDB
c = CouchDB('', '', url='http://localhost:5984', admin_party=True, connect=True)
# create large document with non ascii characters
x = c['my_database'].create_document({'_id': 'doc', 'value': u'好'  * 1000000})
```

Now test loading it with different versions of `cloudant` installed:

**2.10.0 is fast**
```
$ pip install cloudant==2.10.0
$ time python -c "from cloudant.client import CouchDB
c = CouchDB('', '', url='http://localhost:5984', admin_party=True, connect=True)
c['my_database']['doc']
"

real	0m0.365s
user	0m0.126s
sys	0m0.090s
```
**2.10.1 is slow**
```
$ pip install cloudant==2.10.1
$ time python -c "from cloudant.client import CouchDB
c = CouchDB('', '', url='http://localhost:5984', admin_party=True, connect=True)
c['my_database']['doc']
"

real	0m30.739s
user	0m29.667s
sys	0m0.314s
```

I expect it to be fast—about as fast as making the request directly with `requests.get(...).json()` or `curl`—and as fast as it is with `cloudant<=2.10.0`.

### 3. What actually happened

It takes 20+s to load, depending on the specific doc.

## Environment details
    > 2.10.1+
- Python version
    > 2.7

-------------

the cause is this PR: https://github.com/cloudant/python-cloudant/pull/412. The added slowness is concentrated entirely in `resp.text`, which does various things to infer the encoding, some of which are apparently expensive. The solution is to use `json.loads(resp.content.decode('utf-8'))` instead of `json.loads(resp.text)`.

ricellis:
## Checklist

## Description

Added `encoding='utf-8'` to responses to avoid slow `chardet` path.

Fixes #414 

## Approach

Check if the repsonse encoding is unset and if so set it to `utf-8`.
(Unrelated) Removed unnecessary`pass` statements causing lint failures.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- No new tests because this path is covered by existing tests. We don't have any performance tests where we could automatically validate that this fix works or check for regressions.

## Monitoring and Logging

- "No change"

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -288,6 +288,8 @@ def response_to_json_dict(response, **kwargs):
 
     :returns: dict of JSON response
     """
+    if response.encoding is None:
+        response.encoding = 'utf-8'
     return json.loads(response.text, **kwargs)
 
 # Classes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- [FIXED] A performance regression deserializing JSON in version 2.10.1.
+
 # 2.10.1 (2018-11-16)
 
 - [FIXED] Unexpected keyword argument errors when using the library with the
--- a/src/cloudant/_client_session.py
+++ b/src/cloudant/_client_session.py
@@ -95,12 +95,14 @@ class ClientSession(Session):
         """
         No-op method - not implemented here.
         """
+        # pylint: disable=unnecessary-pass
         pass
 
     def logout(self):
         """
         No-op method - not implemented here.
         """
+        # pylint: disable=unnecessary-pass
         pass
 
 


src/cloudant/_common_util.py
```diff
@@ -288,6 +288,8 @@ def response_to_json_dict(response, **kwargs):
 
     :returns: dict of JSON response
     """
+    if response.encoding is None:
```
smithsz: Perhaps a brief line in the doc string too, something like:
> If the response is encoded with an ASCII based encoding other than UTF-8 (e.g. latin-1), then ensure the appropriate `response.encoding` is set.

PR #428, opened 2019-02-10
Document.atomic context manager

### LINKED ISSUE #427: document.Document context manager ignore exceptions
## Bug Description

`document.Document` context manager ignores exceptions in inner block, making costly manual housekeeping/rollback necessary in case of errors. Using try/except blocks is indeed feasible but still expensive as I'd have to keep a pristine copy in memory (and I'd get a new `_rev` anyway because `save()` in `__exit__`).

```python
import cloudant

client = cloudant.CouchDB('user', 'pass',
                          url='http://localhost:5984', connect=True)
db = client['db']


db.create_document({'_id': 'julia006', 'name': 'julia', 'age': 6})

with cloudant.document.Document(db, 'julia006') as doc:

    doc['age'] = 7
    0 / 0
    doc['prop'] = 'value'

# doc will be partly updated:
# {'_id': 'julia006',
#  '_rev': '2-492dc56da45c97bf6b17ff85e8d83f9d',
#  'name': 'julia', 'age': 7}

```

Document not updated because an error in context manager inner block.

### 3. What actually happened

Document get updated with all the changes made until the error happened.

aogier:
## Checklist

## Description
This PR implements a `Document.atomic` context manager that don't save a document if there is any uncatched exception in inner code block. Tentatively fixes #427.

## Approach

As this change could be breaking (in fact it change a default behaviour) I've decided to implement a brand new context manager using python's contextlib.
When it's needed, one could use `atomic` as a context manager:
```python
with Document(self.db, 'julia006').atomic() as doc:
```
Context Manager implementation follow current one, in fact it uses the [same dunders](https://github.com/aogier/python-cloudant/commit/00c339f635ff5c777c4685d48d1887ebae10de4d#diff-396eae97315a21001968a65d1ed17e61R352).

## Schema & API Changes

- Added a new `Document.atomic` context manager

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - unit.document_tests.test_document_context_manager_atomic_creation_failure
    - unit.document_tests.test_document_context_manager_atomic_update_failure

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 - [IMPROVED] Updated `Getting started` section with a `get_query_result` example. 
 - [FIXED] Fixed parameter type of `selector` in docstring.
+- [NEW] `Document.atomic` context manager.
 
 # 2.11.0 (2019-01-21)
 
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -18,6 +18,7 @@ API module/class for interacting with a document in a database.
 import json
 import requests
 from requests.exceptions import HTTPError
+from contextlib import contextmanager
 
 from ._2to3 import url_quote, url_quote_plus
 from ._common_util import response_to_json_dict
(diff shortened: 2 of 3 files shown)


smithsz:
Thanks again for this. We've decided to change the default behaviour. I've made the change in `.__exit__` and stolen your tests (see #434).

PR #429, opened 2019-02-10
Document.save on create/change only

aogier:
## Checklist

## Description

Trivial change that commit to remote db only when creating or actually changing local data.

Current behaviour:

```python
doc = Document(self.db, 'http://example.com')
doc.save()
doc.save()
# doc _rev is now 2-
```

New behaviour:

```python
doc = Document(self.db, 'http://example.com')
doc.save()
doc.save()
# doc _rev is still 1-
```

## Approach

Implemented a simple boolean check that flip upon Document's dict update methods calls. Document will still be created even if empty, but save will only happen when an actual change is made, minimizing data transfers.

## Schema & API Changes

- `Document.save` only when creating or local copy changed.

## Security and Privacy

- "No change"

## Testing

- Added new tests:
    - tests.unit.document_tests.test_not_update_unchanged_document

## Monitoring and Logging

- "No change"

### DIFF
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -67,6 +67,7 @@ class Document(dict):
             self['_id'] = document_id
         self.encoder = kwargs.get('encoder') or self._client.encoder
         self.decoder = kwargs.get('decoder') or json.JSONDecoder
+        self.dirty = True
 
     @property
     def r_session(self):
@@ -163,8 +164,9 @@ class Document(dict):
             raise CloudantDocumentException(101)
         resp = self.r_session.get(self.document_url)
         resp.raise_for_status()
-        self.clear()
+        self.clear(keep_revision=False)
         self.update(response_to_json_dict(resp, cls=self.decoder))
+        self.dirty = False
 
     def save(self):
         """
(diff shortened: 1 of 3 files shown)


aogier:
forgot pop, popitem and update. Also, the way `update` works make it impractical to implement/prone to false positive

src/cloudant/document.py
```diff
@@ -356,6 +360,7 @@ def __setitem__(self, key, value):
         if key == '_id':
             self._document_id = value
         super(Document, self).__setitem__(key, value)
+        self.dirty = True
```
smithsz: What if a user has edited the doc using `.update(d2)`, `.pop(key)`, `.clear()` etc?

smithsz: I fear we might need to store a hash of the contents and compare each time. Or override all these methods too ofc.

aogier: I'm currently working on overriding MutableMapping methods and also observe static methods stuff, I hope I can push something later today :)

aogier:
Ok I think I'm ready for a code review. Tests are passing and all seems promising but I'd definitely like another couple of eyes :)

aogier:
meh, I forgot `setdefault`...

aogier:
added `setdefault`

aogier:
I've reintroduced _doc_id logic on some dict methods I'll fix it asap

aogier:
_document_id logic re-removed from code

smithsz:
Thanks for all your recent work. It is really appreciated. We’ve discussed this change as a team and feel it’s adding too much complexity to the `Document` class. I’ve tried a different implementation; instead I compute/compare a hash of the document on each save ([here](https://github.com/cloudant/python-cloudant/compare/wip-skip-doc-save-for-noop-updates)). This change is less invasive but likely a more expensive solution.
We’ve decided the library shouldn't offer this feature. If a user is calling `.save()` multiple times without updating the document then that’s their decision.
Thanks again and have a good weekend.

aogier:
Hi, it seems a good conclusion to me and as a couchdb noob I'll take this design choice as an implementation hint especially on where to put my ctx manager. Thank you for your time and keep up the good work :)

ciao

PR #432, opened 2019-02-11
Fixing test_update_field_success_on_retry

### LINKED ISSUE #431: wrong unit test test_update_field_success_on_retry
## Bug Description

Hi, I'd like to point out that `test_update_field_success_on_retry` uses a wrong mock implementation. In fact in order to invalidate test you could place a [fetch call](https://github.com/cloudant/python-cloudant/compare/master...aogier:feature/test_update_field_success_on_retry?expand=1#diff-5426ff62a320d421a5fdd141988a7eabR538) just before asserts, but the entire linked push should be clear enough if you care to give it a look.  

While I'm still not capable to produce a fix, I could point out that a mock should work with [callables and not called functions](https://github.com/cloudant/python-cloudant/blob/00b6cbb4e297477be214b7202d627bb99bc29223/tests/unit/document_tests.py#L528) so while I'm studying for a fix please be aware of that.

Checkout and run [this commit](https://github.com/aogier/python-cloudant/tree/feature/test_update_field_success_on_retry).

Test passing :P

aogier:
## Checklist

## Description
Fixed a document test mock
Fixes #431 

## Approach

Reimplemented a mock so doc.save() gets called during update_field and not mock construction/evaluation. Tests were passing because `update_field` did not called save() at all, thus making it difficult to spot.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified `test_update_field_success_on_retry`

## Monitoring and Logging

- "No change"

### DIFF
--- a/tests/unit/document_tests.py
+++ b/tests/unit/document_tests.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import cloudant
 """
 _document_tests_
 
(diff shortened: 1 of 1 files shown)


tests/unit/document_tests.py
```diff
@@ -523,9 +524,18 @@ def test_update_field_success_on_retry(self):
         # Mock when saving the document
         # 1st call throw a 409
         # 2nd call delegate to the real doc.save()
-        with mock.patch('cloudant.document.Document.save',
-                        side_effect=[requests.HTTPError(response=mock.Mock(status_code=409, reason='conflict')),
-                                     doc.save()]) as m_save:
```
smithsz: Oh I see, the .save() was happening before the mock was even initialised. I guess this bad test properly failed in your pr #429 since the save didn't execute given there are no local document changes at this point.

smithsz: I guess `doc.save` (not`doc.save()`) would be less wrong. But that would just call the mock again.

smithsz:
I'm +1. I think it's pretty clear what that new side_effect def is doing.

The only other thing I can think of doing it mocking the second save with `mock.DEFAULT` and just asserting there were two calls to `.save()`.

@emlaver what do you reckon?

tests/unit/document_tests.py
aogier: I've tried this way but the problem is side_effect as a list raise exceptions or return values. So in this case it would effectively return the callable doc.save instead of running it. And of course, because how the patch is made yes, it would loop.

smithsz: Yeah, just read the docs. I assumed wrong.

PR #434, opened 2019-02-12
Stop document context manager from saving if an error is raised.

smithsz:
## Checklist

## Description

Fix bug where document CM performed save in event of uncaught exception.

## Schema & API Changes

No change.

## Security and Privacy

No change.

## Testing

Includes 2 additional unit tests.

## Monitoring and Logging

No change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,9 @@
 # Unreleased
 
-- [IMPROVED] Updated `Getting started` section with a `get_query_result` example. 
+- [FIXED] Bug where document context manager performed remote save despite
+  uncaught exceptions being raised inside `with` block.
 - [FIXED] Fixed parameter type of `selector` in docstring.
+- [IMPROVED] Updated `Getting started` section with a `get_query_result` example.
 
 # 2.11.0 (2019-01-21)
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,8 @@ import shlex
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #sys.path.insert(0, os.path.abspath('.'))
 
+sys.path.insert(0, os.path.abspath('../src'))
+
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -441,41 +441,45 @@ multiple updates to a single document. Note that we don't save to the server
 after each update. We only save once to the server upon exiting the ``Document``
 context manager.
 
- .. code-block:: python
+.. warning:: Uncaught exceptions inside the ``with`` block will prevent your
+             document changes being saved to the remote server. However, changes
+             will still be applied to your local document object.
 
-    from cloudant import cloudant
-    from cloudant.document import Document
+.. code-block:: python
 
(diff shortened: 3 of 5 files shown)


src/cloudant/document.py
```diff
@@ -341,12 +341,13 @@ def __enter__(self):
 
         return self
 
-    def __exit__(self, *args):
+    def __exit__(self, exc_type, exc_value, traceback):
         """
-        Support context like editing of document fields.  Handles context exit
-        logic.  Executes a Document.save() upon exit.
+        Support context like editing of document fields. Handles context exit
+        logic. Executes a `Document.save()` upon exit iff no exception occurred.
```
emlaver: iff -> if

tests/unit/document_tests.py
```diff
@@ -622,6 +622,47 @@ def test_document_context_manager_no_doc_id(self):
...
+        doc['age'] = 6
+        doc.save()
+
+        # Make a document update and then raise an error.
+        with self.assertRaises(ZeroDivisionError), Document(self.db, 'julia006') as doc:
+            doc['age'] = 7
+            raise ZeroDivisionError()
+
+        # Assert the change persists locally.
+        self.assertEqual(doc['age'], 7)
```
smithsz: Note, if we raise an exception in the block then the local copy still gets updated. It's only the remote save that gets skipped.

src/cloudant/document.py
smithsz: I was abbreviating _if and only if_ but I guess plain old "if" works and will cause less confusion.

tests/unit/document_tests.py
emlaver: I wonder if there's a way for us to use a temporary dict during the context manager that only updates the local doc and save remotely if there's no exception.
If not, we should add a note in the context manager docs that if an exception occurs, the remote doc is not saved but the local doc is.

smithsz: I'll add a further note in the docs. I think a temp dict could add too many complications.

docs/getting_started.rst
```diff
@@ -441,41 +441,45 @@ multiple updates to a single document. Note that we don't save to the server
 after each update. We only save once to the server upon exiting the ``Document``
 context manager.
 
- .. code-block:: python
+.. warning:: Uncaught exceptions inside the ``with`` block will prevent your
```
smithsz: I've just added this warning. The other changes in this file are for correctly indenting the following code blocks.

CHANGES.md
```diff
@@ -1,7 +1,8 @@
 # Unreleased
 
-- [IMPROVED] Updated `Getting started` section with a `get_query_result` example. 
+- [FIXED] Bug where document CM performed save in event of uncaught exception.
```
emlaver: We should spell out context manager

docs/conf.py
```diff
@@ -21,6 +21,8 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #sys.path.insert(0, os.path.abspath('.'))
 
+sys.path.insert(0, os.path.abspath('../src'))
```
emlaver: I've had to use this in the past to get documentation builds to properly run.  I don't think it's needed for building in `readthedocs.org`.

smithsz: Yeah. I always have to add it in too. I'll leave it and hopefully it'll make building locally easier.

PR #435, opened 2019-02-13
Removed document._document_id logic

### LINKED ISSUE #430: dict methods (update, clear, pop*) breaks `_id` <-> '_document_id' internal logic
## Bug Description

Not overriden dict methods are breaking Document class logic in regards to keep `_id` key in sync with `_document_id` class property. As this breaks my work I'm currently implementing a fix and to me the most sensible way is to completely remove this feature, using `_id` property when needed.  
If I correctly understand the code, this is only used in cloudant.document module and that logic seems completely removable. What do you think? TIA

In this example:
```python
import cloudant
from uuid import uuid4

client = cloudant.CouchDB('user', 'pass',
                          url='http://localhost:5984', connect=True)

db = client.create_database('db-' + str(uuid4()))
db.delete()
db.create()


doc = db.create_document({'_id': 'julia006'})
doc['_id'] = 'julia007'
doc.save()
print('_id:')
print('doc:', doc)
print(doc.document_url)

doc = db.create_document({'_id': 'julia008'})
doc.update({'_id': 'julia009'})
doc.save()
print('update:')
print('doc:', doc)
print(doc.document_url)

# db.delete()
```

produce this output:
```
_id:
doc: {'_id': 'julia007', '_rev': '1-967a00dff5e02add41819138abb3284d'}
http://localhost:5984/db-28328934-87a6-4434-ab96-f8cb54351e2d/julia007
update:
doc: {'_id': 'julia009', '_rev': '2-7051cbe5c8faecd085a3fa619e6e6337'}
http://localhost:5984/db-28328934-87a6-4434-ab96-f8cb54351e2d/julia008
```

As you can see, using `__setitem__` correctly produce a new document with an url coherent with `doc._id`. Using `update` instead make a new revision on an `doc._id` mismatching url.

Every dict updating methods (so also `pop`, `popitem`, `clear`, `setdefault`) should work the same way as setitem/delitem dunders.

### 3. What actually happened

Only setitem/delitem keeps ids in sync

aogier:
## Checklist

## Description

Removing _document_id logic so dict's updating methods work by default.
Fixes #430 

## Approach

Removed _document_id logic from Document class, now everything works based in `_id` property instead.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
   uncaught exceptions being raised inside `with` block.
 - [FIXED] Fixed parameter type of `selector` in docstring.
 - [IMPROVED] Updated `Getting started` section with a `get_query_result` example.
+- [FIXED] Removed internal `Document._document_id` property to allow a safe use of
+  dict's methods.
 
 # 2.11.0 (2019-01-21)
 
--- a/src/cloudant/document.py
+++ b/src/cloudant/document.py
@@ -63,9 +63,8 @@ class Document(dict):
         self._database = database
         self._database_host = self._client.server_url
         self._database_name = database.database_name
-        self._document_id = document_id
-        if self._document_id is not None:
-            self['_id'] = self._document_id
+        if document_id:
+            self['_id'] = document_id
         self.encoder = kwargs.get('encoder') or self._client.encoder
         self.decoder = kwargs.get('decoder') or json.JSONDecoder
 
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 - [IMPROVED] Updated `Getting started` section with a `get_query_result` example. 
 - [FIXED] Fixed parameter type of `selector` in docstring.
+- [IMPROVED] Removed internal `Document._document_id`property so dict's
```
emlaver: add a space: `Document._document_id`property ->  `Document._document_id` property

emlaver: I'm wondering if we should write something like:
```
[FIXED] Removed internal `Document._document_id` logic to keep dict' s`_id` field in sync with `_document_id` class parameter.
```
@smithsz wdyt?

smithsz: yeah, mark it as a `[FIXED]`.

aogier: done, also rephrased the changelog entry so it's more clear

smithsz:
👍 Great, thanks @aogier! Can you squash these commits together please so we can merge your work.

aogier:
here you go! Thanks

PR #437, opened 2019-03-05
paginate by startkey

### LINKED ISSUE #436: proposal: a faster result iteration

## Bug Description

Iterating over a large dataset is sometimes slow. This is not a library disfunction but is related to how couchdb works, ie.

```sh
# take an acceptable time
SKIP=0
time curl -sH 'Content-Type: application/json' \
    http://localhost:5984/db/_design/dd/_view/vv\?limit\=500\&skip\=$SKIP\&group\=true

# this will take ages
SKIP=820000
time curl -sH 'Content-Type: application/json' \
    http://localhost:5984/db/_design/dd/_view/vv\?limit\=500\&skip\=$SKIP\&group\=true

```

So this is inherently true for your library when using current `Result` class logic. 
I've noticed that the different pagination approach [manual propose](https://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method) does not incur in this slowdown.

I drafted a switch implementation that can allow to iterate based on `startkey` instead of `skip` and you can [see it here](https://github.com/cloudant/python-cloudant/compare/master...aogier:feature/faster-iterator?expand=1), as you see is quite simple and yes is badly coded :grimacing: because is a PoC and I left original implementation under the new alternative.

What do you think? This is maybe somehow related to #389 or not, anyway for me it's a complete game changer that I'd like to have upstream.

Thank you, regards

aogier:
## Checklist

## Description

Make `Result`'s iteration faster paginating by `startkey` instead of `skip`.

Fixes #436 

## Approach

As per #436 , paginating by `skip` could be super-slow. This patch fix the problem by using `startkey` so every iteration run in constant time.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [IMPROVED] Updated `Result` iteration by paginating with views' `startkey` and
+  queries' `bookmark`.
 - [FIXED] Bug where document context manager performed remote save despite
   uncaught exceptions being raised inside `with` block.
 - [FIXED] Fixed parameter type of `selector` in docstring.
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -15,6 +15,7 @@
 """
 API module for interacting with result collections.
 """
+from functools import partial
 from ._2to3 import STRTYPE
 from .error import ResultException
 from ._common_util import py_to_couch_validate, type_or_none
@@ -327,16 +328,16 @@ class Result(object):
     def __iter__(self):
         """
         Provides iteration support, primarily for large data collections.
-        The iterator uses the ``skip`` and ``limit`` options to consume
-        data in chunks controlled by the ``page_size`` option.  It retrieves
-        a batch of data from the result collection and then yields each
-        element.
+        The iterator uses the ``startkey``, ``startkey_docid``, and ``limit``
+        options to consume data in chunks controlled by the ``page_size``
+        option. It retrieves a batch of data from the result collection
+        and then yields each element.
 
         See :class:`~cloudant.result.Result` for Result iteration examples.
 
         :returns: Iterable data sequence
         """
-        invalid_options = ('skip', 'limit')
+        invalid_options = ('limit', )
         if any(x in invalid_options for x in self.options):
             raise ResultException(103, invalid_options, self.options)
 
@@ -347,21 +348,60 @@ class Result(object):
         except ValueError:
             raise ResultException(104, self._page_size)
 
-        skip = 0
+        init_opts = {
+            'skip': self.options.pop('skip', None),
+            'startkey': self.options.pop('startkey', None)
+        }
+
+        self._call = partial(self._ref,  #pylint: disable=attribute-defined-outside-init
+                             limit=self._real_page_size,
+                             **self.options)
+
+        response = self._call(**{k: v
+                                 for k, v
+                                 in init_opts.items()
+                                 if v is not None})
+
+        return self._iterator(response)
+
+    @property
+    def _real_page_size(self):
+        '''
+        In views we paginate with N+1 items per page.
+        https://docs.couchdb.org/en/stable/ddocs/views/pagination.html#paging-alternate-method
+        '''
+        return self._page_size + 1
+
+    def _iterator(self, response):
+        '''
+        Iterate through view data.
+        '''
+
         while True:
-            response = self._ref(
-                limit=self._page_size,
-                skip=skip,
-                **self.options
-            )
             result = self._parse_data(response)
-            skip += self._page_size
             if result:
+                doc_count = len(result)
+                last = result.pop()
                 for row in result:
                     yield row
-                if len(result) < self._page_size:
+
+                # We expect doc_count = self._page_size + 1 results, if
+                # we have self._page_size or less it means we are on the
+                # last page and need to return the last result.
+                if doc_count < self._real_page_size:
+                    yield last
                     break
                 del result
+
+                # if we are in a view, keys could be duplicate so we
+                # need to start from the right docid
+                if last['id']:
+                    response = self._call(startkey=last['key'],
+                                          startkey_docid=last['id'])
+                # reduce result keys are unique by definition
+                else:
+                    response = self._call(startkey=last['key'])
+
             else:
                 break
 
@@ -510,3 +550,32 @@ class QueryResult(Result):
         query result JSON response content
         """
         return data.get('docs', [])
+
+    @property
+    def _real_page_size(self):
+        '''
+        During queries iteration page size is user-specified
+        '''
+        return self._page_size
+
+    def _iterator(self, response):
+        '''
+        Iterate through query data.
+        '''
+
+        while True:
+            result = self._parse_data(response)
+            bookmark = response.get('bookmark')
+            if result:
+                for row in result:
+                    yield row
+
+                del result
+
+                if not bookmark:
+                    break
+
+                response = self._call(bookmark=bookmark)
+
+            else:
+                break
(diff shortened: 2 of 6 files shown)


src/cloudant/result.py
```diff
@@ -347,24 +347,27 @@ def __iter__(self):
...
+                startkey=last,
                 **self.options
             )
             result = self._parse_data(response)
-            skip += self._page_size
             if result:
+                last = result.pop()
                 for row in result:
                     yield row
                 if len(result) < self._page_size:
```
smithsz: Change to `if len(result) < self._page_size + 1`?

Also, I think a comment here would be good. Something like:

> We expect `len(result) = self._page_size + 1` results, if we have ` self._page_size` or less it means we are on the last page and need to return the last result.

aogier:
I've found a bug while implementing comments requests: when a view's key is an integer, view does not return because startkey='' apparently want to cast to string:

```sh
# this does work
c /temp-cloudant/_design/dd/_view/vv\?limit=3                
{
  "total_rows": 6,
  "offset": 0,
  "rows": [
    {
      "id": "f41a9664608b060ce1a488ed96005a35",
...

# this does not

c /temp-cloudant/_design/dd/_view/vv\?limit=3\&startkey\='""'
{
  "total_rows": 6,
  "offset": 6,
  "rows": []
}

```

src/cloudant/result.py
aogier: you're right, this way we save a hit on the worst case when ndocs%pagesize==0 ... neat !

aogier:
Ok I've changed a bit the approach so we use `startkey` in GET if and only if user specifies one. I've implemented this via a partial object (I like them so much, so elegant). Also, added the comment and avoid an empty hit if we are at the view's end!

aogier:
I'm working on tests to cover the integer key corner case

aogier:
added a test that would not have passed in 19a4abe078f9ba55083ed44fc5baddf18b45437c but pass now in 15d216d46fb68a0f6e7b43089a73a5a6d4324996 ... reverse TDD :+1:

aogier:
have to write more tests

aogier:
ok I've added another test for observing pagination behaviour and now everything seems pretty stable and robust to me. @smithsz thank you for letting me know what you think :)

ciao

src/cloudant/result.py
```diff
@@ -327,7 +328,7 @@ def _handle_result_by_key_slice(self, key_slice):
     def __iter__(self):
         """
         Provides iteration support, primarily for large data collections.
-        The iterator uses the ``skip`` and ``limit`` options to consume
+        The iterator uses the ``startkey`` and ``limit`` options to consume
```
smithsz: and ``startkey_docid``

tests/unit/result_tests.py
```diff
@@ -643,5 +645,43 @@ def test_iteration_no_data(self):
         result = Result(self.view001, startkey='ruby')
         self.assertEqual([x for x in result], [])
 
+    def test_iteration_integer_keys(self):
+        """
+        Test that iteration works as expected when keys are integer.
+        """
+        result = Result(self.view007)
```
smithsz: can you put a small page size in here? maybe `page_size=10`

src/cloudant/result.py
aogier: ah ok, I've left it back on purpose viewing it as an "hidden feature" but I'll amend docs

tests/unit/result_tests.py
aogier: sure

tests/unit/result_tests.py
```diff
@@ -643,5 +645,43 @@ def test_iteration_no_data(self):
...
+        """
+
+        class CallMock:
+            last_call = None
+
+            def __init__(self, outer):
+                self.outer = outer
+
+            def call(self, *args, **kwargs):
+                self.outer.assertNotEqual((args, kwargs), self.last_call,
```
smithsz: This is just asserting that the current view query isn't the same as the previous right? How about checking against a list of expected stuff, i.e.
```
call 1: {'limit': 28}
call 2: {'limit': 28, 'startkey': 1, 'startkey_docid': 'julia027'}
call 3: {'limit': 28, 'startkey': 1, 'startkey_docid': 'julia054'}
call 4: {'limit': 28, 'startkey': 1, 'startkey_docid': 'julia081'}
```

smithsz:
This is a great. I've been doing some testing with other options (i.e. descending=True) and everything works nicely.
I notice there's a check in there for `limit` and `skip` params not being supported. I guess this is still true for `limit` but we should be able to use `skip` now right? It'll want removing from the query params after the first call though. What do you think?

tests/unit/result_tests.py
aogier: It's indeed more strict, I like it

aogier:
Seems a good idea to me, it opens up to more use cases and it's totally manageable in the current implementation. 

BTW, I think we could also reason about allowing `limit`: we use `limit` in the request context for efficiently paginating over a result set, but from the user standpoint a `limit` clause mean "I want to iterate over this set and get at most N total results" and this could totally be feasible by servin count(results) < limit. What do you think? Just a couple cents about another iteration eh :)

I'll begin to implement for a skip param!

src/cloudant/result.py
```diff
@@ -347,21 +348,47 @@ def __iter__(self):
         except ValueError:
             raise ResultException(104, self._page_size)
 
-        skip = 0
+        initial_options = {option: value
```
smithsz: Not sure about this. Bit tricky to read. How about...
```py
        _call = partial(self._ref, limit=self._page_size + 1, **self.options)

        init_opts = {
            'skip': self.options.pop('skip', None),
            'startkey': self.options.pop('startkey', None)
        }

        response = _call(**{k: v for k, v in init_opts.items() if v is not None})
```

aogier: ok, it's more concise and definitely easier to read!

aogier:
did the last changes and also added a changelog entry!

aogier:
queries are to be paginated via `bookmark`, I'm adding more tests

aogier:
ok so basically I was abusing nose's filtering options and skipped query tests, now there is a possibile implementation on which I decided to stay straight and implement a specific inner method for queries and views as they are so different in schema. Thank you, regards

tests/unit/query_result_tests.py
```diff
@@ -428,21 +428,11 @@ def test_iteration_with_invalid_options(self):
         Test that iteration raises an exception when "skip" and/or "limit" are
```
emlaver: Need to fix this comment now that tests with skip are removed.

emlaver:
For tests to run under Python 3, this will need to be rebased to include #438 which fixes the `pylint` warning.

tests/unit/query_result_tests.py
aogier: you're right!

aogier:
ok I've fixed test docstring and rebased on 2eb1261838d02b5ee63d4a572095a78bcfcee825, were jenkins failing because of pylint?

emlaver:
@aogier Yes, the Python 3 tests were showing the pylint warning.

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [IMPROVED] Faster `Result` iterator.
```
emlaver: We should probably give some detail and not use 'Faster'
i.e. `Improve Result iteration by paginating with 'startkey'`.  @smithsz wdyt?

aogier:
Great !

I've squashed it and modified the changelog entry:
```
- [IMPROVED] Updated `Result` iteration by paginating with views' `startkey` and
  query's `bookmark`.
```

Thank you, regards

PR #440, opened 2019-03-08
slimming Result iterator

aogier:
## Checklist

## Description

Hello, this PR implements a deflating FIFO Result iterator. To me, this is especially useful when I run large multiprocessing ingestion batches where having as little allocated ram as possible give a better scalability. While we could obtain somethin similar reverse()ing lists, manual states that `collection` types are [more efficient](https://docs.python.org/2/library/collections.html#deque-objects).

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

## Monitoring and Logging

- "No change"

### DIFF
--- a/src/cloudant/result.py
+++ b/src/cloudant/result.py
@@ -15,6 +15,7 @@
 """
 API module for interacting with result collections.
 """
+from collections import deque
 from functools import partial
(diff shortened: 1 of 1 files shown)


aogier:
rebased against #437

PR #441, opened 2019-03-19
Partitioned databases

smithsz:
## Checklist

## Description

Add partitioned databases feature.

See [partitioned_databases.rst](https://github.com/cloudant/python-cloudant/blob/63f58a4b219da63cc945771db8d07725f95af1b8/docs/partitioned_databases.rst).

## Schema & API Changes
New API (non-breaking).

## Security and Privacy
No change.

## Testing
Includes additional testing.

## Monitoring and Logging
No change.

### DIFF
--- a/src/cloudant/database.py
+++ b/src/cloudant/database.py
@@ -26,9 +26,9 @@ from ._common_util import (
     SEARCH_INDEX_ARGS,
     SPECIAL_INDEX_TYPE,
     TEXT_INDEX_TYPE,
+    TYPE_CONVERTERS,
     get_docs,
-    response_to_json_dict,
-    )
+    response_to_json_dict)
 from .document import Document
 from .design_document import DesignDocument
 from .security_document import SecurityDocument
@@ -50,13 +50,17 @@ class CouchDatabase(dict):
     :param str database_name: Database name used to reference the database.
     :param int fetch_limit: Optional fetch limit used to set the max number of
         documents to fetch per query during iteration cycles.  Defaults to 100.
+    :param bool partitioned: Create as a partitioned database. Defaults to
+        ``False``.
     """
-    def __init__(self, client, database_name, fetch_limit=100):
+    def __init__(self, client, database_name, fetch_limit=100,
+                 partitioned=False):
         super(CouchDatabase, self).__init__()
         self.client = client
         self._database_host = client.server_url
         self.database_name = database_name
         self._fetch_limit = fetch_limit
+        self._partitioned = partitioned
         self.result = Result(self.all_docs)
 
     @property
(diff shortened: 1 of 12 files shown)


src/cloudant/index.py
```diff
@@ -40,18 +40,21 @@ class Index(object):
         Index.
     :param str design_document_id: Optional identifier of the design document.
     :param str name: Optional name of the index.
+    :param bool partitioned: Optional. Create as a partitioned index. Defaults
+        to ``False``.
```
ricellis: Just to check - this implies that we don't follow the server default of `true` if the database is partitioned and `false` otherwise. I'm OK with that as long as it is clear and it looks like the code matches up, but I just wanted to call out the subtle behaviour difference.

src/cloudant/design_document.py
```diff
@@ -39,11 +39,17 @@ class DesignDocument(Document):
         either a ``CouchDatabase`` or ``CloudantDatabase`` instance.
     :param str document_id: Optional document id.  If provided and does not
         start with ``_design/``, it will be prepended with ``_design/``.
+    :param bool partitioned: Optional. Create as a partitioned design document.
+        Defaults to ``False``.
```
ricellis: Again this doesn't match the server side default (which I believe is variable depending on the type of the DB). That's OK, but I wanted to call it out.

src/cloudant/database.py
```diff
@@ -105,6 +109,18 @@ def creds(self):
             "user_ctx": session.get('userCtx')
         }
 
+    def database_partition_url(self, partition_key):
+        """
+        Get database partition URL.
```
emlaver: `Get the URL of the database partition.`

src/cloudant/index.py
smithsz: Ah, yes. I overlooked this subtlety.

The default is `False`. The database object is passed to the index constructor so I could get the `partitioned` value and use that as a default if preferred. But if you're happy then I'll leave it alone. The current implementation matches java-cloudant too.

ricellis: We probably need to more explicitly note that the behaviour is different to avoid any confusion.
I think most "real production" use cases aren't creating indexes and design docs on the fly, they are version controlled and deployed so as to avoid index re-builds and so on, so I don't think it is critical that this matches the server as long as we are clear about what is happening.

emlaver:
Does it make sense to have `Partitioned Databases` in its own separate doc section and not under `Getting Started`?
Screen Shot 2019-03-21 at 12 10 33 PM

smithsz:
@emlaver Moved under the 'Databases' heading...

Screenshot 2019-03-25 at 15 03 25

src/cloudant/index.py
smithsz: Updated docstring for extra clarity:
> Defaults to ``False`` for both partitioned and non-partitioned databases.

PR #447, opened 2019-05-16
Raise exception when database name is invalid

### LINKED ISSUE #446: HTTP Error 404

## Bug Description
Keep getting HTTP Error 404 when trying to populate an instance. I have looked into a similar issue posted but I didn't seem to get a solution out of the issue post.

Here is a simplified version of my code

```
from datetime import datetime, date, timedelta
import json

# Connect with Cloudant
client = Cloudant(USERNAME, PASSWORD, url=URL, connect = True, auto_renew = True)
dayData =[{'Date/Time': '2019-05-01 00:00', 'Temp (°C)': 4.0, 'Dew Point Temp (°C)': -7.1}, 
  {'Date/Time': '2019-05-01 01:00', 'Temp (°C)': 4.3, 'Dew Point Temp (°C)': -8.2}] 
#Just a way to get a date in dbName
dbName = "weatherData_"+str(datetime.strptime(dayData[0]['Date/Time'],'%Y-%m-%d %H:%M').date())
db = client.create_database(dbName)
time.sleep(0.2)
# for each of the values in the dayData list upload as document
for i in range(len(dayData)):
    with Document(db, dayData[0]['Date/Time'] ) as document:
        # The JSON String is posted in Data
        document["data"] = json.dumps(dayData[i])
        time.sleep(0.2)
```

This is a simplified version of my code that is suppose to create a database named after a specific date, and then uses that created database and uploads a data string to it.

### 3. What actually happened
I keep on getting this error. 

```
HTTPError: 404 Client Error: Object Not Found not_found Database does not exist. for url: https://####-####.cloudantnosqldb.appdomain.cloud/weatherData_2019-05-01
```

Ive tested all the logic in my original code, and works if i set the name to a something but if i have a dynamic name I keep getting the same error. I even put in time.sleep for request limiting

## Environment details
    > 2.0.2
- Python version
    > 3.6.2

emlaver:
## Checklist

## Description

Update `create_database` to raise exception when database name is invalid.
Fixes #446 

## Approach
Added additional `if` to raise `CloudantDatabaseException` when status code is `400`.

## Schema & API Changes
 "No change"

## Security and Privacy
 "No change"

## Testing
Added new test `test_create_invalid_database_name`

## Monitoring and Logging
"No change"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- [FIXED] Correctly raise exceptions from `create_database` calls.
+
 # 2.12.0 (2019-03-28)
 
 - [NEW] Added partitioned database support.
--- a/src/cloudant/client.py
+++ b/src/cloudant/client.py
@@ -290,6 +290,7 @@ class CouchDB(dict):
         except CloudantDatabaseException as ex:
             if ex.status_code == 412:
                 raise CloudantClientException(412, dbname)
+            raise ex
         super(CouchDB, self).__setitem__(dbname, new_db)
         return new_db
 
--- a/tests/unit/client_tests.py
+++ b/tests/unit/client_tests.py
@@ -34,7 +34,8 @@ from cloudant import cloudant, cloudant_bluemix, couchdb, couchdb_admin_party
 from cloudant._client_session import BasicSession, CookieSession
 from cloudant.client import Cloudant, CouchDB
 from cloudant.database import CloudantDatabase
-from cloudant.error import CloudantArgumentError, CloudantClientException
+from cloudant.error import (CloudantArgumentError, CloudantClientException,
+                            CloudantDatabaseException)
 from cloudant.feed import Feed, InfiniteFeed
 from nose.plugins.attrib import attr
 from requests import ConnectTimeout, HTTPError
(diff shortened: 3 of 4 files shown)


src/cloudant/client.py
```diff
@@ -290,6 +290,8 @@ def create_database(self, dbname, partitioned=False, **kwargs):
         except CloudantDatabaseException as ex:
             if ex.status_code == 412:
                 raise CloudantClientException(412, dbname)
+            if ex.status_code == 400:
```
ricellis: I don't think we should special case `400` here, I think the problem is that `412` is special cased without an `else` suppressing all the other exceptions. For example if the service returns a `401` or a `500` we should still be raising here. So I think what we really want is:
```python
else:
  raise ex
```

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+- [FIXED] Raise exception in `create_database` when database name is invalid.
```
ricellis: Based on my recommended change I think this would need to be updated to something more general like:
`- [FIXED] Correctly raise exceptions from `create_database` calls.`

tests/unit/client_tests.py
```diff
@@ -398,6 +399,17 @@ def test_create_existing_database(self):
...
+        """
+        Test creation of database with an invalid name
+        """
+        dbname = 'invalidDbName_'
+        self.client.connect()
+        with self.assertRaises(CloudantDatabaseException) as cm:
+            self.client.create_database(dbname, throw_on_exists=False)
+        self.assertEqual(cm.exception.status_code, 400)
+        self.client.disconnect()
+
```
ricellis: We probably should have an additional test that returns some other response code too. It could use a mock.

src/cloudant/client.py
```diff
@@ -290,6 +290,8 @@ def create_database(self, dbname, partitioned=False, **kwargs):
...
+                raise ex
```
smithsz: Why do we wrap `new_db.create` in a try/catch here? It looks like we're suppressing 500s and 401s too.

src/cloudant/client.py
smithsz: Oops, I missed this earlier. Didn't refresh the page. I'm agreement; we need to catch all failures here.

emlaver: We have custom messages for `CloudantDatabaseException` and `CloudantClientException` in `_messages.py`.  If we want to catch and throw 500 under a database exception I'll add the message `The server encountered an unexpected condition which prevented it from fulfilling the request.` to the messages module.

emlaver:
@smithsz @ricellis `CloudantDatabaseException` needed to be updated to handle status codes that do not exist in `_messages.py`.  If not, the Cloudant exception would throw a 100 default error code for an HTTP 500 error which can cause confusion.

PR #453, opened 2019-10-15
Fix broken Python 2 build

bessbd:
## Checklist

- [X] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [ ] Added tests for code changes _or_ test/build only changes
- [X] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [X] Completed the PR template below:

## Description

f2c1298 accidentally broke the build as collections.abc is only available in Python 3.x.

This commit adds a workaround that enables the import that works both on 2 and 3.

## Schema & API Changes
"No change"

## Testing
No new tests because the change is not semantic.

## Monitoring and Logging
"No change"

### DIFF
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -19,16 +19,15 @@ throughout the library.
 
 import sys
 import platform
-import collections
 import json
 
 from ._2to3 import LONGTYPE, STRTYPE, NONETYPE, UNITYPE, iteritems_
 from .error import CloudantArgumentError, CloudantException, CloudantClientException
 
 try:
-    Sequence = collections.abc.Sequence  # noqa
+    from collections.abc import Sequence
 except AttributeError:
-    Sequence = collections.Sequence  # noqa
+    from collections import Sequence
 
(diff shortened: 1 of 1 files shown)


src/cloudant/_common_util.py
```diff
@@ -19,12 +19,17 @@
...
 import platform
-from collections.abc import Sequence
+import collections
 import json
 
 from ._2to3 import LONGTYPE, STRTYPE, NONETYPE, UNITYPE, iteritems_
 from .error import CloudantArgumentError, CloudantException, CloudantClientException
 
+try:
+    Sequence = collections.abc.Sequence  # noqa
```
ricellis: seems like this causes another linter issue; maybe
`sequence = collections.abc.Sequence` will be better, though of course requires some further changes.

bessbd:
@ricellis , @emlaver : Thank you for the review!
I'm about to merge this PR.

PR #460, opened 2020-02-19
Intermittent read timeout

ricellis:
## Checklist

## Description

Adjust read timeout in intermittently failing test and update test matrices.

Run `tests.unit.replicator_tests.ReplicatorTests.test_timeout_in_create_replication`

Test to pass.

### 3. What actually happened

It intermittently fails.

## Approach

Change the timeout from `.5` to `(30,5)` - this means the connect timeout stays at the default 30s instead of also being reduced to .5s and the read timeout is increased from .5 seconds to 5 s. Whilst this means the test takes longer to run, it is much more reliable.

## Schema & API Changes

- "No change"

## Security and Privacy

- "No change"

## Testing

- Modified existing tests because to change timeout to resolve intermittent failure.
- Modified test matrix to remove Python 2 as it is EOL.
- Modified test matrix for CouchDB version
    *  `1.7.1` -> `1.7.2` (although Couch 1.x is EOL, we'll still run these tests until the next major I think)
        * Had to fix `tests.unit.client_tests.ClientTests.test_create_invalid_database_name` as a result of this change
    * `2.1.1` -> `2.3.1`
        * Had to fix test set up special casing version as a result of this change.

## Monitoring and Logging

- "No change"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_script:
 # command to run tests
 script:
     - pylint ./src/cloudant
-    - nosetests -A 'not db or ((db is "couch" or "couch" in db) and (not couchapi or couchapi <='${COUCHDB_VERSION:0:1}'))' -w ./tests/unit
+    - nosetests -A 'not db or ((db == "couch" or "couch" in db) and (not couchapi or couchapi <='${COUCHDB_VERSION:0:1}'))' -w ./tests/unit
 
 notifications:
   email: false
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -3,14 +3,13 @@ sudo: required
 language: python
 
 python:
-  - "2.7"
```
emlaver: +1 for removing 2.7 now that's it's EOL.

PR #463, opened 2020-04-14
Proposed 2.13.0 release

bessbd:
# 2.13.0 (2020-04-14)

- [FIXED] Correctly raise exceptions from `create_database` calls.
- [FIXED] Fix "DeprecationWarning" from "collections"


CHANGES.md
```diff
@@ -1,6 +1,7 @@
-# Unreleased
+# 2.13.0 (2020-04-14)
```
ricellis, line 1: Sorry didn't review this yesterday :)
Maybe update it for today, but it's not a big deal

CHANGES.md
```diff
@@ -1,6 +1,7 @@
...
 
 - [FIXED] Correctly raise exceptions from `create_database` calls.
+- [FIXED] Fix "DeprecationWarning" from "collections"
```
ricellis, line 4: nits:
* needs an ending period/full stop `.`
* I'd prefer to see DeprecationWarning and collections, marked up as monospace rather than double-quoted.

bessbd:
> +1, but a couple of nits that I think are worth addressing

Thank you for the review, @ricellis . I've just pushed 5f01eeb that does the changes you've requested. Let me know what you think.

ricellis:
LGTM, let's give the CI a chance to run, if that test failure pops up again we may need to include a fix for it before we do the release (otherwise the release will just get blocked by the CI later).

PR #470, opened 2020-05-12
Make `create_query_index` always supply `partitioned`

### LINKED ISSUE #468: create_query_index() failing on unpartitioned CouchDB database with message invalid_partitioned_option

## Bug Description
On trying to create query index in CouchDB 3.0.0, the request fails with an unpartitioned DB with message `HTTPError: 400 Client Error: Bad Request invalid_partitioned_option`

Execute the following code
```
from cloudant import couchdb

with couchdb("admin", "admin", url="http://127.0.0.1:5984") as couchClient:
        _ = couchClient.create_database("foo")
        
with couchdb("admin", "admin", url="http://127.0.0.1:5984") as couchClient:
    db = couchClient["foo"]
    ddoc = DesignDocument(db, document_id="bar")
    ddoc["language"] = "query"
    ddoc.save()

with couchdb("admin", "admin", url="http://127.0.0.1:5984") as couchClient:
    db = couchClient["foo"]
    db.create_query_index(design_document_id="_design/bar",fields=["key"])
```
The index should get created
### 3. What actually happened
Error is thrown with message - `HTTPError: 400 Client Error: Bad Request invalid_partitioned_option Requested partitioned option does not match existing value on design document _design/bar for url: http://127.0.0.1:5984/foo/_index`
## Environment details

- CouchDB - 3.0.0
- Python 3.7
- python-cloudant - 2.13

## 4. Workaround
In the **cloudant/index.py**, partitioned option was set only for `partitioned = True`. I added an else clause for `partition = False` and that fixed the issue.
Updating line 157-158 of **cloudant/index.py**
```
if self._partitioned:
  payload['partitioned'] = True
else:
  payload['partitioned'] = False
```

bessbd:
## Checklist

## Description
The issue is described in #468 .
`create_query_index` does not always supply `partitioned` which is
always required by the service.

## Approach
This PR changes the behavior that so that `partitioned` is now always
present in the `create_query_index` requests.

## Schema & API Changes
~~- "No change"~~

EDIT: Two default `False` parameters have been changed to `None`. The users should not notice any difference.

## Security and Privacy
- "No change"

## Testing

- Added new test case

## Monitoring and Logging
- "No change"

This fixes #468

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # UNRELEASED
 
 - [FIXED] Set default value for `partitioned` parameter to false when creating a design document.
+- [FIXED] Always set `partitioned` for `create_query_index` requests
 
 # 2.13.0 (2020-04-16)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # UNRELEASED
 
 - [FIXED] Set default value for `partitioned` parameter to false when creating a design document.
-- [FIXED] Always set `partitioned` for `create_query_index` requests
+- [FIXED] Corrected setting of `partitioned` flag for `create_query_index` requests.
 
 # 2.13.0 (2020-04-16)
 
--- a/src/cloudant/index.py
+++ b/src/cloudant/index.py
@@ -154,8 +154,7 @@ class Index(object):
         self._def_check()
         payload['index'] = self._def
 
-        if self._partitioned:
-            payload['partitioned'] = True
+        payload['partitioned'] = bool(self._partitioned)
 
         headers = {'Content-Type': 'application/json'}
         resp = self._r_session.post(
--- a/src/cloudant/index.py
+++ b/src/cloudant/index.py
@@ -47,7 +47,7 @@ class Index(object):
         :func:`~cloudant.database.CloudantDatabase.create_query_index`.
     """
 
-    def __init__(self, database, design_document_id=None, name=None, partitioned=False, **kwargs):
+    def __init__(self, database, design_document_id=None, name=None, partitioned=None, **kwargs):
         self._database = database
         self._r_session = self._database.r_session
         self._ddoc_id = design_document_id
@@ -154,7 +154,8 @@ class Index(object):
         self._def_check()
         payload['index'] = self._def
 
-        payload['partitioned'] = bool(self._partitioned)
+        if self._partitioned is not None:
+            payload['partitioned'] = bool(self._partitioned)
 
         headers = {'Content-Type': 'application/json'}
         resp = self._r_session.post(
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -284,8 +284,12 @@ when constructing the new ``DesignDocument`` class.
     ddoc.add_view('myview','function(doc) { emit(doc.foo, doc.bar); }')
     ddoc.save()
 
-Similarly, to define a partitioned Cloudant Query index you must set the
-``partitioned=True`` optional.
+
+To define a partitioned Cloudant Query index you may set the
+``partitioned=True`` optional, but it is not required as the index will be
+partitioned by default in a partitioned database. Conversely, you must
+set the ``partitioned=False`` optional if you wish to create a global
+(non-partitioned) index in a partitioned database.
 
 .. code-block:: python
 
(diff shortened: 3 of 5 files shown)


src/cloudant/index.py
```diff
@@ -47,7 +47,7 @@ class Index(object):
         :func:`~cloudant.database.CloudantDatabase.create_query_index`.
     """
 
-    def __init__(self, database, design_document_id=None, name=None, partitioned=False, **kwargs):
+    def __init__(self, database, design_document_id=None, name=None, partitioned=None, **kwargs):
```
ricellis, line 50: I think you'll need to update the [docs](https://python-cloudant.readthedocs.io/en/stable/getting_started.html#creating-design-documents) to match the behaviour.

>Similarly, to define a partitioned Cloudant Query index you must set the partitioned=True optional.

I think we didn't actually honour this behaviour before - i.e. we didn't write `False` and the server side default was `true` in a partitioned database so it wasn't actually a requirement to set `partitioned=True` (even though I think that was the intention†).

† - IIRC the intention was to provide a stable default value (`False`) client-side rather than following the server-side approach where the default changed depending on whether the database is partitioned or not.

bessbd: The docs seem correct - as in it looks like it reflects the current behavior.
Would you like me to add that `partitioned=False` needs to be explicitly set for unpartitioned cases?

ricellis: Is this truth table accurate?

| Client-side value | Partitioned Database | Resulting index is partitioned |
| --- | --- | --- |
`None` | `False` | `False`
`None` | `True` | `True`
`False` | `False` | `False`
`False` | `True` | `False`
`True` | `False` | `400`
`True` | `True` | `True`

If so then doesn't the second row contradict the docs because you can get a partitioned index without setting `partitioned=True` in one case?

bessbd: I was only considering this from a `python-cloudant` perspective, where a specified `partitioned=False` was not passed to the service because of https://github.com/cloudant/python-cloudant/pull/470/files#diff-eeec48d80ce666f16517361d09cf63f1L157 , but that is required according to https://docs.couchdb.org/en/stable/api/database/find.html#post--db-_index , so a "third" boolean state is necessary ( https://github.com/cloudant/python-cloudant/pull/470/files#diff-eeec48d80ce666f16517361d09cf63f1R157 ) the "high impedance" None, because some versions of the service return a 400 for a specified `partitioned`: https://travis-ci.org/github/cloudant/python-cloudant/jobs/686082038#L426 .

> If so then doesn't the second row contradict the docs because you can get a partitioned index without setting `partitioned=True` in one case?

My understanding as per https://docs.couchdb.org/en/stable/api/database/find.html#post--db-_index is that `partitioned` always has to be specified in recent versions of CouchDB.

Do you have any specific doc changes in mind? That should help me understand what I'm missing.

ricellis: What I was trying to say is that your change means that in the case where the database _is_ partitioned and the user _does not_ specify a value for `partitioned` in python-cloudant the resulting index will still be created as partitioned (because of the server-side default).
The documentation states that:
>you **must** set the partitioned=True

But that "must" does not apply in the case described. I understand the confusion because the code didn't actually work as the doc described before (i.e. your change preserves the existing behaviour for this particular case).

The docs statement:
`Similarly, to define a partitioned Cloudant Query index you must set the partitioned=True optional.`
needs to change to something like:
>To define a partitioned Cloudant Query index you may set the `partitioned=True` optional, but it is not required as the index will be partitioned by default in a partitioned database. Conversely, you must set the `partitioned=False` optional if you wish to create a global (non-partitioned) index in a partitioned database.

bessbd: Added the lines verbatim in f6c9146dd627683cd418394cd0667bd0f64b29a2 . Is this what you've meant, @ricellis ?

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 # UNRELEASED
 
 - [FIXED] Set default value for `partitioned` parameter to false when creating a design document.
+- [FIXED] Always set `partitioned` for `create_query_index` requests
```
ricellis, line 4: I think this could be a little confusing for a user because it could imply that we're always setting `partitioned=True` and nit for a trailing period/full stop.
How about:
```
- [FIXED] -  Corrected setting of `partitioned` flag for `create_query_index` requests.
```

bessbd: Done in 675de0d .

bessbd:
Thank you for the review, @ricellis ! I'm about to merge this once CI passes.

PR #474, opened 2020-05-26
Override dict's get method

dominickj-tdi:
This allows the get method to be able to work with remote databases. Currently calling get only returns a doc if it is in the cache. The original behavior is preserved by use of the `remote` parameter, consistent with other overridden methods such as `keys()`, therefore this should be a non-breaking change.

## Checklist

- [ ] Added tests for code changes _or_ test/build only changes

## Description
Currently, calling the `get()` method on a database object returns None, even for documents I know exist in the database. This was because the method had not been overridden, therefore was only returning documents from the cache. With this change, it should now be possible to use the `get()` method on databases, rather than try/except blocks.

## Approach

The code simply wraps a try/except block over the `__getitem__ ()` when the `remote` paramaeter is True, and falls back to the original behavior by calling super if it is False.

## Schema & API Changes

Updated the `CouchDatabase.get()` method, adding a `remote` parameter for that method

## Security and Privacy

No Change

## Testing

Existing tests should suffice for this change, because it makes no large changes, it is only a shorthand for calling an existing method.

## Monitoring and Logging

No Change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 - [FIXED] Set default value for `partitioned` parameter to false when creating a design document.
 - [FIXED] Corrected setting of `partitioned` flag for `create_query_index` requests.
+- [NEW] Override the `get()` method for CouchDatabase, allowing it to retrive remote docuemnts if the `remote` parameter is specified.
 
 # 2.13.0 (2020-04-16)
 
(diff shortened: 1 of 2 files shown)


mojito317:
@dominickj-tdi we appreciate your implementation. 

We had to extend it with some tests, so I brought your commits into #505.
Now you can find your code as part of 436eddb. I'm closing this PR.

PR #477, opened 2020-08-14
Make setup.py compatible with python 2

bessbd:
## Checklist

- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [x] Added tests for code changes _or_ test/**build only changes**
- [x] Updated the change log file (`CHANGES.md`) _or_ test/build only changes
- [x] Completed the PR template below:

## Description

This fixes https://github.com/cloudant/python-cloudant/issues/476

## Approach
See related https://stackoverflow.com/a/10971047/5323166

PR #478, opened 2020-08-14
Remove Python 2 compatibility

bessbd:
## Checklist

## Description
Python 2 was EoLed back in January. This commit removes it from the list of supported environments.


bessbd:
> Updates also needed here:
> https://github.com/cloudant/python-cloudant/blob/d080687766a3101b1e876957674983b956851a76/setup.py#L57
> 
> I'm not sure if PyPi does anything to prevent installs on 2 if that ^ information changes. Given 2.14 _will_ work with python 2 because of #477 - then I guess this change should potentially only merge after that release, such that the tagged README etc for 2.14 matches it's capability and then everything on master after that does not guarantee compatibility. Does that make sense?

Done in 7182ef4

Holding off with this makes sense.

bessbd:
Thank you for the review, @ricellis and @emlaver . I'm about to merge this.

PR #487, opened 2020-11-20
Fix result paging for grouped view queries

### LINKED ISSUE #456: KeyError: 'id' on iterating over grouped view

>    for item in db.get_view_result(ddoc_id, view_name, group=True): print item

Iterate over all grouped reductions.

### 3. What actually happened

Iterate over grouped reductions, followed by:

>  File "/usr/local/lib/python2.7/dist-packages/cloudant/result.py", line 400, in _iterator
>    if last['id']:
>KeyError: 'id'

## Environment details
    > 2.12.0
- Python version
    > 2.7.17

### 4. Simple fix

Change the offending line to:
>    if 'id' in last:

emlaver:
## Checklist

## Description
When querying a grouped view where either a) there's more documents than the default page size or b) there's more documents that the user-defined page size option, a `KeyError` occurs.  This happens when the view data iteration logic tries to retrieve the next set of results.

When querying a grouped view where pagination is needed, a `KeyError` occurs when trying to grab the next page.  This error is happening because the iteration logic is expecting an `id` field which does not exist for grouped view results.

fixes #456 

## Approach
Update [line 400 in `cloudant/result.py`](https://github.com/cloudant/python-cloudant/blob/master/src/cloudant/result.py#L400) to check that the `id` field exists.

## Testing
Added new test `DatabaseTests.test_retrieve_grouped_view_result_with_page_size`.

### DIFF
--- a/tests/unit/database_tests.py
+++ b/tests/unit/database_tests.py
@@ -408,6 +408,34 @@ class DatabaseTests(UnitTestDbBase):
         self.assertIsInstance(rslt, Result)
         self.assertEqual(rslt[:1], rslt['julia099'])
 
+    def test_retrieve_grouped_view_result_with_page_size(self):
+        """
+        Test retrieving Result wrapped output from a design document grouped view
+        that uses a custom page size
+
+        The view used here along with group=True will generate rows of
+        data where each key will be grouped into groups of 2.  Such as:
+        {'key': 0, 'value': 2},
+        {'key': 1, 'value': 2},
+        ...
+        """
+        map_func = 'function(doc) {\n emit(Math.floor(doc.age / 2), 1); \n}'
+        data = {'_id': '_design/ddoc01','views': {'view01': {"map": map_func, "reduce": "_count"}}}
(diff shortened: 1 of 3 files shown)


tests/unit/database_tests.py
```diff
@@ -408,6 +408,34 @@ def test_retrieve_view_results(self):
...
+        self.populate_db_with_documents(5)
+
+        rslt = self.db.get_view_result(
+            '_design/ddoc01',
+            'view01',
+            group=True,
+            page_size=1)
+        self.assertIsInstance(rslt, Result)
+        i = 0
+        for doc in rslt:
```
ricellis, line 434: nit, for clarity it would be preferable to call this `row` instead of `doc` since it's a reduce.

ricellis (dismissed):
+1, 1 nit to take or leave

eiri (approved):
lgtm

eiri (approved):

mojito317 (approved):

PR #497, opened 2021-05-10
DeprecationWarning fix for method_whitelist

app63:
FIX: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead

## Checklist

## Description

DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 - [FIXED] Fixed the documentation for `bookmarks`.
 - [FIXED] Also exit `follow_replication` for `failed` state.
 - [FIXED] Fixed result paging for grouped view queries.
+- [FIXED] Replaced a depricated method `method_whitelist` with `allowed_methods`
 
 # 2.14.0 (2020-08-17)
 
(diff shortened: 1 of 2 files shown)


ricellis:
Thanks for this. Since our `urllib3` dependency version is transitive via `requests` I think we need to wait for them to move [their minimum](https://github.com/psf/requests/blob/f6d43b03fbb9a1e75ed63a9aa15738a8fce99b50/setup.py#L46) to `1.26.0` before we can accept this change, at which point we'd also need to update [our `requests` minimum version](https://github.com/cloudant/python-cloudant/blob/33781463ff85c8bb535bd6ede2049e4d3bec6c4c/requirements.txt#L1) to match.

Otherwise we risk using `allowed_methods` with a version of `urllib3` that doesn't support it. We could specify a `urllib3` version here, but I'm reluctant to add a constraint that could get out of step with `requests`.

ricellis:
Closing; `requests` haven't yet updated to a sufficiently recent version of `urllib3` and this library is now EOL.

PR #501, opened 2021-06-01
Use custom encoder (if provided) for all view `key` params not just `keys`

mojito317:
## Checklist

## Description

Fixes https://github.com/cloudant/python-cloudant/issues/221#issuecomment-752424499

## Approach

Used custom encoder at `_py_to_couch_translate` for `endkey`, `key`, and `startkey` parameters.

## Schema & API Changes

`endkey`, `key` and `startkey` parameters will be encoded with the custom encoder.

## Security and Privacy

- No change

## Testing

Test that all_docs uses the custom encoder. Fixed `test_valid_endkey`, `test_valid_key` and `test_valid_startkey ` with long type.

## Monitoring and Logging

- No change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 - [FIXED] Fixed result paging for grouped view queries.
 - [FIXED] Incorrect use of username as account name in `Cloudant.bluemix()`.
 - [IMPROVED] Documented use of None account name and url override for `Cloudant.iam()`.
-- [FIXED] Use custom encoder on get_docs params.
+- [FIXED] Use custom encoder (if provided) for all view `key` params not just `keys`.
 
 # 2.14.0 (2020-08-17)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - [FIXED] Fixed result paging for grouped view queries.
 - [FIXED] Incorrect use of username as account name in `Cloudant.bluemix()`.
 - [IMPROVED] Documented use of None account name and url override for `Cloudant.iam()`.
+- [FIXED] Use custom encoder on get_docs params.
 
 # 2.14.0 (2020-08-17)
 
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -209,7 +209,9 @@ def _py_to_couch_translate(key, val, encoder=None):
     try:
         if key in ['keys', 'endkey_docid', 'startkey_docid', 'stale', 'update']:
             return {key: val}
-        elif key in ['endkey', 'key', 'startkey']:
+        elif key in ['endkey', 'key', 'startkey'] and type(val) is not (int or LONGTYPE or bool):
+            # if type(val) == list:
+            #     return {key: json.dumps(list(val), cls=encoder)}
             return {key: json.dumps(val, cls=encoder)}
         if val is None:
             return {key: None}
--- a/src/cloudant/_common_util.py
+++ b/src/cloudant/_common_util.py
@@ -29,6 +29,8 @@ try:
 except ImportError:
     from collections import Sequence
 
+ENCODER = None
+
 # Library Constants
 
 USER_AGENT = '/'.join([
@@ -76,12 +78,12 @@ RESULT_ARG_TYPES = {
 
 # pylint: disable=unnecessary-lambda
 TYPE_CONVERTERS = {
-    STRTYPE: lambda x: json.dumps(x),
-    str: lambda x: json.dumps(x),
-    UNITYPE: lambda x: json.dumps(x),
-    Sequence: lambda x: json.dumps(list(x)),
-    list: lambda x: json.dumps(x),
-    tuple: lambda x: json.dumps(list(x)),
+    STRTYPE: lambda x: json.dumps(x, cls=ENCODER),
+    str: lambda x: json.dumps(x, cls=ENCODER),
+    UNITYPE: lambda x: json.dumps(x, cls=ENCODER),
+    Sequence: lambda x: json.dumps(list(x), cls=ENCODER),
+    list: lambda x: json.dumps(x, cls=ENCODER),
+    tuple: lambda x: json.dumps(list(x), cls=ENCODER),
     int: lambda x: x,
     LONGTYPE: lambda x: x,
     bool: lambda x: 'true' if x else 'false',
@@ -160,7 +162,7 @@ def feed_arg_types(feed_type):
         return _COUCH_DB_UPDATES_ARG_TYPES
     return _CHANGES_ARG_TYPES
 
-def python_to_couch(options):
+def python_to_couch(options, encoder=None):
     """
     Translates query options from python style options into CouchDB/Cloudant
     query options.  For example ``{'include_docs': True}`` will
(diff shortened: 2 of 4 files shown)


src/cloudant/_common_util.py
```diff
@@ -160,7 +162,7 @@ def feed_arg_types(feed_type):
         return _COUCH_DB_UPDATES_ARG_TYPES
     return _CHANGES_ARG_TYPES
 
-def python_to_couch(options):
+def python_to_couch(options, encoder=None):
     """
     Translates query options from python style options into CouchDB/Cloudant
```
emlaver, line 167: I think we'll want to add a comment about the new `encoder` parameter.

emlaver (commented):
Looks good, we may want to add a test to assert the custom encoder while calling `all_docs` (e.g. `test_all_docs_get_uses_custom_endoer` in `tests/unit/database_tests.py`)

src/cloudant/_common_util.py
mojito317: Thanks! I moved it to the database tests in f5610d5.

src/cloudant/_common_util.py
```diff
@@ -177,7 +177,7 @@ def python_to_couch(options):
     translation = dict()
```
emlaver, line 177: It doesn't allow me to comment on line 174 but we should add something like:
```
:param encoder: Custom encoder, defaults to None
```

emlaver (approved):

src/cloudant/_common_util.py
mojito317: Thanks! I added in https://github.com/cloudant/python-cloudant/pull/501/commits/88a1703c9609c4f1d3e7b3fa1f23040088e41417

src/cloudant/_common_util.py
```diff
@@ -160,7 +160,7 @@ def feed_arg_types(feed_type):
         return _COUCH_DB_UPDATES_ARG_TYPES
     return _CHANGES_ARG_TYPES
 
-def python_to_couch(options):
+def python_to_couch(options, encoder=None):
```
ricellis, line 163: AFAICT all `get_docs` calls have an encoder supplied which should in turn get here via https://github.com/cloudant/python-cloudant/blob/105fe2055562adbeb7c5ae41f97712794bf8716b/src/cloudant/_common_util.py#L252

However, I noticed that the view call to `get_docs` was slightly different in using the design document's encoder rather than the client one. 
https://github.com/cloudant/python-cloudant/blob/105fe2055562adbeb7c5ae41f97712794bf8716b/src/cloudant/view.py#L238

I think this is correct because if the design document has an encoder set it will be used in preference to the client one, but otherwise fall back to the client encoder if unset (via https://github.com/cloudant/python-cloudant/blob/105fe2055562adbeb7c5ae41f97712794bf8716b/src/cloudant/document.py#L68).

I think the caveat here is that someone who set an encoder specifically on a document (and no client level encoder) would also want to set that encoder on their design document before making view requests to guarantee the same encoder was used for the fields in the documents and the keys in the view query.

I don't think there is a problem compared to current behaviour but I wanted to call attention to it in case anyone had other thoughts.

ricellis (commented):
Also needs a `CHANGES` entry

src/cloudant/_common_util.py
```diff
@@ -201,14 +202,16 @@ def py_to_couch_validate(key, val):
...
-def _py_to_couch_translate(key, val):
+def _py_to_couch_translate(key, val, encoder=None):
     """
     Performs the conversion of the Python parameter value to its CouchDB
     equivalent.
     """
     try:
         if key in ['keys', 'endkey_docid', 'startkey_docid', 'stale', 'update']:
             return {key: val}
+        if key in ['endkey', 'key', 'startkey'] and not isinstance(val, (int, LONGTYPE, bool)):
```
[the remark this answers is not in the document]
mojito317: `json.dumps` converts primitive types into strings, so the following assertions will fail if I remove the instance check (actual and expected values are swapped in our tests, so in these cases the actual value is the expected and vice versa):

https://github.com/cloudant/python-cloudant/blob/master/tests/unit/param_translation_tests.py#L47 with
```
AssertionError: {'endkey': '10'} != {'endkey': 10}
- {'endkey': '10'}
?            -  -

+ {'endkey': 10}
```

https://github.com/cloudant/python-cloudant/blob/master/tests/unit/param_translation_tests.py#L123 with

```
{'key': 10} != {'key': '10'}

Expected :{'key': '10'}
Actual   :{'key': 10}
```
and
https://github.com/cloudant/python-cloudant/blob/master/tests/unit/param_translation_tests.py#L197 with
```
{'startkey': 10} != {'startkey': '10'}

Expected :{'startkey': '10'}
Actual   :{'startkey': 10}
```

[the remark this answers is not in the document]
ricellis: >`json.dumps` converts primitive types into strings

Well `json.dumps` always produces a python string, that's what it's for, but it shouldn't produce a JSON string for a primitive type (see: https://docs.python.org/3/library/json.html#py-to-json-table)

So the behaviour of `json.dumps` is like this:
int: `10` -> `'10'` (i.e. a python string)
str: `'10'` -> `'"10"'` (i.e. a python string of a JSON string)

>the following assertions will fail if I remove the instance check

Based on above are the tests are wrong or do we encounter a second round of encoding somewhere before the query string?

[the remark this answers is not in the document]
mojito317: You are right, the tests are wrong. Fixed in a7ac8a7.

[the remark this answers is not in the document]
ricellis: OK, so I see the reason the tests passed previously was because the `TYPE_CONVERTERS` for numbers did
`lambda x: x` - and that in turn would pass the number type to the requests module.
AFAICT [requests will ensure a `str` param value is utf-8 encoded](https://github.com/psf/requests/blob/1466ad713cf84738cd28f1224a7ab4a19e50e361/requests/models.py#L104) but just pass other types into the urlencoder. I guess the urlencoder itself handles numbers. In short I don't think this change is risky in the behaviour change from passing `str` instead of e.g. `int` in the request `params` dict.

ricellis:
Still needs a changes entry https://github.com/cloudant/python-cloudant/pull/501#pullrequestreview-676251324

but otherwise I'm ok with this

mojito317:
> Still needs a changes entry [#501 (review)](https://github.com/cloudant/python-cloudant/pull/501#pullrequestreview-676251324)
> 
> but otherwise I'm ok with this

Right, I added it in ab304b2.

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 - [FIXED] Fixed result paging for grouped view queries.
 - [FIXED] Incorrect use of username as account name in `Cloudant.bluemix()`.
 - [IMPROVED] Documented use of None account name and url override for `Cloudant.iam()`.
+- [FIXED] Use custom encoder on get_docs params.
```
ricellis, line 8: I think many won't know where `get_docs` is used as it's an internal function
```suggestion
- [FIXED] Use custom encoder (if provided) for all view `key` params not just `keys`.
```

ricellis (approved):
I suggested a slight clarification for the change log, but I'm +1 anyway.

mojito317:
Thanks for the review @emlaver and @ricellis! I'm on it to merge this.

